PR #21, opened 2021-01-19
add unit test

gaoteng-git:
1. Test profiler parsing and processing. Each test case covers one feature or boundary case.
2. Test range related functions.

### DIFF
--- a/tb_plugin/test/test_profiler.py
+++ b/tb_plugin/test/test_profiler.py
@@ -1,0 +1,700 @@
+import json
+import unittest
+
+import tensorboard_plugin_torch_profiler.profiler.trace as trace
+from tensorboard_plugin_torch_profiler.profiler.data import RunProfileData
+
+SCHEMA_VERSION = 1
+WORKER_NAME = "worker0"
+
+
+def parse_json_trace(json_content):
+    trace_json = json.loads(json_content)
+    profile = RunProfileData(WORKER_NAME)
+    parser = trace.get_event_parser(SCHEMA_VERSION)
+    profile.events = []
+    for data in trace_json:
+        event = parser.parse(data)
+        if event is not None:
+            profile.events.append(event)
+    return profile
+
+
+'''
+All the events in json string are only simulation, not actual generated events.
+We removed the data fields that not used by current version of our profiler,
+for easy to check correctness and shorter in length. 
+We even renamed the data values such as kernel name or "ts", to simplify the string.
+'''
+
+
+class TestProfiler(unittest.TestCase):
+    #  A test case including all 7 event categories.
+    def test_all_categories(self):
+        json_content = """
+          [{
+            "ph": "X", "cat": "Operator", 
+            "name": "enumerate(DataLoader)#_SingleProcessDataLoaderIter.__next__", "pid": 13721, "tid": "123",
+            "ts": 100, "dur": 180,
+            "args": {"Input dims": [], "External id": 2}
+          },
+          {
+            "ph": "X", "cat": "Operator", 
+            "name": "aten::to", "pid": 13721, "tid": "123",
+            "ts": 200, "dur": 60,
+            "args": {"Input dims": [[2, 8, 5], [], [], [], [], [], [], []], "External id": 3}
+          },
+          {
+            "ph": "X", "cat": "Operator", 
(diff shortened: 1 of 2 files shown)


tb_plugin/test/test_module_parser.py
```diff
@@ -0,0 +1,69 @@
+import math
+import pickle
+import unittest
+
+from tensorboard_plugin_torch_profiler.profiler.data import RunProfileData
+from tensorboard_plugin_torch_profiler.profiler.module_parser import ModuleParser
+
+OP_LIST = "./data/data_op_list_groupby_name.pkl"
+OP_LIST_BY_INPUT = "./data/data_op_list_groupby_name_input.pkl"
```
nbcsm, line 9: hmm, seems not easy to maintain many pickle files.

gdankel (commented):
Do you want to run this as part of CI? I'll look into enabling it for this repo.

gaoteng-git:
> Do you want to run this as part of CI? I'll look into enabling it for this repo.

Yes. Not urgent. Thanks a lot!

gdankel:
A few questions, out of curiosity:
1. What is the plan for this going forward? Add additional trace files and resulting baseline files as needed?
2. Isn't the pre-computed baseline files very sensitive to code changes? Will they have to be updated frequently?
3. Will there be instructions for how to re-generate traces (e.g. new features added, schema updates) and precomputed results?
4. I assume the input trace files will have a lot of repetition that doesn't add anything to the test. On the other hand I assume it's easier to maintain this way. But checking in MBs of trace files might not be a good long-term solution. Any thoughts on this?

gaoteng-git:
> A few questions, out of curiosity:
> 
> 1. What is the plan for this going forward? Add additional trace files and resulting baseline files as needed?
> 2. Isn't the pre-computed baseline files very sensitive to code changes? Will they have to be updated frequently?
> 3. Will there be instructions for how to re-generate traces (e.g. new features added, schema updates) and precomputed results?
> 4. I assume the input trace files will have a lot of repetition that doesn't add anything to the test. On the other hand I assume it's easier to maintain this way. But checking in MBs of trace files might not be a good long-term solution. Any thoughts on this?

Good questions. These real model profiling cases are more "real world", so we think they could be more robust to find out code bugs.  Their update frequency may not be too high because the data tables have slow evolving rate. We have re-generate scripts in private repo.
Yes, they are too big in size and hard to maintain in github. We'll make them as internal tests, and manually design some test cases with small input data to submit to this github.

tb_plugin/test/test_profiler.py
```diff
@@ -0,0 +1,437 @@
...
+            "ph": "X", "cat": "Kernel", 
+            "name": "void gemmSN_TN_kernel_64addr", "pid": 0, "tid": "stream 7",
+            "ts": 220, "dur": 8,
+            "args": {"correlation": 335, "external id": 0}
+          }]
+        """
+        profile = parse_json_trace(json_content)
+        profile.process()
+        self.assertEqual(profile.op_list_groupby_name[0].device_duration, 0)
+        self.assertEqual(profile.op_list_groupby_name[0].self_device_duration, 0)
```
guyang3532, line 359: Should this also test the accumulating device time?

gaoteng-git: In this test case, the Runtime is not invoked by the Operator. So the device durations of this Operator are 0.

gaoteng-git: > Should this also test the accumulating device time?

In this test case, the Runtime is not invoked by the Operator(because external id not equal). So the device durations of this Operator are 0.

tb_plugin/test/test_profiler.py
```diff
@@ -0,0 +1,640 @@
+import json
+import unittest
+
+import tensorboard_plugin_torch_profiler.profiler.trace as trace
```
nbcsm, line 4: can use relative import?

nbcsm (approved):

tb_plugin/test/test_profiler.py
gaoteng-git: > can use relative import?

Just follow same style of pytorch tests, such as [test_profiler.py](https://github.com/pytorch/pytorch/blob/master/test/test_profiler.py).

PR #25, opened 2021-01-25
Fix event sort bug and use external_id to attach runtime with operator

gaoteng-git:
1. Fix the bug: sort operator events by "(x.start_time, -x.end_time)".
2. Use external_id to attach RuntimeNode with OperatorNode. To handle the case "when RuntimeNode has duration 0 and starts at same time as a OperatorNode, just use interval containing relationship can't tell it is child or brother of the OperatorNode."
    2.1 In "remove_dup_nodes", give removed node's runtimes to father node.
    2.2 Handle runtime with external_id=0. Attach them to root node of calling tree.

### DIFF
--- a/tb_plugin/tensorboard_plugin_torch_profiler/profiler/module_parser.py
+++ b/tb_plugin/tensorboard_plugin_torch_profiler/profiler/module_parser.py
@@ -4,8 +4,8 @@
 
 import sys
 
-from .. import utils
 from .trace import EventTypes
+from .. import utils
 
 logger = utils.get_logger()
 
@@ -28,7 +28,8 @@ class HostNode(BaseNode):
 class OperatorNode(HostNode):
     def __init__(self):
         super(OperatorNode, self).__init__()
-        self.children = []
+        self.children = []  # OperatorNode and ProfilerStepNode.
+        self.runtimes = []  # RuntimeNode
         self.input_shape = None
         self.self_host_duration = 0
         self.self_device_duration = 0
@@ -38,10 +39,10 @@ class OperatorNode(HostNode):
         for child in self.children:
             self.device_duration += child.device_duration
             # To be consistent with pytorch autograd profiler.Include child Runtime time as self time.
-            if child.type != EventTypes.RUNTIME:
-                self.self_host_duration -= (child.end_time - child.start_time)
-            if isinstance(child, RuntimeNode):
-                self.self_device_duration += child.device_duration
+            self.self_host_duration -= (child.end_time - child.start_time)
+        for rt in self.runtimes:
+            self.device_duration += rt.device_duration
+            self.self_device_duration += rt.device_duration
 
 
 class ProfilerStepNode(OperatorNode):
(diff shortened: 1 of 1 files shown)


tb_plugin/tensorboard_plugin_torch_profiler/profiler/module_parser.py
```diff
@@ -235,9 +243,16 @@ def build_node(node, event):
                             logger.warning(
                                 "Runtime and Device-op have same correlation id but with different external id!"
                             )
-                if not tid in tid2list:
-                    tid2list[tid] = []
-                tid2list[tid].append(rt_node)
+                if rt_node.external_id in externalid_to_runtime:
+                    externalid_to_runtime[rt_node.external_id].append(rt_node)
+                else:
+                    externalid_to_runtime[rt_node.external_id] = [rt_node]
```
guyang3532, line 249: Is this external_id unique in multi-threads scenario?

tb_plugin/tensorboard_plugin_torch_profiler/profiler/module_parser.py
```diff
@@ -300,14 +315,30 @@ def parse_kernels(kernel_list):
...
-            parse_event(event, corrid_to_device, corrid_to_runtime, tid2list)
-        for tid, host_node_list in tid2list.items():
-            host_node_list.sort(key=lambda x: (x.start_time, x.end_time))
-            root_node = self._build_tree(host_node_list)
+            parse_event(event, corrid_to_device, corrid_to_runtime, externalid_to_runtime, tid2list, tid2rt_list)
+        # associate CUDA Runtimes with CPU events
+        for _, op_list in tid2list.items():
+            for op in op_list:
+                if op.external_id in externalid_to_runtime:
+                    op.runtimes.extend(externalid_to_runtime[op.external_id])
```
guyang3532, line 336: Do you want to add a check that if any runtime has external id that doesn't correspond to any op?

tb_plugin/tensorboard_plugin_torch_profiler/profiler/module_parser.py
```diff
@@ -300,14 +315,30 @@ def parse_kernels(kernel_list):
...
+        for tid, op_list in tid2list.items():
+            rt_list = tid2rt_list[tid] if tid in tid2rt_list else []
```
guyang3532, line 338: Rename this 'rt_list' to 'zero_rt_list' 'extra_rt_list' or something else? it will easy to understand.

tb_plugin/tensorboard_plugin_torch_profiler/profiler/module_parser.py
```diff
@@ -109,30 +110,34 @@ def __init__(self):
...
-                            if tail_node.type != EventTypes.RUNTIME:
-                                if (node.type == EventTypes.RUNTIME and
-                                        node.external_id != 0 and
-                                        tail_node.external_id != node.external_id):
-                                    logger.warning("Operator contains Runtime but with different external id!")
-                                tail_node.children.append(node)
-                                node_stack.append(node)
-                            else:
-                                logger.error("Error in input data: runtime range should not have children!")
+                            if (node.type == EventTypes.RUNTIME and
```
guyang3532, line 129: this node.type == EventTypes.RUNTIME will always be false, right?

tb_plugin/tensorboard_plugin_torch_profiler/profiler/module_parser.py
guyang3532, line 336: Is it necessary to add a check that if any runtime has external_id that is no zero and doesn't correlate to any operator?

tb_plugin/tensorboard_plugin_torch_profiler/profiler/module_parser.py
gaoteng-git: Good question. I double-checked the pytorch code. Now the external id is a static variable inside a function, so it is shared by multi-threads. Because it is atomic type, it is thread safe.
image
But as we see the above TODO, it seems they have plan to make it not unique. But making it not unique will make kernel ambiguous to attach to operator in which thread. I'll contact Gisle to confirm this.

tb_plugin/tensorboard_plugin_torch_profiler/profiler/module_parser.py
gaoteng-git: Thanks. Done.

tb_plugin/tensorboard_plugin_torch_profiler/profiler/module_parser.py
gaoteng-git: Yes. Removed it.

tb_plugin/tensorboard_plugin_torch_profiler/profiler/module_parser.py
gaoteng-git: Good. Done.

guyang3532 (approved):

tb_plugin/tensorboard_plugin_torch_profiler/profiler/module_parser.py
gdankel: That's right - as the comment says we might change the implementation but will still be unique within a single process.

PR #26, opened 2021-01-25
Support collapsible menu

Kingwl:


gaoteng-git:
Cool! However, please commit to tb_plugin branch firstly. Then we can test it more widely before merging to master branch.

PR #32, opened 2021-01-28
add dumping stack traces to chrome tracing file

gaoteng-git:
In order to show stack-traces in tensorboard plugin, we have to dump stack traces to chrome tracing file.
If this is merged, we will submit PR to torch/csrc/autograd/profiler_kineto.cpp, mainly adding following code into finalizeCPUTrace:
      if (kineto_events_[idx].hasStack()) {
        cpu_trace->activities[idx].callStack = stacksToStr(kineto_events_[idx].stack());
      }

### DIFF
--- a/libkineto/include/ClientTraceActivity.h
+++ b/libkineto/include/ClientTraceActivity.h
@@ -69,6 +69,7 @@ struct ClientTraceActivity : TraceActivity {
   std::string outputTypes;
   std::string inputNames;
   std::string outputNames;
+  std::string stackTraces;
 };
 
 } // namespace libkineto
--- a/libkineto/include/ClientTraceActivity.h
+++ b/libkineto/include/ClientTraceActivity.h
@@ -69,7 +69,7 @@ struct ClientTraceActivity : TraceActivity {
   std::string outputTypes;
   std::string inputNames;
   std::string outputNames;
-  std::string stackTraces;
+  std::string callStack;
 };
 
 } // namespace libkineto
--- a/libkineto/src/output_json.cpp
+++ b/libkineto/src/output_json.cpp
@@ -189,6 +189,7 @@ void ChromeTraceLogger::handleCpuActivity(
        "Input dims": {}, "Input type": {}, "Input names": {},
        "Output dims": {}, "Output type": {}, "Output names": {},
        "Device": {}, "External id": {}, "Extra arguments": {},
+       "Stack traces": "{}",
        "Trace name": "{}", "Trace iteration": {}
     }}
   }},)JSON",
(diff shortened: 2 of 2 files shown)


libkineto/include/ClientTraceActivity.h
```diff
@@ -69,6 +69,7 @@ struct ClientTraceActivity : TraceActivity {
   std::string outputTypes;
   std::string inputNames;
   std::string outputNames;
+  std::string stackTraces;
```
gdankel, line 72: Can we use callStack and Call stack?

gdankel:
Did you consider using the "sf" or "stack" fields for this? Also, can you point to some context - I might've missed parts of the discussion around this.

gaoteng-git:
> Did you consider using the "sf" or "stack" fields for this? Also, can you point to some context - I might've missed parts of the discussion around this.

Yes, we considered this before. You can find reference [here](https://docs.google.com/document/d/1CvAClvFfyA5R-PhYUmn5OOQtYMH4h6I0nSsKchNAySU/preview#heading=h.yr703knxre9f). From my try, "sf" works but ["stack" seems not implemented](https://github.com/catapult-project/catapult/issues/3996).
Because one advantage of "sf" is decreasing file size. We compared the chrome tracing file size. Profiling the the resnet50 tranining for 6 steps, the chrome tracing file without dumping callstacks is 26MB, the chrome tracing file with dumping callstacks as this PR is 40MB. So the increased size is not too much.
Another advantage of "sf" is showing call stacks in multiple lines in chrome, like this:
image

if using "sf", the estimated change is changing the "stackTraces" in "ClientTraceActivity" from string to vector<string>(each element is a frame of call stack), and in kineto side adding a function to organize all operators' stack frames into a tree, and dumping the tree into a global secion "stackFrames" in the result json file.

What's your opinion on this? Should we adopt "sf" or keep this in this version?

libkineto/include/ClientTraceActivity.h
gaoteng-git: Of course, updated.

gaoteng-git: > Can we use callStack and Call stack?

Of course, updated.

gdankel:
Thanks I recall the previous discussion now. I'm fine with landing this now and updating it in a future version. Only potential issue will be to change the type later without breaking anything.

gaoteng-git:
> Thanks I recall the previous discussion now. I'm fine with landing this now and updating it in a future version. Only potential issue will be to change the type later without breaking anything.

Thanks. Change which "type"? What do I need to do on my side?

gdankel (approved):

PR #34, opened 2021-02-02
fix grammer bug

gaoteng-git:


tb_plugin/tensorboard_plugin_torch_profiler/profiler/module_parser.py
```diff
@@ -334,7 +334,7 @@ def parse_kernels(kernel_list):
         for ext_id in externalid_to_runtime:
             if ext_id != 0:
                 logger.warning("{} Runtime with external id {} don't correlate to any operator!".format(
```
gdankel, line 336: This can happen quite frequently depending on workload... We can discuss how to report these sort of warnings in a better way if they become too noisy.

gdankel (approved):

tb_plugin/tensorboard_plugin_torch_profiler/profiler/module_parser.py
gaoteng-git: So far I've not seen it happens. I'll compress it to more concise message if this happens in future try.

PR #39, opened 2021-02-04
Add weak acc_get_device_type symbol

malfet:
Summary: cuPTI has a weak dependency on acc_get_device_type symbol, which might accidentally get resolved against OpenMP library

Differential Revision: D26256093


seemethere:
Relates to https://github.com/pytorch/pytorch/issues/51026

PR #48, opened 2021-02-08
Add basic CI github build action fo libkineto

gdankel:
Summary: Add a build action for libkineto. This will build the cpu-only version, without cupti headers. Next, I'll figure out how to get a build with cupti which will allow us to enable unit tests.

Differential Revision: D26322147


gdankel:
Something broke in the internal FB flow, closing.

PR #50, opened 2021-02-09
remove example and sample with autograd API

gaoteng-git:
Because autograd profiler API is going to deprecate, so I remove the related example code and sample data.


nbcsm (approved):

nbcsm:
@gaoteng-git , please update tests accordingly.

PR #52, opened 2021-02-09
[draft] Add github action to build and push fe files

Kingwl:
Build & update fe files after push if there's any changes (Which means will not commit without fe changes).
You can see [this commit](https://github.com/Kingwl/kineto/pull/1/commits/e6c803a1542e7767450957bf7502ad115b2ea73f)

### DIFF
--- a/.github/workflows/fe_build.yml
+++ b/.github/workflows/fe_build.yml
@@ -1,0 +1,25 @@
+name: FE Build
+
+on: [push]
+
+jobs:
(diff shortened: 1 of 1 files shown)


nbcsm:
@Kingwl , thanks for the contribution Wenlu.
Our CI is not quite ready yet, so I suggest to hold this change for now...
We will revisit it later once our CI/CD is ready.

PR #61, opened 2021-02-10
Merge tb_plugin branch to master

gaoteng-git:
1. Add end to end test. 
    1.1 test_tensorboard_end2end.py: It starts a tensorboard server, compare the web response with content of benchmark file(result_check_file.txt).
    1.2 test_compare_with_aurograd.py: It run a PyTorch sample code with profiling, and compare the result of PyTorch profiler with result of tensorboard plugin profiler.
2. Remove the profiling example using autograd API. (Because autograd API is going to be deprecated, so we should not show user example using it.)
3. Enable CI for tb_plugin
4. Update self_cpu_time to exclude runtime time from it. In order to keep consistent with PyTorch profiler lastest update.
5. Update .gitignore
6. Rename plugin and update related code influenced by renaming.


.github/workflows/test.yml
```diff
@@ -0,0 +1,41 @@
+name: Test
```
gdankel, line 1: Can you give this workflow a more specific name, e.g. tb_plugin_test.yml?

gdankel (changes requested):

.github/workflows/test.yml
gaoteng-git: Thanks, renamed.

PR #64, opened 2021-02-19
Profiler UI Update

liuziyue:
- Add a sorting arrow to show the sort order is active upon page load. 
- Adjust opacity for the step chart. Move the chart legend to the right. 
- Update ReadMe on install and uninstall instructions.


Kingwl (approved):
LGTM

nbcsm:
thanks for the fix, @liuziyue 👍 
cherrypicked into https://github.com/pytorch/kineto/pull/74

PR #70, opened 2021-02-23
Profiler UI AutoScale Unit

liuziyue:
Scale all numbers in the "Overview" view automatically from us to ms if needed. 

- Scale to ms if the average is > 10000
- Keep two decimals


Kingwl:
Just for note: Run `yarn prettier` before commit please.

nbcsm:
close it for now.

PR #73, opened 2021-02-24
[plugin release 0.1] Update trace view if switching workers

nbcsm:
when there are multiple workers in a run, Trace view is NOT refreshed when switching workers.

thanks @Kingwl for the fix.


Kingwl (approved):
LGTM

PR #84, opened 2021-02-25
split is_gpu_used into 3 switches

gaoteng-git:
Use has_runtime, has_kernel, has_memory to control GPU components in UI, in order to deal with more boundary cases.

PR #91, opened 2021-02-26
[plugin release 0.1] merge plugin change back to master

nbcsm:
- [Remove ci test for cuda11.2](https://github.com/pytorch/kineto/pull/90)
- [Add autograd api example](https://github.com/pytorch/kineto/pull/89)
- [Refine setup.py of torch_tb_profiler](https://github.com/pytorch/kineto/pull/88)
- [Update readme](https://github.com/pytorch/kineto/pull/87)
- [split is_gpu_used into 3 switches](https://github.com/pytorch/kineto/pull/84)
- [Fix the setup.py](https://github.com/pytorch/kineto/pull/81)


tb_plugin/examples/resnet50_autograd_api.py
```diff
@@ -0,0 +1,46 @@
+import os
```
gdankel, line 1: Why was this added back in?

gdankel: @ilia-cher are we OK with this being here?

ilia-cher: @gdankel i assume this is just an example for the tb plugin, was there some other concern?

nbcsm: since new profiler api will still be prototype or beta in PT 1.8, so I assume many users may still use the old profiler api. Not adding the autograd profiler example may make user feel that we don't support it. So I feel it is better to add an example.

nbcsm: we will remove this example after autograd profiler is officially deprecated in future release.

PR #93, opened 2021-02-27
Fix Keyboard Actions in Trace View UI

liuziyue:
Enable keyboard actions (ASDW keys to move around trace view) when the side panel is collapsed/reopened.


Kingwl:
It's not only the side panel but also whole outside In my memory...

Kingwl (approved):

gaoteng-git:
Moved to #101

PR #108, opened 2021-03-08
Consider from device side in overall view's step time

gaoteng-git:
Update overall view step time to following cases:
1. No step is specified
    Use left bound and max bound of all events as one big step in view.
2. One or more than one steps has no GPU event
    Use original host side step span as view's step time span.
3. All steps have GPU event.
    Record all runtimes that totally included in this step. Find all gpu events launched these runtimes, and record their min bound and max bound as "step device time".
    Use max{previous step's device side end time, this step's host side start time} as this view's step's start time.
    Use max{this step's host side end time, this step's device side end time} as this view's step's end time.
    The 1st step in the profiling file's "previous step's device side end time" is got by finding all device events that are not launched by the steps in this profiling file.


tb_plugin/torch_tb_profiler/profiler/overall_parser.py
```diff
@@ -177,22 +177,85 @@ def __init__(self):
...
+            elif runtime_node_list[i_runtime].end_time <= self.steps[i_step][1]:
+                # This runtime is inside this step span. Scan its device_nodes.
+                rt = runtime_node_list[i_runtime]
+                if rt.device_nodes is not None:
+                    for device_node in rt.device_nodes:
+                        step_device_min_ts = min(device_node.start_time, step_device_min_ts)
+                        step_device_max_ts = max(device_node.end_time, step_device_max_ts)
+                        matched_device_nodes.add(device_node)
+                i_runtime += 1
+            elif runtime_node_list[i_runtime].start_time <= self.steps[i_step][1]:
```
guyang3532, line 206: In what condition this intersection would happen?

tb_plugin/torch_tb_profiler/profiler/overall_parser.py
```diff
@@ -177,22 +177,85 @@ def __init__(self):
...
+        for steps_device_item in steps_device:
+            if steps_device_item[0] == sys.maxsize and steps_device_item[1] == -sys.maxsize - 1:
+                is_all_gpu = False
+        logger.debug("is_all_gpu={}".format(is_all_gpu))
+        if is_all_gpu:
+            prev_step_end_time = self.steps[0][0]
+            for device_node in device_node_list:
+                if device_node not in matched_device_nodes:
+                    # Now this device_node is not launched inside any step span.
+                    if device_node.end_time < steps_device_item[0]:
```
guyang3532, line 230: this 'steps_device_item' is define in for loop of  line 221

nbcsm:
please well test this change with some real workload, and provide some screenshots, before vs. after, for some typical cases.

gaoteng-git (commented):
> please well test this change with some real workload, and provide some screenshots, before vs. after, for some typical cases.

Before this PR, resnet50, DataLoader workers 0:
resnet50_num_workers_0
After this PR, resnet50, DataLoader workers 0:
resnet50_num_workers_0

======================================

Before this PR, resnet50, DataLoader workers 4:
resnet50_num_workers_4
After this PR, resnet50, DataLoader workers 4:
resnet50_num_workers_4

======================================

Before this PR, Bert, only CPU:
bert_newAPI_cpu
After this PR, Bert, only CPU:
bert_newAPI_cpu

======================================

Before this PR, Bert, DDP process0:
bert_newAPI_fp16_ddp_worker0
After this PR, Bert, DDP process0:
bert_newAPI_fp16_ddp_worker0
Before this PR, Bert, DDP process1:
bert_newAPI_fp16_ddp_worker1
After this PR, Bert, DDP process1:
bert_newAPI_fp16_ddp_worker1

======================================

Before this PR, Bert, autograd profiler:
bert_oldAPI_fp16
After this PR, Bert, autograd profiler:
bert_oldAPI_fp16

nbcsm:
why step time is smaller for resnet? expected?
for BERT, all numbers are exactly the same?

gaoteng-git:
> why step time is smaller for resnet? expected?
> for BERT, all numbers are exactly the same?

Resnet50:
This is all step time before this PR:
image
This is all step time after this PR:
image
So after this PR the average step time is shorter.

Bert:
Yes, it is expected. All steps' kernel time ends before its host side step. So the host side step time will be adopted as final step time.
image

tb_plugin/torch_tb_profiler/profiler/overall_parser.py
```diff
@@ -177,22 +177,99 @@ def __init__(self):
...
+                    if device_node not in matched_device_nodes:
+                        # Now this device_node is not launched inside any step span.
+                        if device_node.end_time < steps_device[0][0]:
+                            prev_step_end_time = max(prev_step_end_time, device_node.end_time)
+            for i_step in range(len(self.steps)):
+                step_start_time = max(prev_step_end_time, self.steps[i_step][0])
+                step_end_time = self.steps[i_step][1]
+                if steps_device[i_step][0] == sys.maxsize:  # When step i_step has no device event.
+                    # Assign to step_start_time when kernel is behind host step end.
+                    step_end_time = max(step_end_time, step_start_time)
```
guotuofeng, line 243: we need only update the step_end_time when there is step_devices time? for sys.maxsize case, just ignore it?

tb_plugin/torch_tb_profiler/profiler/overall_parser.py
```diff
@@ -177,22 +177,99 @@ def __init__(self):
...
+            # If there are steps during [i_step+1, len(self.steps) which has no runtime,
+            # then their steps_device will keep as initial value "(sys.maxsize, -sys.maxsize - 1)".
+        # Change step time to device side on the condition that any step have device time.
+        is_use_gpu = False
+        for steps_device_item in steps_device:
+            if steps_device_item[0] != sys.maxsize:
+                is_use_gpu = True
+                break
+        if is_use_gpu:
+            prev_step_end_time = self.steps[0][0]
```
guotuofeng, line 231: Is it possible that the steps[0][0] is sys.maxsize?

tb_plugin/torch_tb_profiler/profiler/overall_parser.py
```diff
@@ -177,22 +177,99 @@ def __init__(self):
...
+                # This runtime intersects with this step span. Skip this runtime.
+                i_runtime += 1
+            else:
+                # runtime_node_list[i_runtime].start_time >= step_host_end_time
+                # This runtime starts after this step's end. Record and move forward this step.
+                steps_device[i_step] = (step_device_min_ts, step_device_max_ts)
+                i_step += 1
+                step_device_min_ts = sys.maxsize
+                step_device_max_ts = -sys.maxsize - 1
+        if i_step < len(self.steps):
```
guotuofeng, line 220: should we use while instead of if?

tb_plugin/torch_tb_profiler/profiler/overall_parser.py
```diff
@@ -177,22 +177,99 @@ def __init__(self):
...
+                            prev_step_end_time = max(prev_step_end_time, device_node.end_time)
+            for i_step in range(len(self.steps)):
+                step_start_time = max(prev_step_end_time, self.steps[i_step][0])
+                step_end_time = self.steps[i_step][1]
+                if steps_device[i_step][0] == sys.maxsize:  # When step i_step has no device event.
+                    # Assign to step_start_time when kernel is behind host step end.
+                    step_end_time = max(step_end_time, step_start_time)
+                else:
+                    step_end_time = max(step_end_time, steps_device[i_step][1])
+                    if step_end_time < step_start_time:
```
guotuofeng, line 246: under what conditions we will hit the start > end?

tb_plugin/torch_tb_profiler/profiler/overall_parser.py
gaoteng-git: "steps" is got from host side "ProfilerStep#" and will not be sys.maxsize. Only "steps_device" could be.

tb_plugin/torch_tb_profiler/profiler/overall_parser.py
gaoteng-git: This is tricky, I'm not sure whether it will really happens, just add a guard here. When the step "i_step" launches kernels in a different stream than "i_step-1", this may occur.
image

tb_plugin/torch_tb_profiler/profiler/overall_parser.py
gaoteng-git: If we remove this if, then if "steps_device[i_step][0] == sys.maxsize and step_start_time > step_end_time" is truee, the following "logger.warning" will be triggered. But this case is our expected case and we don't want warning happens.

tb_plugin/torch_tb_profiler/profiler/overall_parser.py
gaoteng-git: Good style! Thanks. Done.

tb_plugin/torch_tb_profiler/profiler/overall_parser.py
```diff
@@ -177,22 +177,99 @@ def __init__(self):
...
+            else:
+                # runtime_node_list[i_runtime].start_time >= step_host_end_time
+                # This runtime starts after this step's end. Record and move forward this step.
+                steps_device[i_step] = (step_device_min_ts, step_device_max_ts)
+                i_step += 1
+                step_device_min_ts = sys.maxsize
+                step_device_max_ts = -sys.maxsize - 1
+        while i_step < len(self.steps):
+            # This step doesn't launch any device side event, just assign it as empty.
+            steps_device[i_step] = (step_device_min_ts, step_device_max_ts)
```
guyang3532, line 222: if should add 
step_device_min_ts = sys.maxsize
step_device_max_ts = -sys.maxsize - 1
after this line?

tb_plugin/torch_tb_profiler/profiler/overall_parser.py
```diff
@@ -177,22 +177,99 @@ def __init__(self):
...
+            i_step += 1
+        # Change step time to device side on the condition that any step have device time.
+        is_use_gpu = False
+        for steps_device_item in steps_device:
```
guyang3532, line 226: replace the for loop with 'matched_device_nodes is not empty' ?

guyang3532:
Maybe you need test it with a pure cpu case.

tb_plugin/torch_tb_profiler/profiler/overall_parser.py
gaoteng-git: Thanks. Fixed.

tb_plugin/torch_tb_profiler/profiler/overall_parser.py
gaoteng-git: Thanks. Done.

gaoteng-git:
> Maybe you need test it with a pure cpu case.

Thanks. Done.

gaoteng-git:
Test on PyTorch 1.8.1 RC1:
resnet50, batch_size=128:
image
image

resnet50, batch_size=32:
image
image

nbcsm:
how is looks like for before this change?

gaoteng-git:
> how is looks like for before this change?
> for bs=128, maybe exclude step 10?
Using the same tracing files as above, roll back to release/0.1 branch which uses host-side step.

resnet50, batch_size=128:
image

resnet50, batch_size=32:
image

gaoteng-git:
> how is looks like for before this change?
> for bs=128, maybe exclude step 10?

1. Pure CPU:
image
Before this PR:
image
After this PR:
image

===========================
2. No step used:
image
Before this PR:
image
After this PR:
image

===========================
3. batch_size=1, kernels are all inside host side step:
image
Before this PR:
image
After this PR:
image

===========================
4. batch_size=32, kernels are outside host side step, not remove last step:
image
Before this PR:
image
After this PR:
image

===========================
5. batch_size=128, kernels are outside host side step, remove last step:
image
Before this PR:
image
After this PR:
image

tb_plugin/torch_tb_profiler/profiler/overall_parser.py
```diff
@@ -177,22 +177,119 @@ def __init__(self):
...
+                keep_steps = i_step + 1
+                if i_step > 0 and steps_matched_device_nodes[i_step - 1] * 0.8 > steps_matched_device_nodes[i_step]:
+                    keep_steps = i_step
+                if keep_steps < len(self.steps):
+                    logger.warning(
+                        "Remove the last {} steps because of suspected lost of kernels in tracing file".format(
+                            len(self.steps) - keep_steps
+                        ))
+                    self.steps = self.steps[:keep_steps]
+                    self.steps_names = self.steps_names[:keep_steps]
```
nbcsm, line 272: it is possible that some normal step may contain less kernels, e.g. gradient accumulation scenario.

as we discussed, find the last kernel, map it to step, then make exclusion decision.

tb_plugin/torch_tb_profiler/profiler/overall_parser.py
```diff
@@ -177,22 +177,119 @@ def __init__(self):
...
+                if steps_matched_device_nodes[i_step] > 0:
+                    break
+                i_step -= 1
+            if i_step >= 0:
+                keep_steps = i_step + 1
+                if i_step > 0 and steps_matched_device_nodes[i_step - 1] * 0.8 > steps_matched_device_nodes[i_step]:
+                    keep_steps = i_step
+                if keep_steps < len(self.steps):
+                    logger.warning(
+                        "Remove the last {} steps because of suspected lost of kernels in tracing file".format(
```
nbcsm, line 268: refine error message.

nbcsm (changes requested):

nbcsm:
what is the result for 
- no profiler.step call, but tailing kernels are missing
- with only 1 or 2 steps, and tailing kernels are missing

nbcsm:
also, for no step used case above, why it is still autograd profiler?

tb_plugin/torch_tb_profiler/profiler/overall_parser.py
```diff
@@ -177,22 +177,119 @@ def __init__(self):
...
+                            len(self.steps) - keep_steps
+                        ))
+                    self.steps = self.steps[:keep_steps]
```
nbcsm, line 271: does this affect data loader related recommendation rule?

tb_plugin/torch_tb_profiler/profiler/overall_parser.py
```diff
@@ -177,22 +177,119 @@ def __init__(self):
...
+        is_remove_tail_steps = True  # TODO: Use tensorboard argument instead.
+        if is_use_gpu and len(self.steps) > 1 and is_remove_tail_steps:
+            i_step = len(self.steps) - 1
+            while i_step >= 0:
+                if steps_matched_device_nodes[i_step] > 0:
+                    break
+                i_step -= 1
+            if i_step >= 0:
+                keep_steps = i_step + 1
+                if i_step > 0 and steps_matched_device_nodes[i_step - 1] * 0.8 > steps_matched_device_nodes[i_step]:
```
guotuofeng, line 264: the 80% seems to be total steps instead of steps which missing 80% kernels?

gaoteng-git:
> what is the result for
> 
> * no profiler.step call, but tailing kernels are missing
> * with only 1 or 2 steps, and tailing kernels are missing

no profiler.step call, but tailing kernels are missing:
image
Zoom in, and see that the tailing kernels are missing:
image
Before this PR:
image
After this PR:
image

 with only 1 step, then just show this step:
image
Before this PR:
image
After this PR: (step time is shorter than the above, because the head kernels are launched by previous step, not this step)
image

with only 2 steps, then remove the 2nd step:
image
Before this PR:
image
After this PR:
image

tb_plugin/torch_tb_profiler/profiler/overall_parser.py
gaoteng-git: Thanks. Done.

gaoteng-git:
> also, for no step used case above, why it is still autograd profiler?
autograd profiler doesn't use "prof.step()".

tb_plugin/torch_tb_profiler/profiler/overall_parser.py
gaoteng-git: The removed steps will not participate in time-break-down and recommendation rule.

tb_plugin/torch_tb_profiler/profiler/overall_parser.py
gaoteng-git: For example, the step "i_step-1" has 100 kernels, then if step "i_step" has less than 80 kernels, we will remove "i_step".

gaoteng-git:
Bert, batch_size=32:
image
Before this PR:
image
After this PR:
image

nbcsm (approved):

guotuofeng (approved):

guyang3532 (approved):

PR #111, opened 2021-03-10
Fix Pie Chart Tooltip Overlap

liuziyue:
- Fixed pie chart tooltip overlap bug by removing excessive tooltip.
- Refined TraceView code.


Kingwl (approved):
LGTM, Thanks!

PR #128, opened 2021-03-24
add surrounding quote for N/A to make sure it can be loaded by chrome…

guotuofeng:
Add double quote to fix the bug when loading tracing file in chrome tracing.


guotuofeng:
abandon this PR since there is already another fix

PR #131, opened 2021-03-25
add workaround for N/A bug in kineto

guotuofeng:
Add workaround for N/A which already fixed in libkineto but not in next release.

### DIFF
--- a/tb_plugin/torch_tb_profiler/profiler/data.py
+++ b/tb_plugin/torch_tb_profiler/profiler/data.py
@@ -66,7 +66,22 @@ class RunProfileData(object):
             # Kineto may export json file with non-ascii code. before this is fixed, use a workaround
             # to handleJSONDecodeError, re-encode it and save to a temp file
             with fopen(trace_path, 'r') as f:
-                trace_json = json.load(f, strict=False)
+                try:
+                    trace_json = json.load(f, strict=False)
+                except json.decoder.JSONDecodeError:
+                    # TODO: remove the workaround after the libkineto fix for N/A is merged into pytorch
+                    if os.path.splitext(trace_path)[1].lower() != ".json":
+                        # ignore gz files
+                        raise
+
+                    f.seek(0)
+                    patched_json = trace_path + "_fixed.json"
+                    with open(patched_json, "wt") as fout:
+                        for line in f:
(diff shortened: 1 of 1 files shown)


tb_plugin/torch_tb_profiler/profiler/data.py
```diff
@@ -66,7 +66,22 @@ def parse(run_dir, worker):
             # Kineto may export json file with non-ascii code. before this is fixed, use a workaround
```
nbcsm, line 66: is N/A case still `json.decoder.JSONDecodeError`?

tb_plugin/torch_tb_profiler/profiler/data.py
```diff
@@ -66,7 +66,22 @@ def parse(run_dir, worker):
...
+                    # TODO: remove the workaround after the libkineto fix for N/A is merged into pytorch
+                    if os.path.splitext(trace_path)[1].lower() != ".json":
+                        # ignore gz files
+                        raise
+
+                    f.seek(0)
+                    patched_json = trace_path + "_fixed.json"
+                    with open(patched_json, "wt") as fout:
+                        for line in f:
+                            fout.write(line.replace('N/A', '\"N/A\"'))
```
nbcsm, line 81: maybe more strict?

tb_plugin/torch_tb_profiler/profiler/data.py
```diff
@@ -66,7 +66,22 @@ def parse(run_dir, worker):
             # Kineto may export json file with non-ascii code. before this is fixed, use a workaround
             # to handleJSONDecodeError, re-encode it and save to a temp file
             with fopen(trace_path, 'r') as f:
-                trace_json = json.load(f, strict=False)
+                try:
+                    trace_json = json.load(f, strict=False)
+                except json.decoder.JSONDecodeError:
+                    # TODO: remove the workaround after the libkineto fix for N/A is merged into pytorch
+                    if os.path.splitext(trace_path)[1].lower() != ".json":
```
gaoteng-git, line 73: ".gz" seems also works.

tb_plugin/torch_tb_profiler/profiler/data.py
guotuofeng: yes

PR #133, opened 2021-03-25
add N/A workaround when there is no quote in trace file

guotuofeng:


tb_plugin/torch_tb_profiler/profiler/data.py
```diff
@@ -65,8 +66,17 @@ def parse(run_dir, worker):
...
-                trace_json = json.load(f, strict=False)
+            with fopen(trace_path, 'rt') as f:
+                try:
+                    trace_json = json.load(f, strict=False)
+                except json.decoder.JSONDecodeError:
+                    # TODO: remove the workaround after the libkineto fix for N/A is merged into pytorch
+                    f.seek(0)
+                    with io.StringIO() as fout:
+                        for line in f:
+                            fout.write(line.replace('N/A', '\"N/A\"'))
```
nbcsm, line 77: better to be more strict.

guotuofeng: will use regex r'(?<!")N/A(?!")' to replace only the string without double quote

nbcsm (approved):

tb_plugin/torch_tb_profiler/profiler/data.py
```diff
@@ -65,8 +67,18 @@ def parse(run_dir, worker):
         except json.decoder.JSONDecodeError as e:
             # Kineto may export json file with non-ascii code. before this is fixed, use a workaround
```
gdankel, line 68: Please update comment as this is no longer the case that is being handled.

PR #140, opened 2021-03-26
Resize the graph when window resized

### LINKED ISSUE #125: In plugin Overall View, chart does not resize with size of browser window
Start tensorboard with non-max browser Window, then maximize browser Window.
In Overall View, charts stay small.

image

Kingwl:
Fixes #125

This PR added a hooks called `useResizeEventDependency` as well as some minor style fix.

It exposed a value `version` who will be increase when the window resized. And we could just put the `version` into chat dependency list to re-draw the chart when window resized.


nbcsm (changes requested):
need to build index.html?

Kingwl:
Yep.
And I strongly recommends to add the github action who automatic build the bundle after push (https://github.com/pytorch/kineto/pull/52)

PR #145, opened 2021-04-01
[draft] Add metadata support for distributed and cuda memory info

guotuofeng:
Prototype of adding metadata when starting profiler.

### DIFF
--- a/libkineto/src/ActivityProfilerController.cpp
+++ b/libkineto/src/ActivityProfilerController.cpp
@@ -118,7 +118,7 @@ void ActivityProfilerController::scheduleTrace(const Config& config) {
   }
 }
 
-void ActivityProfilerController::prepareTrace(const Config& config, const Metadata& metadata) {
+void ActivityProfilerController::prepareTrace(const Config& config, std::unique_ptr<Metadata>&& metadata) {
   // Requests from ActivityProfilerApi have higher priority than
   // requests from other sources (signal, daemon).
   // Cancel any ongoing request and refuse new ones.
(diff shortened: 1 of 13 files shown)


libkineto/src/output_json.cpp
```diff
@@ -31,12 +31,48 @@ namespace KINETO_NAMESPACE {
...
+      }}
+    )JSON";
+  }
+
+  if (!metadata.distributed_.backend_.empty()){
+    stream << fmt::format(R"JSON(
+      "distributed": {{
+        "backend": "{}",
+        "rank": {},
+        "work_size": {}
```
nbcsm, line 67: world

libkineto/src/output_json.cpp
```diff
@@ -31,12 +31,48 @@ namespace KINETO_NAMESPACE {
...
+  
+    for (size_t i = 0; i < metadata.gpus_.size(); i++) {
+      const auto& gpu = metadata.gpus_[i];
+      stream << fmt::format(R"JSON(
+        "GPU {}": {{
+          "id": {},
+          "sku": "{}",
+          "memory": {}
+        }},
+      )JSON", i, i, gpu.name_, gpu.totalMemory_);
```
nbcsm, line 54: `id` must align with `device id`

libkineto/src/ActivityProfilerController.cpp
```diff
@@ -134,6 +134,7 @@ void ActivityProfilerController::prepareTrace(const Config& config) {
   }
 
   profiler_->configure(config, now);
+  profiler_->prepareTrace(metadata);
```
guyang3532: Merging these two function seems more concise? like profiler_->configure(config, now, metadata); ?

gdankel:
Thanks for the patch! I will implement it differently for a couple of reasons:
1) We want to support metadata like this but in a more generic way
2) I have some refactoring in progress and I'll add support for this in a cleaner way once that's in place.

nbcsm:
> Thanks for the patch! I will implement it differently for a couple of reasons:
> 
> 1. We want to support metadata like this but in a more generic way
> 2. I have some refactoring in progress and I'll add support for this in a cleaner way once that's in place.

Cool, thanks.
When will this happen?

PR #166, opened 2021-04-14
fix security warning like X-Content-Type-Options is required to be no…

guotuofeng:
Fix the warning complained by tensorboard like:

W0414 14:56:00.098793 140429893994240 security_validator.py:49] In 3.0, this warning will become an error:
X-Content-Type-Options is required to be "nosniff"


guotuofeng:
bug is created to track this issue.
https://github.com/pytorch/kineto/issues/169

PR #168, opened 2021-04-15
Add validation for the TopN ops/kernels field

TomWildenhain-Microsoft:


tb_plugin/fe/src/utils/top.ts
```diff
@@ -30,8 +30,13 @@ export function useTopN(options?: IOptions) {
...
+    setActualDebounce(useTop === UseTop.Use && topIsValid(topText) ? Number(topText): actualTop)
+  }, [topText, useTop])
 
-  return [top, actualTop, useTop, setTop, setUseTop] as const
+  return [topText, actualTop, useTop, setTopText, setUseTop] as const
+}
+
+export function topIsValid(topText: string) {
+  const top = Number(topText)
+  return top > 0 && Number.isInteger(top)
```
guotuofeng, line 41: need we us isNaN for top?

tb_plugin/fe/src/utils/top.ts
```diff
@@ -30,8 +30,13 @@ export function useTopN(options?: IOptions) {
     ? React.useCallback(debounce(setActualTop, options.wait ?? 500), [])
     : setActualTop
   React.useEffect(() => {
-    setActualDebounce(useTop === UseTop.Use && top > 0 ? top : undefined)
-  }, [top, useTop])
+    setActualDebounce(useTop === UseTop.Use && topIsValid(topText) ? Number(topText): actualTop)
```
guotuofeng, line 33: in case of not valid, what the actualtop will be?

guotuofeng (changes requested):

tb_plugin/fe/src/utils/top.ts
TomWildenhain-Microsoft: For style I probably should. Technically NaN > 0 is false

tb_plugin/fe/src/utils/top.ts
TomWildenhain-Microsoft: The previous valid value.  If the user enters something invalid, the graph will not be updated until a valid value is provided.

Issue #162, opened 2021-04-13
Page scroll up


gaoteng-git:
Scroll the page down:
![image](https://user-images.githubusercontent.com/10429114/114506174-ffaaa180-9c63-11eb-9bac-abb5e0f5cd4f.png)

Input a string into "Search by Name" and press enter:
![image](https://user-images.githubusercontent.com/10429114/114511052-fde3dc80-9c69-11eb-8320-ee43644c73c9.png)

The page is automatically scrolled up. It is not a good user experience which makes user lose focus.

The "Group By" still behave like this. 
The "Group By" and "Search by Name" in "Kernel View" also behave like this.

TomWildenhain-Microsoft:
The scrolling is caused by the table redraw.  When the table is deleted, the page scrolls up.  As long as this doesn't bring us to the very top of the page, Chrome scrolls back down when the table is re-added, but on the Kernels page it is too short so it gets stuck at the top and doesn't go back.  Probably the best solution would be hooking the table redraw to add a placeholder box while drawing so the page height doesn't change.

gaoteng-git:
Solved by PR https://github.com/pytorch/kineto/pull/183

PR #183, opened 2021-04-22
Prevent Kernels page from scrolling to top on table redraw

TomWildenhain-Microsoft:
Solve issue https://github.com/pytorch/kineto/issues/162


tb_plugin/fe/src/components/charts/TableChart.tsx
```diff
@@ -69,7 +69,11 @@ export const TableChart: React.FC<IProps> = (props) => {
 
     const chart = new google.visualization.Table(element)
 
+    if (element.offsetHeight > 0) {
```
guotuofeng, line 72: it seems a little bit hack to work around this issue by simply set the height before redraw and reset after redraw?

Are there any better approach?

gaoteng-git:
Please also add the index.html updates.

gaoteng-git:
In my try, it works in the "Operator view". But it still doesn't work in the "Kernel view".

tb_plugin/fe/src/components/charts/TableChart.tsx
TomWildenhain-Microsoft: Yes it's definitely a bit of a hack but I think it might be the easiest solution that doesn't require editing the Google Charts library.

guotuofeng: could we add some comments about the reason why use the hack way?

TomWildenhain-Microsoft:
> In my try, it works in the "Operator view". But it still doesn't work in the "Kernel view".

Worked with @gaoteng-git offline and it worked after rebuilding.

guotuofeng (approved):

PR #185, opened 2021-04-26
add 2 kernel metrics

gaoteng-git:
1. Update "warps per sm" to "blocks per sm".
2. Add "occupancy" per kernel.


libkineto/src/output_json.cpp
```diff
@@ -292,12 +293,34 @@ void ChromeTraceLogger::handleGpuActivity(
...
   if (smCount_) {
-    warps_per_sm = (kernel->gridX * kernel->gridY * kernel->gridZ) *
-        (kernel->blockX * kernel->blockY * kernel->blockZ) /
-        (float) threads_per_warp / smCount_;
+    blocks_per_sm = (kernel->gridX * kernel->gridY * kernel->gridZ) / (float) smCount_;
   }
+
+  // Calculate occupancy
+  float occupancy = -1.0;
+  if (kernel->deviceId < occProps_.size()) {
```
gdankel, line 303: Please separate this out into its own function, or even better, its own file, e.g. CudaDeviceProperties.h / cpp.

libkineto/src/CuptiActivityInterface.cpp
```diff
@@ -92,6 +92,35 @@ int CuptiActivityInterface::smCount() {
   return sm_count;
 }
 
+#ifdef HAS_CUPTI
+static std::vector<cudaOccDeviceProp> getOccDeviceProp() {
```
gdankel, line 96: Since this is not CUPTI let's move it into its own file, as suggested below... e.g. CudaDeviceProperties.h/cpp

libkineto/src/output_json.cpp
```diff
@@ -55,6 +55,7 @@ ChromeTraceLogger::ChromeTraceLogger(const std::string& traceFileName, int smCou
   openTraceFile(fileName_, traceOf_);
 #ifdef HAS_CUPTI
   smCount_ = CuptiActivityInterface::singleton().smCount();
```
gdankel, line 57: I assume sm count is also available via cudaOccDeviceProp? If so, let's move to that (in a separate PR - I can do it as well).

gdankel (changes requested):

libkineto/src/output_json.cpp
gaoteng-git: Thanks. Removed to CudaDeviceProperties.cpp

libkineto/src/CuptiActivityInterface.cpp
gaoteng-git: Thanks. Removed to CudaDeviceProperties.cpp

libkineto/src/output_json.cpp
gaoteng-git: Yes, we can reuse cudaOccDeviceProp.numSms. Let's do it in another PR.

libkineto/src/CudaDeviceProperties.cpp
```diff
@@ -0,0 +1,68 @@
+/*
+ * Copyright (c) Facebook, Inc. and its affiliates.
```
gdankel, line 2: Let me check whether we add a different copyright header...

libkineto/src/CudaDeviceProperties.cpp
```diff
@@ -0,0 +1,68 @@
...
+#include "CudaDeviceProperties.h"
+
+#include <vector>
+
+#include <cuda_runtime.h>
+#include <cuda_occupancy.h>
+
+namespace KINETO_NAMESPACE {
+
+std::vector<cudaOccDeviceProp> occProps;
```
gdankel, line 17: static or anon namespace?

libkineto/src/CudaDeviceProperties.cpp
```diff
@@ -0,0 +1,68 @@
...
+
+void initOccDeviceProp() {
```
gdankel, line 19: How about calling this lazily from within this file so it doesn't have to be explicitly initialized:

namespace {

std::vector<cudaOccDeviceProp> createOccDeviceProps() {
  std::vector<cudaOccDeviceProp> res;
  ...
  return res;
}

std::vector<cudaOccDeviceProps>& occDeviceProps() {
  static std::vector<cudaOccDeviceProp> props = createOccDeviceProps();
  return props;
}

} // namespace

Then in getKernelOccupancy replace occProps with occDeviceProps()

gdankel (changes requested):

libkineto/src/CudaDeviceProperties.cpp
gaoteng-git: Thanks! Fixed.

libkineto/src/CudaDeviceProperties.cpp
gaoteng-git: The previous "occProps" is a static variable which does not expose to other files. Now removed and put to static variable inside function "occDeviceProps".

Issue #189, opened 2021-04-27
Tensorboard hangs in VS code


TomWildenhain-Microsoft:
VScode opens tensorboard but hangs forever (displaying a spinner).  The same happens in the browser but is resolved on refreshing the page.  The cause appear to be a race between tensorboard loading/finding the trace files and the web page opening.  Since VScode loads the frame at the same time as starting tensorboard and there is no refresh for the frame in VSCode, there's no easy way for the user to get the page to load.

![image](https://user-images.githubusercontent.com/67606533/116169138-08728d00-a6d2-11eb-9088-caf5235b21f2.png)

PR #191, opened 2021-04-28
Continuously fetch new runs

TomWildenhain-Microsoft:
Fixes the issue of hanging in VS code and updates runs automatically without refresh.  The runs list should be small (has just run name, not content) and should be inexpensive to refresh.
Fix issue: https://github.com/pytorch/kineto/issues/189


guotuofeng (approved):

Kingwl:
Sorry. May I ask why we fetch the runs forever? 

I thought we could just get any run and assume the server was ready.

TomWildenhain-Microsoft:
> Sorry. May I ask why we fetch the runs forever?
> 
> I thought we could just get any run and assume the server was ready.
> 
> And even if we were to get it forever, it should be spaced out over a longer period (1s? I guess).

If you want it to be longer (1 sec instead of 0.1 sec) that's fine.  It fetches continuously because TB actually takes time to parse each log and makes them available as it parses them. If you only wait for non-empty runs you will often only get the first run.  We could do a back-off type strategy (delay 0.1 sec for the first 5 secs and 1 sec after).  That said, the call is not expensive (just a list of names) so I don't think the 0.1 sec frequency should be an issue.  Keep in mind that it is a 0.1 sec delay after getting a response, so if the response time is really long and exceeds 0.1 secs it won't queue up multiple requests.

guotuofeng:
> Sorry. May I ask why we fetch the runs forever?
> 
> I thought we could just get any run and assume the server was ready.
> 
> And even if we were to get it forever, it should be spaced out over a longer period (1s? I guess).

We add support for AWS, Azure Blobs and Google Cloud. In this case, there is no guarantee when all data could be downloaded to local. From this point of view, the waiting time is undetermined.  @TomWildenhain-Microsoft's fix would fix the long loading timer when the log dir is in remote path like AWS, Azure Blobs.

Kingwl:
Okay. got it.

/cc: @guyang3532

PR #209, opened 2021-05-10
dump cudaOccDeviceProp

gaoteng-git:
Dump cudaOccDeviceProp to json trace file, for post-processing occupancy's timeline in tb_plugin.

Example output (because the value of metadata is type of string, the inner double quote must be headed with a slash):
```
  "metadata": {
      "cudaOccDeviceProps": "[{\"computeMajor\": 7, \"computeMinor\": 0, \"maxThreadsPerBlock\": 1024, \"maxThreadsPerMultiprocessor\": 2048, \"regsPerBlock\": 65536, \"regsPerMultiprocessor\": 65536, \"warpSize\": 32, \"sharedMemPerBlock\": 49152, \"sharedMemPerMultiprocessor\": 98304, \"numSms\": 80, \"sharedMemPerBlockOptin\": 98304}, {\"computeMajor\": 7, \"computeMinor\": 0, \"maxThreadsPerBlock\": 1024, \"maxThreadsPerMultiprocessor\": 2048, \"regsPerBlock\": 65536, \"regsPerMultiprocessor\": 65536, \"warpSize\": 32, \"sharedMemPerBlock\": 49152, \"sharedMemPerMultiprocessor\": 98304, \"numSms\": 80, \"sharedMemPerBlockOptin\": 98304}, {\"computeMajor\": 7, \"computeMinor\": 0, \"maxThreadsPerBlock\": 1024, \"maxThreadsPerMultiprocessor\": 2048, \"regsPerBlock\": 65536, \"regsPerMultiprocessor\": 65536, \"warpSize\": 32, \"sharedMemPerBlock\": 49152, \"sharedMemPerMultiprocessor\": 98304, \"numSms\": 80, \"sharedMemPerBlockOptin\": 98304}, {\"computeMajor\": 7, \"computeMinor\": 0, \"maxThreadsPerBlock\": 1024, \"maxThreadsPerMultiprocessor\": 2048, \"regsPerBlock\": 65536, \"regsPerMultiprocessor\": 65536, \"warpSize\": 32, \"sharedMemPerBlock\": 49152, \"sharedMemPerMultiprocessor\": 98304, \"numSms\": 80, \"sharedMemPerBlockOptin\": 98304}]"
  },
```

PR #213, opened 2021-05-11
Log device properties to trace file

gaoteng-git:
Device properties are useful in any case, but especially for performing analysis on traces such as occupancy.

This patch is a re-implementation of #209 
And it added 2 fields "name" and "totalGlobalMem", to remove duplicate from PyTorch side dump.
It adds "ifdef" to ignore dumping these fields when build on pure cpu.

### DIFF
--- a/libkineto/src/output_json.cpp
+++ b/libkineto/src/output_json.cpp
@@ -39,21 +39,20 @@ void ChromeTraceLogger::handleTraceStart(
   )JSON", kSchemaVersion);
 
   if (!metadata.empty()) {
-    traceOf_ << R"JSON(
-  "metadata": {
-  )JSON";
-    bool first = true;
+    std::vector<std::string> vals;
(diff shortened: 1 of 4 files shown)


libkineto/src/output_json.cpp
```diff
@@ -39,22 +39,23 @@ void ChromeTraceLogger::handleTraceStart(
...
-  )JSON";
+    traceOf_ << fmt::format(R"JSON(
+  "metadata": {{
+    {}
+  }},)JSON", fmt::join(vals, ",\n"));
   }
 
+#ifdef HAS_CUPTI
+  traceOf_ << fmt::format(R"JSON(
+  "computeProperties": [
```
guotuofeng, line 54: need confirm with @gdankel about using top level property or putting it into metadata.

My personal preference is put it into metadata.

guotuofeng (changes requested):

Issue #156, opened 2021-04-08
[Task] Multiple tracing files support


guotuofeng:

guotuofeng:
https://github.com/pytorch/kineto/issues/190

PR #215, opened 2021-05-11
Add multiple span route support

guotuofeng:
Add multiple span route support (part 1)


guotuofeng:
https://github.com/pytorch/kineto/issues/156

tb_plugin/torch_tb_profiler/plugin.py
```diff
@@ -121,8 +122,13 @@ def views_route(self, request):
         name = request.args.get("run")
         worker = request.args.get("worker")
         run = self._get_run(name)
-        profile = run.get_profile(worker)
-        views = sorted(profile.views, key=lambda x: x.id)
+        worker = run.workers[0]
```
guyang3532, line 125: worker has been received from 'worker = request.args.get("worker")' in line123 ?

guyang3532: Or this function can modify to 'profile = self._get_profile_for_request(request); views = sorted(profile.views,...);'

tb_plugin/torch_tb_profiler/profiler/run_generator.py
```diff
@@ -332,7 +333,8 @@ def _generate_wait_graph(self):
 
     def _generate_ops_table(self):
         workers_to_comm_ops = OrderedDict()
-        for worker,data in self.all_profile_data.items():
+        # Ignore the span for distributed view
+        for (worker, span), data in self.all_profile_data.items():
```
guyang3532, line 337: This need to be modified, because there may be data of same worker with different spans. We should aggregate data with same worker and multiple spans.

tb_plugin/torch_tb_profiler/profiler/loader.py
```diff
@@ -40,33 +41,54 @@ def _parse(self):
...
+        if spans is None:
+            self.has_communication = self._process_profiles(self.run.profiles.values())
+        else:
+            self.has_communication = {}
+            for span in spans:
+                profiles = self.run.get_profiles(span)
+                self.has_communication[span] = self._process_profiles(profiles)
+
+    def _process_profiles(self, profiles):
+        has_communication = True
```
guyang3532, line 75: Add an check : 'if not profiles: return False'?

tb_plugin/torch_tb_profiler/profiler/data.py
```diff
@@ -27,9 +27,33 @@ def __init__(self, name, run_dir):
...
+
+        spans = list(set(spans))
+        if len(spans) == 1 and spans[0] is None:
+            return None
+        else:
+            return spans
+
+    def get_profiles(self, worker=None, span=None):
+        if worker and span:
+            return self.get_profile(worker, span)
```
guyang3532, line 44: This class seems has no get_profile function?

tb_plugin/torch_tb_profiler/profiler/loader.py
guotuofeng: we should not add the check here. if it happens, there must be some bugs at other place.

tb_plugin/torch_tb_profiler/consts.py
```diff
@@ -7,8 +7,7 @@
 PLUGIN_NAME = "pytorch_profiler"
 
 WORKER_PATTERN = re.compile(r"""^(.*?) # worker name
-        # TODO: uncomment the following line when we need supprort multiple steps
-        # (?:\.\d+)? # optional timestamp like 1619499959628
+        (\.\d+)? # optional timestamp like 1619499959628 used as step name
```
gaoteng-git, line 10: used as "span time", not "step time".

tb_plugin/torch_tb_profiler/profiler/loader.py
```diff
@@ -1,40 +1,56 @@
...
+            if io.isdir(io.join(self.run_dir, path)):
                 continue
             match = consts.WORKER_PATTERN.match(path)
             if not match:
                 continue
 
             worker = match.group(1)
-            workers.append((worker, path))
+            span = match.group(2)
+            if span:
```
gaoteng-git, line 37: Have you tested that when match.group(2) fails then span will be None?

gaoteng-git (commented):
In trace view, when click to change span, the page will jump to overview page of the new span. The expected behavior is keeping in trace view page of the new span. Maybe update after Tom's coming.

PR #220, opened 2021-05-12
refactor gs/blob file system

guotuofeng:
refactor gs/blob file system


Reubend (approved):
This looks great to me! Thank you for doing this refactoring, I think it's much cleaner now and it makes a lot of sense. :)

PR #221, opened 2021-05-12
Add gpu metrics

gaoteng-git:
1. Add top-level-number "GPU Utilization", "Est. SM Efficiency", "Est. Achieved Occupancy" to overall-view, per GPU.
2. Add counters to chrome tracing file, to draw timeline of "GPU Utilization" and ""Est. SM Efficiency", per GPU.

image

image

Run Resnet50, This PR's overhead is 0.77s. The original run without this PR, from "Find run directory" to "Add run" total, cost is 1.89s. So this PR adds about 41% overhead.
90% overhead of this PR is writing json events with counters to temp file. 

After moving these json generating into just before trace is sent to frontend, the additional time overhead at start is reduced to 0.07s. And the additional time overhead to each viewing trace is 0.06s.


tb_plugin/torch_tb_profiler/profiler/overall_parser.py
```diff
@@ -274,6 +332,139 @@ def update_steps_consider_device_side(self, runtime_node_list, device_node_list)
...
+                    counter_json.append(efficiency_json_finish)
+
+        self.blocks_per_sm_per_device = None  # Release memory.
+
+        return counter_json
+
+
+    # Weighted average. Weighted by kernel's time duration.
+    def calculate_occupancy(self):
+        for i in range(len(self.occupancy_per_device)):
```
guotuofeng, line 458: should the for occupancy in occupancy_per_device be enough?

tb_plugin/torch_tb_profiler/profiler/overall_parser.py
```diff
@@ -274,6 +332,139 @@ def update_steps_consider_device_side(self, runtime_node_list, device_node_list)
...
+            util_json["name"] = "GPU {} Est. SM Efficiency".format(gpu_id)
+            util_json["pid"] = 0
+            util_json["ts"] = start_time
+            util_json["args"] = {"Est. SM Efficiency": counter_value}
+            return util_json
+
+        counter_json = []
+        total_dur = self.steps[-1][1] - self.steps[0][0]
+        approximated_sm_efficency_ranges_per_device = []
+        for device_id, i in self.device_to_index.items():
```
guotuofeng, line 437: if the device_ids is list, then for device_id, in in enumerate(self.device_ids): should be enough?

tb_plugin/torch_tb_profiler/profiler/overall_parser.py
```diff
@@ -274,6 +332,139 @@ def update_steps_consider_device_side(self, runtime_node_list, device_node_list)
...
+            util_json["ph"] = "C"
+            util_json["name"] = "GPU {} Est. SM Efficiency".format(gpu_id)
+            util_json["pid"] = 0
+            util_json["ts"] = start_time
+            util_json["args"] = {"Est. SM Efficiency": counter_value}
+            return util_json
+
+        counter_json = []
+        total_dur = self.steps[-1][1] - self.steps[0][0]
+        approximated_sm_efficency_ranges_per_device = []
```
guotuofeng, line 436: useless?

guotuofeng (changes requested):

tb_plugin/torch_tb_profiler/profiler/overall_parser.py
```diff
@@ -274,6 +332,139 @@ def update_steps_consider_device_side(self, runtime_node_list, device_node_list)
                     self.steps = self.steps[:keep_steps]
                     self.steps_names = self.steps_names[:keep_steps]
 
+    def calculate_gpu_utilization(self):
```
guotuofeng, line 335: It seems the calculate_approximated_sm_efficency and calculate_gpu_utilization only depends on the self.steps? if so, we can extract them into a separated class like GpuParser include all these logic with accept the steps as parameters to decouple the logic with current the huge file?

nbcsm:
Need to measure the overhead added to load files.

tb_plugin/torch_tb_profiler/profiler/overall_parser.py
```diff
@@ -29,6 +30,48 @@ def merge_ranges(src_ranges, is_sorted=False):
     return merged_ranges
 
 
+EndpointTypes = IntEnum('EndpointTypes', ['START', 'END'], start=0)
+
+
+class EndPoint(object):
+    def __init__(self):
```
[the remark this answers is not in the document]
gaoteng-git: Thanks. Removed.

tb_plugin/torch_tb_profiler/profiler/overall_parser.py
```diff
@@ -174,6 +217,21 @@ def __init__(self):
         self.steps = []
         self.steps_names = []
 
+        self.device_to_index = {}
```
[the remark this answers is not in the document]
gaoteng-git: Good. The indexing could be removed for simplicity. Now change to a set from dict. Thanks!

tb_plugin/torch_tb_profiler/profiler/overall_parser.py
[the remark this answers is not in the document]
gaoteng-git: Thanks. Done.

tb_plugin/torch_tb_profiler/profiler/overall_parser.py
```diff
@@ -302,12 +497,26 @@ def parse_events(self, events, runtime_node_list, device_node_list):
         for i in range(len(self.avg_costs.costs)):
             self.avg_costs.costs[i] /= valid_steps
 
+        return
```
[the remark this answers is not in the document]
gaoteng-git: Thanks. Done

tb_plugin/torch_tb_profiler/profiler/overall_parser.py
```diff
@@ -302,12 +497,26 @@ def parse_events(self, events, runtime_node_list, device_node_list):
...
+
     def parse_event(self, event):
         ts = event.ts
         dur = event.duration
         evt_type = event.type
         if evt_type == EventTypes.KERNEL:
             self.role_ranges[ProfileRole.Kernel].append((ts, ts + dur))
+            # For caculating GPU utilization, and approximated SM efficiency.
+            device_id = event.args.get("device", None)
+            if device_id not in self.device_to_index:
```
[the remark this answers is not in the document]
gaoteng-git: Maybe not. But I still prefer to add a check to defend. Done.

tb_plugin/torch_tb_profiler/profiler/data.py
```diff
@@ -118,6 +124,18 @@ def process(self):
...
+        self.gpu_utilization = overall_parser.gpu_utilization
+        self.sm_efficency = overall_parser.avg_approximated_sm_efficency_per_device
+        self.occupancy = overall_parser.avg_occupancy_per_device
+        self.trace_json["traceEvents"].extend(overall_parser.gpu_util_json)
+        self.trace_json["traceEvents"].extend(overall_parser.gpu_sm_efficiency_json)
+        fp = tempfile.NamedTemporaryFile('w+t', suffix='.json.gz', delete=False)
+        fp.close()
+        with gzip.open(fp.name, mode='wt') as fzip:
+            fzip.write(json.dumps(self.trace_json))
+        self.trace_file_path = fp.name
```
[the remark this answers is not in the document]
gaoteng-git: ?

tb_plugin/torch_tb_profiler/profiler/overall_parser.py
gaoteng-git: Remove the "i"? Good style. Done. Thanks.

tb_plugin/torch_tb_profiler/profiler/overall_parser.py
gaoteng-git: Thanks. Now change to a set.

tb_plugin/torch_tb_profiler/profiler/overall_parser.py
gaoteng-git: Thanks. Removed.

tb_plugin/torch_tb_profiler/profiler/overall_parser.py
gaoteng-git: Good idea. Done. Thanks!

gaoteng-git:
> Need to measure the overhead added to load files.

Good question. Run Resnet50, This PR's overhead is 0.77s. The original run without this PR, from "Find run directory" to "Add run" total, cost is 1.89s. So this PR adds about 41% overhead. 
May need further optimization.

tb_plugin/torch_tb_profiler/profiler/data.py
```diff
@@ -138,3 +175,83 @@ def analyze(self):
...
+                   "You could try to increase the blocks number of these kernels.".format(
+                gpu_list_str, has_str)
+            self.recommendations.append(text)
+
+        low_occupancy_gpus = []
+        for gpu_id in self.gpu_ids:
+            if self.occupancy[gpu_id] > 0 and self.occupancy[gpu_id] < 50:
+                low_occupancy_gpus.append(gpu_id)
+        if len(low_occupancy_gpus) > 0:
+            gpu_list_str, has_str = get_gpus_str(low_occupancy_gpus)
```
guotuofeng, line 248: suggest to move the gpu related metrics to a separated file and only use the output here.

tb_plugin/torch_tb_profiler/profiler/overall_parser.py
```diff
@@ -2,120 +2,14 @@
...
 from .. import utils
 from .trace import EventTypes
+from .range_lib import *
 
 logger = utils.get_logger()
 
 
-def merge_ranges(src_ranges, is_sorted=False):
-    merged_ranges = []
-    if len(src_ranges) > 0:
```
guotuofeng, line 15: I already move these logic to a separated file, please merge from plugin/0.2

tb_plugin/torch_tb_profiler/profiler/data.py
gaoteng-git: This is the rule based recommendation. We can refactor the recommendation to another file, but maybe not in this PR.

tb_plugin/torch_tb_profiler/profiler/overall_parser.py
gaoteng-git: Yes, merged.

tb_plugin/torch_tb_profiler/profiler/data.py
guotuofeng: could these logic to be moved into a separated method instead?

tb_plugin/torch_tb_profiler/profiler/range_lib.py
```diff
@@ -0,0 +1,144 @@
...
+            if src_ranges[src_id][1] > merged_ranges[dst_id][1]:
+                if src_ranges[src_id][0] <= merged_ranges[dst_id][1]:
+                    merged_ranges[dst_id] = (merged_ranges[dst_id][0], src_ranges[src_id][1])
+                else:
+                    merged_ranges.append(
+                        (src_ranges[src_id][0], src_ranges[src_id][1]))
+    return merged_ranges
+
+
+def pop_list(range_list, index):
```
guotuofeng, line 60: wrong merge since:
1. the data.py is still use old code
2. most of the logic in this file is already in range_utils.py

tb_plugin/torch_tb_profiler/profiler/module_parser.py
```diff
@@ -53,6 +53,7 @@ def __init__(self):
         self.op_list_groupby_name = []  # For Operator-view.
         self.op_list_groupby_name_input = []  # For Operator-view.
         self.kernel_list_groupby_name_op = {}  # For Kernel-view.
+        self.runtime_node_list = []  # For rule based recommendation of gpu metrics.
```
guotuofeng, line 56: this field is moved to nodeparser already

tb_plugin/torch_tb_profiler/profiler/gpu_metrics_parser.py
```diff
@@ -0,0 +1,175 @@
...
+            unit_str = "us"
+            if unit >= 1000:
+                unit /= 1000
+                unit_str = "ms"
+                if unit >= 1000:
+                    unit /= 1000
+                    unit_str = "s"
+            return int(bucket_size), int(buckets), int(unit), unit_str
+
+        def build_trace_counter(gpu_id, start_time, counter_value):
```
guotuofeng, line 48: need the logic be moved to plugin.py to speed up the loading time?

tb_plugin/torch_tb_profiler/plugin.py
```diff
@@ -206,12 +215,32 @@ def kernel_table_route(self, request):
...
+        util_bytes = profile.gpu_util_json
+
+        counter_json_str = ""
+        for gpu_id in range(len(profile.approximated_sm_efficency_ranges)):
+            ranges = profile.approximated_sm_efficency_ranges[gpu_id]
+            for r in ranges:
+                efficiency_json_start = build_trace_counter(gpu_id, r[0][0], r[1])
+                efficiency_json_finish = build_trace_counter(gpu_id, r[0][1], 0)
+                counter_json_str += (efficiency_json_start + efficiency_json_finish)
+        counter_json_bytes = bytes(counter_json_str, 'utf-8')
```
guotuofeng, line 240: in case of the raw data is gz file, I think the GPU SM efficiency and utilization will not be enabled?

gaoteng-git: Cool! It's a big bug. Thanks a lot!
If it is gz file, I have to decompress it firstly, then append these metrics to the end, and compress it, and add an "update" member function in io.Cache to save the compressed result into local file?
BTW, if the original data file is not "gz", maybe we can save the compressed data into local file and update io.Cache for following calls.

tb_plugin/torch_tb_profiler/profiler/gpu_metrics_parser.py
gaoteng-git: Do you mean lazy initialization, delay it to just on calling trace view?

guotuofeng: just curious the reason that one part json generation is in plugin.py and another part is in gpu_metrics_parser.py

tb_plugin/torch_tb_profiler/profiler/module_parser.py
gaoteng-git: Thanks. Fixed.

tb_plugin/torch_tb_profiler/profiler/range_lib.py
gaoteng-git: Thanks. Fixed.

tb_plugin/torch_tb_profiler/profiler/range_utils.py
```diff
@@ -1,3 +1,43 @@
...
+
+
+class EndPoint(object):
+    def __init__(self, ep_time, ep_pt_type, ep_value):
+        self.time = ep_time
+        self.pt_type = ep_pt_type
+        self.value = ep_value
+
+
+# src_ranges: item of ((start_time, end_time), value)
```
guotuofeng, line 14: could the src_ranges to be type tuple(start_time, end_time, value) to remove the embedded extra tuple? In this way, we can write:
for start_time, end_time, value in src_ranges:
    xxxx

tb_plugin/torch_tb_profiler/profiler/gpu_metrics_parser.py
gaoteng-git: Done. Moved to plugin.py, now consistent.

tb_plugin/torch_tb_profiler/profiler/data.py
gaoteng-git: Thanks. Done.

tb_plugin/torch_tb_profiler/profiler/data.py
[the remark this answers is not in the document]
gaoteng-git: Thanks. Called "add_tempfile".

tb_plugin/torch_tb_profiler/profiler/range_utils.py
gaoteng-git: Cool! Done.

tb_plugin/torch_tb_profiler/io/cache.py
```diff
@@ -37,17 +40,31 @@ def __setstate__(self, state):
...
+                        self._cache_dict[filename] = local_file
 
         logger.debug("reading local cache %s for file %s" % (local_file, filename))
         with File(local_file, 'rb') as f:
-            return f.read()
+            return f.read(), filename in self._gpu_metrics_cache_dict
+
+    def write_gpu_metrics(self, raw_data, filename):
+        fp = tempfile.NamedTemporaryFile('w+b', suffix='.json.gz', delete=False)
+        fp.close()
```
guotuofeng, line 60: basically, the cache class has no idea about the gpu metrics. Instead, these logic should be moved out of this cache class.

It should simply has some logic like: add_local_cache(source_file, dest_file), which is called in the plugin.py

guotuofeng: actually, the trace view route doesn't need the cache class at all since a simply self._trace_file in plugin.py is enough for the functionality.

tb_plugin/torch_tb_profiler/profiler/data.py
```diff
@@ -31,13 +33,21 @@ def __init__(self, worker):
         self.data_schema_version = None
         self.events = None
         self.trace_file_path = None
+        self.trace_json = None
```
guotuofeng, line 36: do we still need it?

tb_plugin/torch_tb_profiler/plugin.py
```diff
@@ -206,15 +216,52 @@ def kernel_table_route(self, request):
 
     @wrappers.Request.application
     def trace_route(self, request):
```
guotuofeng, line 218: from the time being, the trace_route include two functionalities: 1), transcode the raw data by adding the gpu utilization content. 2), send the transcode result to the client side. It would be better that if we could make the transcode logic in a separated method or function to make the functionality boundary more clear for maintenance

tb_plugin/torch_tb_profiler/profiler/data.py
gaoteng-git: Thanks! Removed.

tb_plugin/torch_tb_profiler/io/cache.py
gaoteng-git: Sounds good. Thanks. Done.

tb_plugin/torch_tb_profiler/plugin.py
gaoteng-git: Cool. Thanks. Done.

tb_plugin/torch_tb_profiler/plugin.py
```diff
@@ -222,15 +222,19 @@ def trace_route(self, request):
...
+        file_with_gpu_metrics = profile.gpu_metrics_file_mapping.get(profile.trace_file_path)
+        if file_with_gpu_metrics is None:
+            raw_data = self._cache.read(profile.trace_file_path)
+        else:
+            with open(file_with_gpu_metrics, 'rb') as f:
+                raw_data = f.read()
 
         if not profile.has_kernel:  # Pure CPU.
             if not profile.trace_file_path.endswith('.gz'):
                 raw_data = gzip.compress(raw_data, 1)
```
guotuofeng, line 234: move the logic into the if file_with_gpu_metrics. something like:

if file_with_gpu_metrics is None:
    raw_data = self._cache.read(profile.trace_file_path)
    if profile.has_kernel:
        raw_data = self._append_gpu_metrics(profile, raw_data)
    else:
        if not profile.trace_file_path.endswith('.gz'):
            raw_data = gzip.compress(raw_data, 1)
else:
    with open(file_with_gpu_metrics, 'rb') as f:
        raw_data = f.read()

gaoteng-git: Cool! Thanks. Done.

tb_plugin/torch_tb_profiler/profiler/gpu_metrics_parser.py
```diff
@@ -0,0 +1,163 @@
...
+        self.calculate_gpu_utilization(steps_start_time, steps_end_time)
+        self.calculate_approximated_sm_efficency(steps_start_time, steps_end_time)
+        self.calculate_occupancy()
+
+    def parse_event(self, event):
+        ts = event.ts
+        dur = event.duration
+        evt_type = event.type
+        if evt_type == EventTypes.KERNEL:
+            gpu_id = event.args.get("device", None)
```
guotuofeng, line 145: need we add warning in case of pid and deviceId is not same?

tb_plugin/torch_tb_profiler/profiler/gpu_metrics_parser.py
```diff
@@ -0,0 +1,163 @@
...
+            if gpu_id is not None:
+                if gpu_id not in self.gpu_ids:
+                    self.kernel_ranges_per_device.extend([[] for _ in range(gpu_id + 1 - len(self.gpu_ids))])
```
guotuofeng, line 148: what if the gpu_id is in reverse order in events?

tb_plugin/torch_tb_profiler/profiler/gpu_metrics_parser.py
gaoteng-git: Thanks. Done.

tb_plugin/torch_tb_profiler/profiler/gpu_metrics_parser.py
gaoteng-git: Thanks. All changed to length 64.

guotuofeng (approved):

PR #230, opened 2021-05-18
Fix bug with top.ts

TomWildenhain-Microsoft:


tb_plugin/fe/src/utils/top.ts
```diff
@@ -32,9 +32,13 @@ export function useTopN(options?: IOptions) {
     ? React.useCallback(debounce(setActualTop, options.wait ?? 500), [])
     : setActualTop
   React.useEffect(() => {
-    setActualDebounce(
-      useTop === UseTop.Use && topIsValid(topText) ? Number(topText) : actualTop
```
guotuofeng, line 36: could you please add comment about what's the bug?

TomWildenhain-Microsoft: Previously useTop wasn't being used to determine whether undefined should be returned.  undefined indicates that there is no maximum number of ops specified.  Setting the radio button to use all ops didn't have the correct effect.

PR #231, opened 2021-05-18
refactor io.cache to support general file cache

guotuofeng:
refactor io.cache to make its responsibility more clear boundary for resuse


tb_plugin/torch_tb_profiler/plugin.py
```diff
@@ -479,10 +409,28 @@ def _get_run_name(self, run_dir):
     def _get_profile(self, name, worker):
         run = self._get_run(name)
         profile = run.get_profile(worker)
+        if profile is None:
+            e = errors.NotFoundError("could not find the profile for %s/%s " %(name, worker))
+            e.headers.extend(TorchProfilerPlugin.headers)
+            print("_get_profile", e)
```
guyang3532, line 415: this print is for debug?

guotuofeng: good catch

PR #241, opened 2021-05-24
refine gpu metrics trace view

gaoteng-git:
1. Add baseline counter with value 1 to trace json, at the beginning of profiling. For easy to see the absolute values.
2. Use the time range "PyTorch Profiler (0)" to make buckets for GPU utilization's timeline. If it doesn't exist, use global wise min and max time as it.
3. Make GPU utilization's last bucket reach the end of profiling, maybe longer than bucket_size.
4. Add unit test for parsing the string with gpu metrics appended as valid json.


tb_plugin/test/test_profiler.py
```diff
@@ -1043,6 +1048,11 @@ def test_dump_gpu_metrics(self):
             data_expected = file.read()
         self.assertEqual(data_with_gpu_metrics_flat, data_expected)
```
guotuofeng, line 1049: the compare could fail if we manually change the file but the content is same?

tb_plugin/torch_tb_profiler/run.py
```diff
@@ -82,17 +82,31 @@ def build_trace_counter_sm_efficiency(gpu_id, start_time, counter_value):
             )
             return util_json
 
+        def add_trace_counter_gpu_util(gpu_id, start_time, counter_value):
+            json_str = build_trace_counter_gpu_util(gpu_id, start_time, counter_value)
+            counter_json_list.append(json_str)
```
guotuofeng, line 87: pass by parameter to make it clear so that this method could be move outside of current method.

tb_plugin/torch_tb_profiler/profiler/gpu_metrics_parser.py
```diff
@@ -77,19 +86,20 @@ def get_bucket_info(range_micro_seconds):
...
                             current_bucket_index += 1
-                            current_bucket = (steps_start_time + current_bucket_index * bucket_size,
-                                              steps_start_time + (current_bucket_index + 1) * bucket_size)
+                            current_bucket = buckets_ranges[current_bucket_index] if current_bucket_index < buckets \
+                                else None
                         else:
                             current_range_index += 1
                             if current_range_index < len(self.kernel_ranges_per_device[gpu_id]):
                                 current_range = self.kernel_ranges_per_device[gpu_id][current_range_index]
                 for i_bucket in range(buckets):
```
guotuofeng, line 95: could the two loops be merged?

tb_plugin/torch_tb_profiler/profiler/trace.py
```diff
@@ -9,7 +9,7 @@
 
 
 class EventTypes(object):
-    NET = "Net"
```
guotuofeng, line 12: would this change ignore all events whose category is Net?

tb_plugin/test/test_profiler.py
guotuofeng: https://stackoverflow.com/questions/25851183/how-to-compare-two-json-objects-with-the-same-elements-in-a-different-order-equa

gaoteng-git: Got it. Done. Thanks.

tb_plugin/torch_tb_profiler/run.py
gaoteng-git: Good. Done. Thanks.

tb_plugin/torch_tb_profiler/profiler/gpu_metrics_parser.py
gaoteng-git: I prefer uses 2 separate loop because the later one has some processing code out of loop. And the buckets are less than 100, little perf. overhead.

tb_plugin/torch_tb_profiler/profiler/trace.py
gaoteng-git: The plugin has no use of NET.

tb_plugin/torch_tb_profiler/profiler/gpu_metrics_parser.py
guotuofeng: it would be better that we can add some comments about the reason why uses two loops here.

gaoteng-git: Well, merged into one loop.

guotuofeng (approved):

PR #242, opened 2021-05-24
improve data load performance by using parallel loading

guotuofeng:


tb_plugin/torch_tb_profiler/profiler/loader.py
```diff
@@ -42,19 +32,62 @@ def _parse(self):
...
-                data = RunProfileData.parse(self.run.run_dir, worker, path, self.caches)
-                self.run.profiles[worker] = data
-            except Exception as ex:
+            p = Process(target=self._process_data, args=(worker, path, barrier))
+            p.start()
+
+        logger.info("starting all processing")
+        # since there is one queue, its data must be read before join.
+        # https://stackoverflow.com/questions/31665328/python-3-multiprocessing-queue-deadlock-when-calling-join-before-the-queue-is-em
+        barrier.wait()
```
gaoteng-git, line 43: How about just counting the queue.get() and exiting loop on reaching "workers"? Then Barrier could be removed.

guotuofeng: We need handle some worker process return nothing.

PR #262, opened 2021-05-28
Kernel count bug fix

gaoteng-git:
Remove the filter and enable duration 0 kernel to count:
1. Add exception handling inside "weighted_avg".
2. KernelAggByNameOp's property "avg_blocks_per_sm" and "avg_occupancy" already handles divided by 0.


tb_plugin/torch_tb_profiler/profiler/kernel_parser.py
```diff
@@ -20,9 +20,15 @@ def parse_events(self, events):
         events = pd.DataFrame(events)
         events = events.astype({"type": "category", "category": "category", "name": "string"}, copy=False)
         kernels = events[events["category"] == "Kernel"]
```
guotuofeng, line 22: could we merge this line with event.category == "Kernel" line by only adding kernel events?

gaoteng-git: Thanks. Done.

guotuofeng (approved):

PR #265, opened 2021-05-28
Add gpu info

gaoteng-git:
1. Add GPU device info.
2. Pass title as “GPU Summary” to UI.
3. Change "------" to "`<hr/>`"

Corresponding front-end PR: https://github.com/pytorch/kineto/pull/264

image

Note: The 1st and 2nd V100 have different memory size, interesting....
image
image


tb_plugin/torch_tb_profiler/run.py
```diff
@@ -124,19 +125,50 @@ def append_gpu_metrics(self, raw_data):
         return raw_data
 
 
+    def get_gpu_info(self, gpu_id):
```
guotuofeng, line 128: could we reuse the _generate_gpu_info in run_generator.py?

guotuofeng: since most of the code same with that function

gaoteng-git: Reasonable. Should reuse. Done. Thanks!

tb_plugin/torch_tb_profiler/run.py
```diff
@@ -130,13 +131,22 @@ def get_gpu_metrics_data(profile):
...
             for gpu_id in profile.gpu_ids:
                 if not is_first:
                     # Append separator line for beautiful to see.
-                    gpu_metrics_data.append({"title": "--------",
+                    gpu_metrics_data.append({"title": "<hr/>",
                                              "value": ""})
+
                 gpu_metrics_data.append({"title": "GPU {}:".format(gpu_id),
                                          "value": ""})
+                gpu_info = profile.gpu_infos.get(gpu_id, None)
```
guotuofeng, line 143: need we check the gpu_infos be null or not?

guotuofeng (approved):

tb_plugin/torch_tb_profiler/run.py
gaoteng-git: Don't need in this case. Initialized as {} in run_generator.py

PR #298, opened 2021-06-16
remove the Record Window End event

guotuofeng:
remove the Record Window End event if the duration is larger than 1 day to workaround the high resolution clock bug in pytorch profiler

### DIFF
--- a/tb_plugin/torch_tb_profiler/profiler/data.py
+++ b/tb_plugin/torch_tb_profiler/profiler/data.py
@@ -6,7 +6,6 @@ import io as sysio
 import json
 import re
 import tempfile
-from collections import OrderedDict
 from json.decoder import JSONDecodeError
 
 from .. import io, utils
@@ -105,6 +104,7 @@ class RunProfileData(object):
         if trace_path.endswith('.gz'):
             data = gzip.decompress(data)
 
+        json_reencode = False
         try:
             trace_json = json.loads(data)
         except JSONDecodeError as e:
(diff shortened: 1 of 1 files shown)


guotuofeng:
https://github.com/pytorch/pytorch/pull/60070

tb_plugin/torch_tb_profiler/profiler/data.py
```diff
@@ -118,12 +118,33 @@ def _preprocess_file(caches, trace_path):
...
+            if event_list[i]["name"] == "Record Window End":
+                end_index = i
+            elif event_list[i]["name"].startswith("Iteration Start:"):
+                start_index = i
+            if start_index is not None and end_index is not None:
+                break
+
+        if start_index is not None and end_index is not None:
+            dur = event_list[end_index]["ts"] - event_list[start_index]["ts"]
+            print(dur)
```
gaoteng-git, line 138: Replace this print with log?

tb_plugin/torch_tb_profiler/profiler/data.py
```diff
@@ -118,12 +118,33 @@ def _preprocess_file(caches, trace_path):
...
                     trace_json = json.loads(fout.getvalue())
-
+                    logger.warning("Get JSONDecodeError: %s, Re-encode it to temp file" % e.msg)
+                    json_reencode = True
+
+        # work-around to remove the "Record Window End" events to avoid the huge end timestamp
+        event_list = trace_json["traceEvents"]
+        end_index = None
+        start_index = None
+        for i in reversed(range(len(event_list))):
```
gdankel, line 128: Is there not a better place to do this in some existing iteration to avoid extra load latency? Or is it fast enough?

gdankel: I guess it's quickly found since you search backwards.

tb_plugin/torch_tb_profiler/profiler/data.py
```diff
@@ -118,12 +118,33 @@ def _preprocess_file(caches, trace_path):
...
+            if event_list[i]["name"] == "Record Window End":
+                end_index = i
+            elif event_list[i]["name"].startswith("Iteration Start:"):
```
gdankel, line 131: Are you sure Iteration Start is also a problem? Looking at the code, the timestamp for this is actually copied from profiler_kineto.

gdankel (changes requested):

tb_plugin/torch_tb_profiler/profiler/data.py
guotuofeng: removed.

tb_plugin/torch_tb_profiler/profiler/data.py
guotuofeng: yes, the last two items will be the "Iteration Start: PyTorch Profiler" and "Record Window End" instant events.

tb_plugin/torch_tb_profiler/profiler/data.py
guotuofeng: I just want to find the start and end timestamp to remove the end timestamp in case of the duration is larger than one day. I don't remove the Iteration Start.

gdankel: Ah yes - thanks for clarifying.

gdankel (approved):

guotuofeng:
@gdankel, please hold on to merge the PR since we might need create another branch for 0.2.1 release.

PR #306, opened 2021-06-17
Update readme

gaoteng-git:


tb_plugin/docs/gpu_utilization.md
```diff
@@ -1,15 +1,15 @@
...
+* Est. SM Efficiency: Estimated Stream Multiprocessor Efficiency. The higher, the better. This metric of a kernel, SM_Eff_K = min(blocks of this kernel / SM number of this GPU, 100%). 
                       This overall number is the sum of all kernels' SM_Eff_K weighted by kernel's execution duration, divided by “all steps time”. 
                       It shows GPU Stream Multiprocessors’ utilization. 
                       Although it is finer grained than above “GPU Utilization”, it still can’t tell the whole story. 
                       For example, a kernel with only one thread per block can’t fully utilize each SM. 
 
-* Est. Achieved Occupancy: The bigger, the better. The definition of occupancy is [here](https://docs.nvidia.com/gameworks/content/developertools/desktop/analysis/report/cudaexperiments/kernellevel/achievedoccupancy.htm). 
+* Est. Achieved Occupancy: For most cases such as memory bandwidth bounded kernels, the higher the better. The definition of occupancy is [here](https://docs.nvidia.com/gameworks/content/developertools/desktop/analysis/report/cudaexperiments/kernellevel/achievedoccupancy.htm).
```
cloudhan, line 12: lets simply call it `memory bounded` instead of `memory bandwidth bounded`

cloudhan: I am actually not quite sure how to make it clear in the case of occupancy. The core logic behind occupancy is to over-subscribe the GPU to hide the latency, since intra-warp context switch is quite fast. Once target occupancy is achieved, it does not make it better after that.

See page 13 of [GPU Performance Analysis and Optimization](http://developer.download.nvidia.com/GTC/PDF/GTC2012/PresentationPDF/S0514-GTC2012-GPU-Performance-Analysis.pdf)

I think we should remove the whole sentence here.

gaoteng-git: Memory bounded could be interpreted as size bounded or bandwidth bounded. So I prefer to add "bandwidth".

cloudhan (approved):

gaoteng-git:
@cloudhan Cool reference URL. I've added it after this sentence as Reference. Giving user more info to judge is good.

Issue #308, opened 2021-06-17
Speed of profiling


nasretdinovr:
It takes about 30min to 1 hour to profile my network (transformer-like architecture, about 20 million parameters). Is it normal or I do something wrong?

gaoteng-git:
@nasretdinovr How many steps do you profile? What's the arguments of torch.profiler.profile? Would you please share your profiling code here?

nasretdinovr:
@gaoteng-git I took the code from ResNet example in PyTorch doc and adapted it for my network. To profile ResNet it takes about 5 min. 

```
with torch.profiler.profile(
     activities=[
        torch.profiler.ProfilerActivity.CPU,
        torch.profiler.ProfilerActivity.CUDA],
    schedule=torch.profiler.schedule(
        wait=2,
        warmup=3,
        active=6),
    on_trace_ready=torch.profiler.tensorboard_trace_handler('./result', worker_name='worker0'),
    record_shapes=True,
    profile_memory=True,
    with_stack=True
) as p:
    for batch_idx, data in enumerate(train_loader, 0):
         print("step:{}".format(batch_idx))

        spectrograms, labels, input_lengths, label_lengths = data
        spectrograms, labels = spectrograms[:, :, :, :max(input_lengths)].to(device), labels.to(
            device)  # (batch, 1, feat_dim, time)
        spectrograms = spectrograms.squeeze(1).transpose(1, 2)  # (batch, time, feat_dim,)
        optimizer.zero_grad()

        output = model(spectrograms, input_lengths)  # (batch, time, n_classes)
        output = F.log_softmax(output, dim=2)
        output = output.transpose(0, 1)  # (time, batch, n_class)
        input_lengths = [x // 4 for x in input_lengths]
        loss = criterion(output, labels, input_lengths, label_lengths)
        loss.backward()

        torch.nn.utils.clip_grad_norm_(model.parameters(), 5.0)
        optimizer.step()

        if batch_idx + 1 >= 22:
            break
        p.step()
```

gaoteng-git:
@nasretdinovr You could try disable "profile_memory" by setting it to False or just removing it.
The current PyTorch profiler's memory events handling is low efficient and very slow. We will contact Facebook to optimize it.

nasretdinovr:
@gaoteng-git Disabling "profile_memory" really helped! Thank you a lot!

guotuofeng:
[...]

PR #316, opened 2021-06-20
Workaround: update example to speedup

gaoteng-git:
The original example takes about 6 minutes at the end of 6 active steps. And 12 minutes for the 12steps in total. This is too long for user to wait.
The cause is "profile_memory=True". In PyTorch code, this [line](https://github.com/pytorch/pytorch/blob/c19acf816f984124bb00fa7d32f4376d2d76c331/torch/autograd/profiler.py#L1136) and this [line](https://github.com/pytorch/pytorch/blob/c19acf816f984124bb00fa7d32f4376d2d76c331/torch/autograd/profiler.py#L1147) leads to O(M*N) time complexity, M is op events count and N is memory events count. We should optimize it.

Community user also complains this: https://github.com/pytorch/kineto/issues/308
We should fix it quickly.
Here I just add a comment and reduce to 2 active steps to reduce profiling time to 40s.

Before fix, overhead is about 360s * 2:
image
After fix, overhead is about 40s:
image

### DIFF
--- a/tb_plugin/examples/resnet50_profiler_api.py
+++ b/tb_plugin/examples/resnet50_profiler_api.py
@@ -31,12 +31,12 @@ with torch.profiler.profile(
         torch.profiler.ProfilerActivity.CPU,
         torch.profiler.ProfilerActivity.CUDA],
     schedule=torch.profiler.schedule(
-        wait=2,
-        warmup=3,
-        active=6),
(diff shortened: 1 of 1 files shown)


gdankel:
Yes this is not good. Let's fix it.

guotuofeng (approved):

Issue #321, opened 2021-06-24
Seems like with tensorboard 1.15, there will be an ImportError: cannot import name 'errors'


BruceDai003:
https://github.com/pytorch/kineto/blob/ba8bdffc5e63958d89dfc78e8a671c7504a3593d/tb_plugin/setup.py#L29

I have pre-installed tensorflow 1.15 and tensorboard 1.15. And I just installed torch-tb-profiler=0.2.0.
But when I try to call tensorboard now, I got an error:
```
root@298562e873aa:/opt/sw_home/framework-tests/pt/resnet# tensorboard --logdir=./log --host=10.201.40.22
Traceback (most recent call last):
  File "/usr/local/bin/tensorboard", line 8, in <module>
    sys.exit(run_main())
  File "/usr/local/lib/python3.6/dist-packages/tensorboard/main.py", line 58, in run_main
    default.get_plugins() + default.get_dynamic_plugins(),
  File "/usr/local/lib/python3.6/dist-packages/tensorboard/default.py", line 110, in get_dynamic_plugins
    for entry_point in pkg_resources.iter_entry_points('tensorboard_plugins')
  File "/usr/local/lib/python3.6/dist-packages/tensorboard/default.py", line 110, in <listcomp>
    for entry_point in pkg_resources.iter_entry_points('tensorboard_plugins')
  File "/usr/local/lib/python3.6/dist-packages/pkg_resources/__init__.py", line 2450, in load
    return self.resolve()
  File "/usr/local/lib/python3.6/dist-packages/pkg_resources/__init__.py", line 2456, in resolve
    module = __import__(self.module_name, fromlist=['__name__'], level=0)
  File "/usr/local/lib/python3.6/dist-packages/torch_tb_profiler/plugin.py", line 17, in <module>
    from tensorboard import errors
ImportError: cannot import name 'errors'
```
Seems like tensorboard 1.15.0 version doesn't have errors.py module, also I just confirmed this in the tensorboard github repo. So,  does this mean I have to upgrade to tensorboard 2.0?

guotuofeng:
It is our bug, will fix it.

PR #323, opened 2021-06-25
Fix bugs happens with tensorboard 1.15

guotuofeng:
Fix bugs happens with tensorboard 1.15 related to import errors which is available in Tensorboard 2.0

### DIFF
--- a/tb_plugin/torch_tb_profiler/plugin.py
+++ b/tb_plugin/torch_tb_profiler/plugin.py
@@ -14,9 +14,8 @@ from collections import OrderedDict
 from queue import Queue
 
 import werkzeug
-from tensorboard import errors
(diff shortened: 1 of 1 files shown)


guotuofeng:
https://github.com/pytorch/kineto/issues/321

gaoteng-git:
Tested on real case by triggering the throw of "exceptions.NotFound“ and "exceptions.BadRequest"?

guotuofeng:
@gaoteng-git , the issue is that we need use tensorboard 1.15. The CI pipeline would install the latest version. The bug would be caught if the tensorboard 1.15 is installed in CI pipeline regardless whether the exception is raise or not. The error happens when import the error.

gaoteng-git (approved):

Issue #270, opened 2021-06-03
No Empty State for TB Plugin


Reubend:
If you load the TB plugin on a log directory that _**does not**_ have any traces inside, it will display a loading spinner endlessly. I think there should be a message saying that no traces were loaded; other plugins (like the graph plugin, hparams plugin, images plugin, etc.) behave by displaying such a message. 

Expected behavior:
the TB plugin should say that no traces were found
Actual behavior:
A spinner shows up but no error message ever shows

<img width="1424" alt="Screen Shot 2021-06-02 at 9 00 14 PM" src="https://user-images.githubusercontent.com/13208038/120570193-8c144d80-c3e5-11eb-8dc3-aaf66db1fa38.png">

xilin1991:
I also encountered the same problem。

xilin1991:
I solved this problem in my environment. You can uninstall 'torch-tb-profiler' in your virtual environment(such as anaconda).

Reubend:
> I solved this problem in my environment. You can uninstall 'torch-tb-profiler' in your virtual environment(such as anaconda).

Maybe I don't understand fully, but I don't want to uninstall it. I just want it to show an empty state message when there are no traces found.

xilin1991:
Do you want to see the results of the model in Tensorboard dashboard?
Tensorboard can work even after uninstalling 'torch-tb-profiler' in my environment.
During this process, there will be a warning in the following picture.
[...]

PR #326, opened 2021-06-26
tb_plugin: fix startup hang when no profile data

mautier:
When starting up tensorboard and the plugin in a directory where there
is no profile data, `TorchProfilerPlugin.is_active()` hangs forever. This
makes the tensorboard server hang, and client-side webpage remain blank.

This commit fixes the issue by ensuring that the thread monitoring the
run directories for data notifies the `_is_active_initialized_event`
flag not only when data is found, but also when the search is complete
and no data was found.

A simple unit test is added which validates the fix: without this
commit, the test hangs indefinitely.

### DIFF
--- a/tb_plugin/test/test_plugin.py
+++ b/tb_plugin/test/test_plugin.py
@@ -1,0 +1,15 @@
+import unittest
+
+from tensorboard.plugins.base_plugin import TBContext
+
+from torch_tb_profiler.plugin import TorchProfilerPlugin
+
+
+class TestPlugin(unittest.TestCase):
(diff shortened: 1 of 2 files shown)


guotuofeng:
This is a known issue, https://github.com/pytorch/kineto/issues/270. We will think about it and gave a symmetrical fix.

guotuofeng:
The fix is no exhaustive. will close it.

guotuofeng:
https://github.com/pytorch/kineto/pull/340/ should fix the issue

mautier:
Awesome thank you!

PR #327, opened 2021-06-26
tb_plugin: follow symlinks when looking for runs

mautier:
This commit makes `walk`-ing directories follow symlinks when searching
for run data (on filesystems that support it!).

This makes the plugin's search behavior consistent with that of
tensorboard itself; using symlink trees to organize runs is one of the
recommendations made in the tensorboard docs to have fine-grained
control over the naming of runs and the location of the data [(link)](https://github.com/tensorflow/tensorboard/blob/master/README.md#logdir--logdir_spec-legacy-mode):

> TensorBoard walks log directories recursively; for finer-grained
> control, prefer using a symlink tree.

A unit test is added to validate the new behavior.

Addresses issue https://github.com/pytorch/kineto/issues/331

### DIFF
--- a/tb_plugin/test/test_tensorboard_end2end.py
+++ b/tb_plugin/test/test_tensorboard_end2end.py
@@ -1,12 +1,18 @@
 import json
 import os
+import shutil
 import socket
+import tempfile
 import time
 import unittest
 import urllib
 import urllib.request
 from subprocess import Popen
 from urllib.error import HTTPError
+
+
+def get_samples_dir():
+    return os.path.join(os.path.dirname(os.path.abspath(__file__)), '../samples')
 
 
 class TestEnd2End(unittest.TestCase):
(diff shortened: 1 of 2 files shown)


tb_plugin/torch_tb_profiler/plugin.py
```diff
@@ -346,7 +346,11 @@ def _get_run_dirs(self):
             /run2
                 /[worker1].pt.trace.json
         """
-        for root, _, files in io.walk(self.logdir):
+        # Note on followlinks=True: per the tensorboard documentation [1], users are encouraged to
+        # use symlink trees to have fine-grained control over the filesystem layout of runs. To
+        # support such trees, we must follow links.
+        # [1] https://github.com/tensorflow/tensorboard/blob/master/README.md#logdir--logdir_spec-legacy-mode
+        for root, _, files in io.walk(self.logdir, followlinks=True):
```
guotuofeng, line 353: It seems only need change the os.walk call since the cloud storage doesn't have the symlink concept?

mautier: Indeed that would make this change much less invasive! I wasn't sure if it was acceptable to enable `followlinks` all the time without giving a choice to the caller; but it appears that currently there's only one caller anyway.

I'll make the change!

tb_plugin/test/test_tensorboard_end2end.py
```diff
@@ -17,24 +23,44 @@ class TestEnd2End(unittest.TestCase):
...
+
+        # Desired layout:
+        # logdir/
+        #     run_A/ --> path/to/samples/resnet50_num_workers_0/
+        #     run_B/ --> path/to/samples/resnet50_num_workers_4/
+        for target, link_name in [
+            ("resnet50_num_workers_0", "run_A"),
+            ("resnet50_num_workers_4", "run_B"),
+        ]:
+            os.symlink(os.path.join(samples_dir, target), os.path.join(logdir, link_name))
```
guotuofeng, line 57: Do we need mix both symlinks and ordinary files together to test both the mix cases?

guotuofeng: BTW, would you please create an issue to track this change?

mautier: Done for both! https://github.com/pytorch/kineto/issues/331

guotuofeng:
@mautier, would you please merge from plugin/0.2 so to remove the unnecessary changes in commit history?

mautier:
@guotuofeng Done!

Issue #130, opened 2021-03-24
Profiler plugin does not refresh traces


lenisha:
Running training multiple times generates more traces, but refresh button does not re-scan the log directory and does not show new traces

![image](https://user-images.githubusercontent.com/6332907/112351299-25b1d700-8ca0-11eb-89fd-96def7d80bef.png)

Kingwl:
https://github.com/tensorflow/tensorboard/issues/4829

Kingwl:
Latest tensorboard has support refresh now. https://github.com/tensorflow/tensorboard/issues/4829#issuecomment-819917959

gdankel:
@Kingwl can we close this?

PR #335, opened 2021-07-05
plugin: Fix hanging loading when no file exists

### LINKED ISSUE #334: No results loaded with torch-tb-profiler when there isn't a profiler in the code
I have installed torch-tb-profiler and, when my code includes a profiler, the results of my experiments are displayed correctly in Tensorboard.
Example:
image

However, if my code doesn't use a profiler, nothing is displayed and the spinner icon spins indefinately:

image

If I uninstall torch-tb-profiler, tensorboard again works as intended (though, of course, I can no longer view the profiler outcomes from experiments that do use it).

Is there a way to overcome this and allow for the visualization of both experiments that do and don't use a profiler?

System Information:

- Ubuntu 20.04
- Anaconda environment

Relevant package versions:

- pytorch 1.8.1
- tensorboard 2.5.0
- tensorboard-plugin-profile 2.4.0
- torch-tb-profiler 0.2.0

manuyavuz:
Previous code was locking the requests from tb's `is_active` method until `_is_active` is set.
However, it was also missing setting it to false when there is no related trace files.

This fix basically sets the value to false if generator returns empty list, which will unlock tb request and return false.

Fixes #334, tensorflow/tensorboard#5076, tensorflow/tensorboard#5111.

### DIFF
--- a/tb_plugin/torch_tb_profiler/plugin.py
+++ b/tb_plugin/torch_tb_profiler/plugin.py
@@ -290,7 +290,6 @@ class TorchProfilerPlugin(base_plugin.TBPlugin):
 
     def _monitor_runs(self):
         logger.info("Monitor runs begin")
-
         try:
             touched = set()
             while True:
@@ -311,6 +310,10 @@ class TorchProfilerPlugin(base_plugin.TBPlugin):
                             # Use threading to avoid UI stall and reduce data parsing time
                             t = threading.Thread(target=self._load_run, args=(run_dir,))
                             t.start()
+                    # Do not forget to set _is_active if no file exists
+                    if len(touched) == 0 and not self._is_active:
+                        self._is_active = False
+                        self._is_active_initialized_event.set()
                 except Exception as ex:
                     logger.warning("Failed to scan runs. Exception=%s", ex, exc_info=True)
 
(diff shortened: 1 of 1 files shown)


guotuofeng:
This is similar with https://github.com/pytorch/kineto/pull/326. The fix is conflict with https://github.com/pytorch/kineto/issues/130

guotuofeng:
similar fix https://github.com/pytorch/kineto/pull/326/files

manuyavuz:
@guotuofeng thanks for the feedback.

Indeed it was the same fix with #326, sorry for the dupe.

I understand that team wants to handle dynamic changes in the logdir after the plugin starts more properly (deletions do not seem to be handled right now for instance). I think that threading Event lock can even be completely removed with a better architecture. However, it might make this issue linger for a much longer period of time. Do you have an ETA for when would team implement the new exhaustive solution that would also fix this issue?

It seems to me that this fix or #326 can act as a minimal hotfix change for a quick 0.2.1 release according to the project's release process. What do you think?

guotuofeng:
If the is_active is set to false and there is some available data later, we have to manually press F5 in browser to allow the plugin be shown. 

On another hand, we are considering to make the is_active always return true to fix this issue and we don't have time to fully test it. 
Since the severity/priority of this issue is not so much high, we plan to fix this issue in next release after 0.2.1.

manuyavuz-pointr:
I recognized the need for force refresh in my testing too.

Also it makes much more sense to have an always active plugin and handle no data gracefully.

I can live with this for now, thanks for the information!

guotuofeng:
@manuyavuz, the https://github.com/pytorch/kineto/pull/340/ should be able to fix the bug.

PR #337, opened 2021-07-07
Move concurrency into contexts

Reubend:
Differential Revision: D29458946

Currently, the TB Plugin runs several operations using the Python multiprocessing module. When that module starts new processes, it uses the `TORCH_PROFILER_START_METHOD` flag to see which method it should use. However, it uses that flag to **globally** set the start method, which messes up other plugins who rely on using the OS default. 

This commit fixes this, by running concurrent operations inside of contexts. This way, the start method for profiler operations can be set *without affecting other plugins*.

### DIFF
--- a/tb_plugin/torch_tb_profiler/io/file.py
+++ b/tb_plugin/torch_tb_profiler/io/file.py
@@ -80,6 +80,16 @@ def get_filesystem(filename):
         raise ValueError("No recognized filesystem for prefix %s" % prefix)
     return fs
 
+# When subprocesses are started without state, they won't retain
+# the filesystem dictionary from the main process. Therefore,
+# we define these helpers functions to restore the mapping.
+def get_all_filesystems():
+    return _REGISTERED_FILESYSTEMS
+
+def restore_filesystems(fs_dictionary):
+    global _REGISTERED_FILESYSTEMS
+    _REGISTERED_FILESYSTEMS = fs_dictionary
+
 
 class LocalFileSystem(LocalPath, BaseFileSystem):
     def __init__(self):
--- a/tb_plugin/torch_tb_profiler/profiler/multiprocessing.py
+++ b/tb_plugin/torch_tb_profiler/profiler/multiprocessing.py
@@ -1,0 +1,16 @@
+# When some filesystems launch, they need to be in a process
+# launched by a certain start method. Here, we provide utility
+# functions to set a start method for only this plugin's
+# subprocesses, without disrupting the start method used in
+# other plugins.
+
+import os
+import multiprocessing as mp
+
+def KinetoContext():
+    # Default to the operating system's start method,
+    # and keep the method as a static attribute
+    method = mp.get_start_method(False)
+    if os.getenv('TORCH_PROFILER_START_METHOD'):
+        method = os.getenv('TORCH_PROFILER_START_METHOD')
+    return mp.get_context(method)
--- a/tb_plugin/torch_tb_profiler/io/cache.py
+++ b/tb_plugin/torch_tb_profiler/io/cache.py
@@ -1,22 +1,23 @@
 # -------------------------------------------------------------------------
 # Copyright (c) Microsoft Corporation. All rights reserved.
 # -------------------------------------------------------------------------
-import multiprocessing as mp
 import os
 
 from .. import utils
 from .file import download_file, read
+from ..profiler.multiprocessing import KinetoContext
 
 logger = utils.get_logger()
 
 class Cache:
     def __init__(self):
-        self._lock = mp.Lock()
-        self._manager = mp.Manager()
+        ctx = KinetoContext()
+        self._lock = ctx.Lock()
+        self._manager = ctx.Manager()
         self._cache_dict = self._manager.dict()
 
     def __getstate__(self):
-        '''The multiprocessing module can start one of three ways: spawn, fork, or forkserver. 
+        '''The multiprocessing module can start one of three ways: spawn, fork, or forkserver.
         The default mode is fork in Unix and spawn on Windows and macOS.
         Therefore, the __getstate__ and __setstate__ are used to pickle/unpickle the state in spawn mode.
         '''
@@ -29,8 +30,8 @@ class Cache:
         return data
 
     def __setstate__(self, state):
-        '''The default logging level in new process is warning. Only warning and error log can be written to 
-        streams. 
+        '''The default logging level in new process is warning. Only warning and error log can be written to
+        streams.
         So, we need call use_absl_handler in the new process.
         '''
         from absl import logging
@@ -73,4 +74,3 @@ class Cache:
     def __exit__(self, exc_type, exc_value, traceback):
         self.close()
         self._manager.__exit__(exc_type, exc_value, traceback)
-
--- a/tb_plugin/torch_tb_profiler/plugin.py
+++ b/tb_plugin/torch_tb_profiler/plugin.py
@@ -4,7 +4,6 @@
 import atexit
 import gzip
 import json
-import multiprocessing as mp
 import os
 import sys
 import tempfile
@@ -44,9 +43,7 @@ class TorchProfilerPlugin(base_plugin.TBPlugin):
           context: A base_plugin.TBContext instance.
         """
         super(TorchProfilerPlugin, self).__init__(context)
-        start_method = os.getenv('TORCH_PROFILER_START_METHOD')
-        if start_method:
-            mp.set_start_method(start_method, force=True)
+
         self.logdir = io.abspath(context.logdir.rstrip('/'))
 
         self._load_lock = threading.Lock()
--- a/tb_plugin/torch_tb_profiler/profiler/loader.py
+++ b/tb_plugin/torch_tb_profiler/profiler/loader.py
@@ -5,12 +5,13 @@ import bisect
 import os
 import sys
 from collections import defaultdict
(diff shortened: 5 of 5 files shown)


tb_plugin/torch_tb_profiler/io/file.py
```diff
@@ -80,6 +80,16 @@ def get_filesystem(filename):
         raise ValueError("No recognized filesystem for prefix %s" % prefix)
     return fs
 
+# When subprocesses are started without state, they won't retain
```
guotuofeng, line 83: what's the reason that need this change? I tried locally by using forkserver start method and there is no need to backup and restore the file systems.

tb_plugin/torch_tb_profiler/profiler/multiprocessing.py
```diff
@@ -0,0 +1,27 @@
+# When some filesystems launch, they need to be in a process
+# launched by a certain start method. Here, we provide utility
+# functions to set a start method for only this plugin's
+# subprocesses, without disrupting the start method used in
+# other plugins.
+
+import os
+import multiprocessing as mp
+
+class KinetoContext():
```
guotuofeng, line 10: would it better to add only the following function instead of the static class?

def get_start_method():
    if os.getenv('TORCH_PROFILER_START_METHOD'):
        return os.getenv('TORCH_PROFILER_START_METHOD')
    else:
        return None

then we only need call the mp.get_context by using the returned start method without need to save the global start method?

guotuofeng:
I suppose https://github.com/pytorch/kineto/pull/344 should be enough. No?

tb_plugin/torch_tb_profiler/io/file.py
Reubend: When the subprocess is started using `forkserver`, it will retain any filesystems that are added inside of `file.py` because those lie inside of its import paths. So things like Azure, GCP, S3, etc... will still work.

However, if a filesystem is added using `register_filesystem()`, then that filesystem will be lost during the initialization, because the subprocess doesn't retain state from its parent. This change is necessary to retain filesystems which were added dynamically.

tb_plugin/torch_tb_profiler/profiler/multiprocessing.py
Reubend: No, this allows us to set the start method dynamically instead of only using the environment flag. Specifically, it lets us call `set_profiler_start_method()` during runtime, so that we can determine the correct start method based on the log directories supplied.

However, this class is still compatible with the environment flag, so it works either way. The reason we need a full class is to store values that were set explicitly by that function.

Reubend:
> I suppose #344 should be enough. No?

No, it doesn't include a few changes which are necessary for our use case. Please see me replies above for more detail.

tb_plugin/torch_tb_profiler/io/file.py
guotuofeng: From the time being, the register_filesystem is only called in file.py. Do you have cases that call register_filesystem? If no, I don't suggest this change.

tb_plugin/torch_tb_profiler/profiler/multiprocessing.py
guotuofeng: If you need set the start method dynamically, just pass the start_method to the multiprocessing.get_context is fine. I don't think we need maintain the global state

guotuofeng: I think set the start method by environment is enough. the set_profiler_start_method and TORCH_PROFILER_START_METHOD are do the exactly same thing.

guotuofeng:
> > I suppose #344 should be enough. No?
> 
> No, it doesn't include a few changes which are necessary for our use case. Please see me replies above for more detail.

could you please elaborate what is your use cases? From what I find in current repo, this changes are needless.

Reubend:
> could you please elaborate what is your use cases? From what I find in current repo, this changes are needless. If would be better if there are some code for you use cases.

Sure, let me describe our use case a bit more. We have an internal file system which can't be open sourced, and we'd like to integrate it with this plugin. We've already integrated it with TB's normal `gfile` system, but since this plugin uses a different filesystem abstraction, we have to port it to work here as well. In order to add it, we basically have two options:
- Fork the code and add our filesystem directly in `file.py` along with the local+Azure+GCP+S3 filesystems
- Call `register_filesystem` at runtime to add our filesystem

The first option works, but we'd need to resolve merge conflicts in the future, and keep our fork up to date. The second option, which was recommended to us by Google when we did the original integration, is cleaner because it allows us to keep our code entirely separate from the open source code. In normal usage, `register_filesystem` can be called at any point in time to add new filesystems, so we decided to use that approach. 

The issues is that when `forkserver` is used (which unfortunately is required for us in some cases due to race conditions in normal forked processes) the variable which stores filesystems gets wiped. So restoring the filesystems is necessary for the edge case where dynamically added filesystems are getting lost. It does make the code a bit more complicated, but it's still backwards compatible and it doesn't break anything. I think it's worth doing this to address this edge case. 

As for the flag vs static class, I think that's not as important. We can use a static flag, but the reason I added the dynamic setting was so that we can optimize better (i.e. start with `fork` for efficiency, and then switch to `forkserver` for compatibility when it's necessary.

guotuofeng:
> As for the flag vs static class, I think that's not as important. We can use a static flag, but the reason I added the dynamic setting was so that we can optimize better (i.e. start with `fork` for efficiency, and then switch to `forkserver` for compatibility when it's necessary. But again, this commit still supports the flags, so it's backwards compatible. It does make the code a bit more complicated, but I think it's worth it.

For the customized file system which is registered by yourself, I think https://github.com/pytorch/kineto/pull/344 should support it after I change the __getstate__ and __setstate__ in cache.py

As for the dynamic flag, I am not still clear the scenario. It seems that the global variable PROFILER_START_METHOD will be used, whose role is exactly same with the environment variable.  So, I am not convinced yet. It would be better if there are some code examples or more explanation about the background that needs it.

tb_plugin/torch_tb_profiler/io/file.py
guotuofeng: with my change in https://github.com/pytorch/kineto/pull/344, the get_all_filesystems/restore_filesystems is needless.

Reubend:
> As for the dynamic flag, I am not still clear the scenario. It seems that the global variable PROFILER_START_METHOD will be used, whose role is exactly same with the environment variable? I am afraid that I am not convinced yet. It would be better if there are some code examples or more explanation about the background or scenario that needs it.

Okay. I just removed the dynamic flag, and simplified the code. Can you please another look?

As for #344, we can't use the following lines:
```python
__all__ = [x for x in dir(mp.get_context(get_start_method())) if not x.startswith('_')]
globals().update((name, getattr(mp.get_context(get_start_method()), name)) for name in __all__)
```
because they override the globals, which can break other plugins that require the default start method. If a global override was usable, then we wouldn't need to use contexts at all, because we could just use `mp` directly.

tb_plugin/torch_tb_profiler/io/file.py
Reubend: Great, let me test that change and see if it works on my end.

guotuofeng:
> > As for the dynamic flag, I am not still clear the scenario. It seems that the global variable PROFILER_START_METHOD will be used, whose role is exactly same with the environment variable? I am afraid that I am not convinced yet. It would be better if there are some code examples or more explanation about the background or scenario that needs it.
> 
> Okay. I just removed the dynamic flag, and simplified the code. Can you please another look?
> 
> As for #344, we can't use the following lines:
> 
> ```python
> __all__ = [x for x in dir(mp.get_context(get_start_method())) if not x.startswith('_')]
> globals().update((name, getattr(mp.get_context(get_start_method()), name)) for name in __all__)
> ```
> 
> because they override the globals, which can break other plugins that require the default start method. If a global override was usable, then we wouldn't need to use contexts at all, because we could just use `mp` directly.

From the document of globals(), it seems the globals only return the dictionary of the variable defined in CURRENT module. So, I think it should not be an issue with your case.

image

guotuofeng:
> > As for the dynamic flag, I am not still clear the scenario. It seems that the global variable PROFILER_START_METHOD will be used, whose role is exactly same with the environment variable? I am afraid that I am not convinced yet. It would be better if there are some code examples or more explanation about the background or scenario that needs it.
> 
> Okay. I just removed the dynamic flag, and simplified the code. Can you please another look?
> 
> As for #344, we can't use the following lines:
> 
> ```python
> __all__ = [x for x in dir(mp.get_context(get_start_method())) if not x.startswith('_')]
> globals().update((name, getattr(mp.get_context(get_start_method()), name)) for name in __all__)
> ```
> 
> because they override the globals, which can break other plugins that require the default start method. If a global override was usable, then we wouldn't need to use contexts at all, because we could just use `mp` directly.

just test the following simple scripts.:
[test.zip](https://github.com/pytorch/kineto/files/6805434/test.zip)

tb_plugin/torch_tb_profiler/io/file.py
guotuofeng: @Reubend, please let me know whether my change works or not for your cases.

Reubend:
Ah, I misunderstood your code then. My apologies. 

#350 looks good, and I've approved that one. I'm closing this PR then, so that your can be merged instead. :)

Issue #338, opened 2021-07-07
TensorBoard plugin blocks TensorBoard UI from fully loading


psybuzz:
Repro:
- In a fresh Python virtualenv, `pip install tensorboard torch-tb-profiler`
- Open TensorBoard (http://localhost:6006 by default)
- The UI under the header is blank, but the loading spinner rotates
![image](https://user-images.githubusercontent.com/2322480/124791363-d87d0c80-df00-11eb-8931-4d7619e208f7.png)
- In the DevTools network panel, the '/plugins_listing' request is stalled
- Eventually showing an error
![image](https://user-images.githubusercontent.com/2322480/124791476-f0ed2700-df00-11eb-9f0a-462dbc15aa0b.png)

On my machine, I tested this with
- tensorboard==2.5.0
- torch-tb-profiler==0.2.0

From a brief read, torch-tb-plugin's `is_active()` method seemingly issues blocking call to wait for data:
https://github.com/pytorch/kineto/blob/be239c044d2f4b62982e1ff64144097acd69f59a/tb_plugin/torch_tb_profiler/plugin.py#L69-L74

On the TensorBoard side, there have been a few reports from users who find TensorBoard responsive again after uninstalling torch-tb-plugin: https://github.com/tensorflow/tensorboard/issues/5076, https://github.com/tensorflow/tensorboard/issues/5111

TensorBoard's other builtin plugins are implemented with `is_active`s that either return quickly or are non-blocking.  Would it be possible for torch-tb-plugin to do something similar?

EricWiener:
I'm also running into this issue with a docker image w/ tensorflow 2.5.0, tensorboard 2.5.0, and torch_tb_profiler 0.2.0.

guotuofeng:
This is a known bug, we will fix it

PR #340, opened 2021-07-08
Fix a couple of bugs

guotuofeng:
Fix bug https://github.com/pytorch/kineto/issues/130
https://github.com/pytorch/kineto/issues/270
https://github.com/pytorch/kineto/issues/338


guotuofeng:
image

tb_plugin/fe/src/app.tsx
```diff
@@ -153,8 +162,10 @@ export const App = () => {
       try {
         const runs = await api.defaultApi.runsGet()
         setRuns(runs)
+        setHasRuns(true)
```
TomWildenhain-Microsoft, line 165: Won't this set `hasRuns` to `true` even if the runs returned from the backend is empty?

TomWildenhain-Microsoft: I don't think you need a new state variable.  Can you just check `if (runs)` in any place you would do `if (hasRuns)`?

guotuofeng: I need the hasRun to differentiate whether the load is done but no any run or the loading is in progress

tb_plugin/fe/src/app.tsx
```diff
@@ -248,6 +259,22 @@ export const App = () => {
   }
 
   const renderContent = () => {
+    if (!hasRuns) {
+      return (<Card className={classes.container}>
```
TomWildenhain-Microsoft, line 263: This will be more consistent with the rest of the UI if you use CardHeader.

tb_plugin/fe/src/app.tsx
```diff
@@ -122,12 +125,18 @@ const useStyles = makeStyles((theme) => ({
   },
   iconButton: {
     padding: '8px'
-  }
+  },
+  container: {
+    display: 'flex',
+    padding: '80px',
```
TomWildenhain-Microsoft, line 131: Any reason for the enormous padding?  If you want the text closer to the center of the screen, I'd prefer a margin so the card to text distance is consistent with other cards.

nbcsm (approved):

tb_plugin/fe/src/app.tsx
TomWildenhain-Microsoft: Ah, I see.  The backend returns a 404, not an empty list when it finishes scanning the directory if there are no runs.

TomWildenhain-Microsoft (changes requested):

PR #344, opened 2021-07-08
use multiprocessing context specific to the start_method

guotuofeng:
use multiprocessing context specific to the start_method to starting the multiprocessing


guotuofeng:
close the PR since the commit history is dirty and created PR 350

PR #351, opened 2021-07-13
add kernel columns

gaoteng-git:
Add the below 4 columns into kernel view's table.
image


tb_plugin/torch_tb_profiler/profiler/module_parser.py
```diff
@@ -181,12 +181,18 @@ def parse_kernels(kernel_list):
             for kernel in kernel_list:
                 dur = kernel.end_time - kernel.start_time
                 op_name = "N/A" if kernel.op_node is None else kernel.op_node.name
-                key = kernel.name + "###" + op_name
+                key = kernel.name + "###" + op_name + "###" + \
```
guotuofeng, line 184: use string.join instead of add operation?

tb_plugin/torch_tb_profiler/profiler/node.py
```diff
@@ -166,6 +171,10 @@ def get_node_argument(event):
         if event.type == EventTypes.KERNEL:
             kwargs["blocks_per_sm"] = event.args.get("blocks per SM", 0)
             kwargs["occupancy"] = event.args.get("est. achieved occupancy %", 0)
+            kwargs["grid"] = str(event.args.get("grid"))
```
guotuofeng, line 174: In case of no these fields, the value will be "None" string. Should we not include these values?

tb_plugin/test/test_profiler.py
```diff
@@ -1644,5 +1644,136 @@ def test_memory_view(self):
...
+            "name": "op2", "pid": 13721, "tid": "456",
+            "ts": 340, "dur": 70,
+            "args": {"Input Dims": [[], [32, 1000], [32], [], [], [], []], "External id": 4}
+          },          
+          {
+            "ph": "X", "cat": "Kernel", 
+            "name": "kernel1", "pid": 0, "tid": "stream 7",
+            "ts": 230, "dur": 15,
+            "args": {"correlation": 1000, "external id": 3, "device": 0,
+                    "grid": [16, 1, 1], "block": [16, 16, 16], "registers per thread": 18, "shared memory": 0}
```
guotuofeng, line 1668: should we add test cases for missing these values?

guotuofeng (changes requested):

tb_plugin/torch_tb_profiler/profiler/module_parser.py
gaoteng-git: Thanks. Done.

tb_plugin/torch_tb_profiler/profiler/node.py
gaoteng-git: Got it. But I prefer still keep the kernel there, and with "None" as this field's value.
I've explicitly set the default value as "None" in latest commit.

tb_plugin/test/test_profiler.py
gaoteng-git: Manually removed a kernel's "block" field, and test pass on UI side:
image

guotuofeng: It would be better to make the edge case automatically

tb_plugin/torch_tb_profiler/profiler/node.py
```diff
@@ -166,6 +171,10 @@ def get_node_argument(event):
         if event.type == EventTypes.KERNEL:
             kwargs["blocks_per_sm"] = event.args.get("blocks per SM", 0)
             kwargs["occupancy"] = event.args.get("est. achieved occupancy %", 0)
+            kwargs["grid"] = str(event.args.get("grid", "None"))
```
guotuofeng, line 174: keep the original value by removing the str

tb_plugin/test/test_profiler.py
gaoteng-git: Done. Added a unit test handling kernel without some field.

tb_plugin/torch_tb_profiler/profiler/node.py
gaoteng-git: Thanks, done.

guotuofeng (approved):

PR #352, opened 2021-07-14
add export graph in examples

guotuofeng:


tb_plugin/examples/resnet50_profiler_api.py
```diff
@@ -52,3 +70,4 @@
         if step + 1 >= 4:
             break
         p.step()
+    export_graph('./result', model, model_inputs)
```
guyang3532, line 73: this './result' keep same as line52 using 'log_dir' ?

PR #353, opened 2021-07-14
Table sort column specified by backend

gaoteng-git:
1. Update to "Kernel Properties + Op Name".
2. Update default sort-by column to be specified by backend, in Operator view table and Kernel view table.
    (The previous table uses a fixed column, such as 3rd one, as default sort-by)

image


gaoteng-git:
GPU run:
image
image
image
image

Pure CPU run:
image
image

PR #359, opened 2021-07-21
refactor plugin to add event hierarchy for different event types

guotuofeng:
Add event classes hierarchy for different event types like: kernel, operator, duration, etc.

### DIFF
--- a/tb_plugin/torch_tb_profiler/profiler/trace.py
+++ b/tb_plugin/torch_tb_profiler/profiler/trace.py
@@ -45,36 +45,41 @@ class BaseEvent(object):
         self.tid = data.get("tid")
         self.args = data.get("args", {})
 
-class TraceEvent(BaseEvent):
+class DurationEvent(BaseEvent):
     def __init__(self, type, data):
         super().__init__(type, data)
+        self.category = data.get("cat", "")
         self.duration = data.get("dur")
 
-    @property
-    def external_id(self):
         extern_id = self.args.get("external id")
         if extern_id is None:
             extern_id = self.args.get("External id")
+        self.external_id = extern_id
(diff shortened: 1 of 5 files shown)


tb_plugin/torch_tb_profiler/profiler/trace.py
```diff
@@ -45,36 +45,41 @@ def __init__(self, type, data):
...
+        self.input_shape = shape
 
-        return shape
-
-    @property
-    def input_type(self):
-        return self.args.get("Input type")
-
-class ProfilerStepEvent(TraceEvent):
+class ProfilerStepEvent(OperatorEvent):
```
gaoteng-git, line 82: I'm not sure why its inherits changed from DurationEvent to OperatorEvent.

gaoteng-git (commented):
As events could be distinguished by class type, such as "OperatorEvent", "KernelEvent", is it possible to remove BaseEvent.type?

gaoteng-git (approved):
LGTM.

tb_plugin/torch_tb_profiler/profiler/trace.py
```diff
@@ -124,8 +122,14 @@ def create_trace_event(event):
         name = event.get("name")
         if name and name.startswith("ProfilerStep#"):
             return ProfilerStepEvent(event)
-
-    if event_type is not None:
-        return TraceEvent(event_type, event)
+        else:
+            return OperatorEvent(event_type, event)
+    elif event_type == EventTypes.PYTHON:
```
guyang3532, line 127: If change EventTypeMap['Python'] = EventTypes.OPERATOR, then this elif can be removed.

PR #361, opened 2021-07-22
Memory curve

cloudhan:


tb_plugin/torch_tb_profiler/profiler/data.py
```diff
@@ -49,6 +52,7 @@ def __init__(self, worker, span=None):
         self.approximated_sm_efficiency_ranges = None  # Cached here. Will be processed to json on first trace view.
         self.blocks_per_sm_count = None
         self.occupancy_count = None
+        self.tid2tree = None
```
guotuofeng, line 55: how many time when we pickle/unpickle the whole tree?

tb_plugin/torch_tb_profiler/profiler/data.py
```diff
@@ -90,7 +84,9 @@ def parse(run_dir, worker, span, path, caches):
         for data in trace_json:
             event = trace.create_event(data)
             if event is not None:
+                profile.profiler_start_ts = min(profile.profiler_start_ts, event.ts)
```
guotuofeng, line 87: we can get the start time by find the root node in tid2tree, which should be much more faster

tb_plugin/torch_tb_profiler/profiler/data.py
```diff
@@ -59,27 +63,17 @@ def __init__(self, worker, span=None):
...
-        return False
-
     @staticmethod
-    def parse(run_dir, worker, span, path, caches):
+    def from_file(run_dir, worker, span, path, caches):
         logger.debug("Parse trace, run_dir=%s, worker=%s", run_dir, path)
-
         trace_path, trace_json = RunProfileData._preprocess_file(caches, io.join(run_dir, path))
+        profile = RunProfileData.from_json(worker, span, trace_json)
+        profile.trace_file_path = path
```
guotuofeng, line 71: bug here

cloudhan: What bug?

tb_plugin/torch_tb_profiler/profiler/data.py
cloudhan: It does not necessary be faster. Since this is scanned only once in subprocess. You inevitablely scan through all events, Record it there will be much resonable.

tb_plugin/torch_tb_profiler/profiler/data.py
cloudhan: 1 for pickle and 1 for unpickle

tb_plugin/torch_tb_profiler/profiler/data.py
cloudhan: Fixed

tb_plugin/torch_tb_profiler/profiler/trace.py
```diff
@@ -9,7 +9,9 @@
 
 logger = utils.get_logger()
 
-DeviceType = IntEnum('DeviceType', ['CPU', 'CUDA'], start=0)
+class DeviceType(IntEnum):
+    CPU = 0
+    CUDA = 1
```
guyang3532, line 14: Why?

tb_plugin/torch_tb_profiler/profiler/trace.py
```diff
@@ -103,6 +103,29 @@ def __init__(self, type, data):
...
+    def bytes(self):
+        return self.args.get("Bytes", 0)
+
+    @property
+    def total_allocated(self):
+        # TODO: Allocated Bytes to be renamed to Total Allocated
+        old = self.args.get("Allocated Bytes")
+        if old is not None:
+            return old
+        return self.args.get("Total Allocated", float("nan"))
```
guotuofeng, line 120: why not None?

cloudhan: Because it is a float. Just like how pandas treating missing value mentioned in https://pandas.pydata.org/pandas-docs/stable/user_guide/missing_data.html
> The choice of using NaN internally to denote missing data was largely for simplicity and performance reasons. Starting from pandas 1.0, some optional data types start experimenting with a native NA scalar using a mask-based approach. See here for more.
No preference here, they will be filtered out latter before returning to frontend.

guotuofeng (approved):

tb_plugin/torch_tb_profiler/profiler/memory_parser.py
```diff
@@ -260,55 +210,19 @@ def update_node(self):
 
                 # the current_node is the one contains the record at this moment.
                 if is_operator_node(current_node):
-                    if not BENCHMARK_MEMORY or record not in current_node.memory_records:
+                    if record not in current_node.memory_records:
```
guotuofeng, line 213: seems needless?

tb_plugin/torch_tb_profiler/profiler/data.py
```diff
@@ -276,6 +267,11 @@ def _analyze_distributed_metrics(self):
                        round(communication_ratio * 100, 1), "https://nvidia.github.io/apex/optimizers.html#apex.optimizers.FusedLAMB")
             self.recommendations.append(text)
 
+    def _memory_events(self) -> List[MemoryEvent]:
+        memory_events = [e for e in self.events if e.type == EventTypes.MEMORY]
```
guotuofeng, line 271: should we only save the processed_records here?

tb_plugin/torch_tb_profiler/profiler/memory_parser.py
```diff
@@ -66,30 +56,12 @@ def __init__(self, tid2tree, op_list):
...
+        self.update_node()
 
-        if BENCHMARK_MEMORY:
-            self.update_node_recursive()
-            self.update_node()
-        else:
-            self.update_node()
-
-    @benchmark
     def get_memory_statistics(self):
```
guotuofeng, line 65: need we calculate the statistic with specific start/end timestamp here?

guotuofeng (changes requested):
should we fix the statistic issue in the time range?

tb_plugin/fe/src/components/MemoryView.tsx
```diff
@@ -104,34 +177,81 @@ export const MemoryView: React.FC<IProps> = (props) => {
...
-                    <Select
-                      labelId="memory-device"
-                      value={device}
-                      onChange={onDeviceChanged}
-                    >
-                      {devices.map((device) => (
-                        <MenuItem value={device}>{device}</MenuItem>
-                      ))}
-                    </Select>
+            <Grid item>
```
guotuofeng, line 180: need we hide the Grid when memoryCurveGraph is nil?

tb_plugin/fe/src/components/MemoryView.tsx
```diff
@@ -104,34 +177,81 @@ export const MemoryView: React.FC<IProps> = (props) => {
...
+                <TextField
+                  classes={{ root: classes.inputWidthOverflow }}
+                  value={searchEventOperatorName}
+                  onChange={onSearchEventOperatorChanged}
+                  type="search"
+                  label="Search by Name"
+                />
+              </Grid>
+            </Grid>
+            <Grid item direction="column">
```
guotuofeng, line 227: hide it when memoryEventsData is nil?

tb_plugin/fe/src/components/MemoryView.tsx
cloudhan: It can be done if curve rows is empty on first load. Let's me check how to hide it first....

guotuofeng (approved):

PR #371, opened 2021-08-02
refactor kernel/operator aggregater so that diff run can leverage

guotuofeng:
split build_tree, aggregate and get kernels/operators into its own class so that diff run can reuse the same logic.

### DIFF
--- a/tb_plugin/torch_tb_profiler/profiler/module_parser.py
+++ b/tb_plugin/torch_tb_profiler/profiler/module_parser.py
@@ -3,6 +3,7 @@
 # --------------------------------------------------------------------------
 import sys
 from collections import defaultdict
+from typing import Dict, List
 
 from .. import utils
 from .node import OperatorNode, RuntimeNode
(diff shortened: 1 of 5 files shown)


tb_plugin/torch_tb_profiler/profiler/module_parser.py
```diff
@@ -68,28 +92,70 @@ def avg_occupancy(self):
...
+            op_list.sort(key=lambda x: (x.start_time, -x.end_time))
+            main_tid = any([op.name.startswith("ProfilerStep#") for op in op_list])
+            if main_tid:
+                # only append the staled device nodes into main thread
+                root_node = self._build_tree(op_list, zero_rt_list, tid, staled_device_nodes)
+            else:
+                root_node = self._build_tree(op_list, zero_rt_list, tid, [])
+            self.tid2tree[int(tid)] = root_node
+
+    def _build_tree(self, host_node_list, zero_rt_list, tid, device_nodes):
```
gaoteng-git: rename "device_nodes" to "staled_device_nodes"?

tb_plugin/torch_tb_profiler/profiler/module_parser.py
```diff
@@ -68,28 +92,70 @@ def avg_occupancy(self):
...
+        '''host_node_list: list of OperatorNode and ProfilerStepNode.
+        zero_rt_list: list of RuntimeNode with external_id=0.'''
 
-        def build_tree_relationship(host_node_list, zero_rt_list):
+        def build_tree_relationship(host_node_list, zero_rt_list, device_nodes):
```
gaoteng-git: rename "device_nodes" to "staled_device_nodes"?

tb_plugin/torch_tb_profiler/profiler/module_parser.py
```diff
@@ -68,28 +92,70 @@ def avg_occupancy(self):
...
+        for _, device_nodes in corrid_to_device.items():
+             staled_device_nodes.extend([n for n in device_nodes if n.type == EventTypes.KERNEL])
+
+        for tid, op_list in tid2list.items():
+            zero_rt_list = tid2zero_rt_list[tid] if tid in tid2zero_rt_list else []
+            # Note that when 2 start_time are equal, the one with bigger end_time should be ahead of the other.
+            op_list.sort(key=lambda x: (x.start_time, -x.end_time))
+            main_tid = any([op.name.startswith("ProfilerStep#") for op in op_list])
+            if main_tid:
+                # only append the staled device nodes into main thread
```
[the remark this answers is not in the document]
guotuofeng: from the time being, we only use the kernels with root node for kernel view and there is no need to create a separated dummy thread by now

gaoteng-git (approved):
It's better to add a comment to highlight future usage of "dummy" RuntimeNode. For example, "Note: Although kernels of this dummy runtime is put under main thread's tree, we don't know which thread launches them. Don't make belonging thread assumption on future usage."

guotuofeng:
> Note: Although kernels of this dummy runtime is put under main thread's tree, we don't know which thread launches them. Don't make belonging thread assumption on future usage."

comments added.

PR #372, opened 2021-08-02
Forward backward correlate

gaoteng-git:
Add new link event "forward_backward" to dumped json file. 
The latency overhead is only about 1% to `profile.__exit__`

image
image

### DIFF
--- a/libkineto/include/GenericTraceActivity.h
+++ b/libkineto/include/GenericTraceActivity.h
@@ -16,6 +16,15 @@
 #include "TraceActivity.h"
 #include "TraceSpan.h"
 
+#define GENERATE_FORWARD_BACKWARD_LINK(link_id, link_type) \
+  ((((uint64_t)(link_type)) << 60) | (link_id))
+#define GET_LINK_TYPE(link) \
+  (((uint64_t)(link)) >> 60)
+#define GET_LINK_ID(link) \
+  ((((uint64_t)1 << 60) - 1) & (link))
+#define LINK_FORWARD_BACKWARD_HEAD 1
+#define LINK_FORWARD_BACKWARD_TAIL 2
+
 namespace libkineto {
 
 // @lint-ignore-every CLANGTIDY cppcoreguidelines-non-private-member-variables-in-classes
@@ -27,7 +36,7 @@ class GenericTraceActivity : public TraceActivity {
 
   GenericTraceActivity(
       const TraceSpan& trace, ActivityType type, const std::string& name)
-      : activityType(type), activityName(name), traceSpan_(&trace) {
+      : activityType(type), activityName(name), traceSpan_(&trace), linkedAct(nullptr), linkId(0){
   }
 
   int64_t deviceId() const override {
@@ -59,7 +68,7 @@ class GenericTraceActivity : public TraceActivity {
   }
 
   const TraceActivity* linkedActivity() const override {
-    return nullptr;
+    return linkedAct;
   }
 
   const TraceSpan* traceSpan() const override {
@@ -86,6 +95,9 @@ class GenericTraceActivity : public TraceActivity {
   int32_t resource{0};
   ActivityType activityType;
   std::string activityName;
+  TraceActivity* linkedAct;
+  // High-4bits represents link type.
+  uint64_t linkId;
 
  private:
   const TraceSpan* traceSpan_;
--- a/libkineto/include/GenericTraceActivity.h
+++ b/libkineto/include/GenericTraceActivity.h
@@ -16,16 +16,10 @@
 #include "TraceActivity.h"
 #include "TraceSpan.h"
 
-#define GENERATE_FORWARD_BACKWARD_LINK(link_id, link_type) \
-  ((((uint64_t)(link_type)) << 60) | (link_id))
-#define GET_LINK_TYPE(link) \
-  (((uint64_t)(link)) >> 60)
-#define GET_LINK_ID(link) \
-  ((((uint64_t)1 << 60) - 1) & (link))
-#define LINK_FORWARD_BACKWARD_HEAD 1
-#define LINK_FORWARD_BACKWARD_TAIL 2
-
 namespace libkineto {
+
+// Link type, used in GenericTraceActivity.flow.type
+constexpr unsigned int kLinkFwdBwd = 1;
 
 // @lint-ignore-every CLANGTIDY cppcoreguidelines-non-private-member-variables-in-classes
 // @lint-ignore-every CLANGTIDY cppcoreguidelines-pro-type-member-init
@@ -36,7 +30,7 @@ class GenericTraceActivity : public TraceActivity {
 
   GenericTraceActivity(
       const TraceSpan& trace, ActivityType type, const std::string& name)
-      : activityType(type), activityName(name), traceSpan_(&trace), linkedAct(nullptr), linkId(0){
+      : activityType(type), activityName(name), traceSpan_(&trace){
   }
 
   int64_t deviceId() const override {
@@ -68,7 +62,7 @@ class GenericTraceActivity : public TraceActivity {
   }
 
   const TraceActivity* linkedActivity() const override {
-    return linkedAct;
+    return flow.linkedActivity;
   }
 
   const TraceSpan* traceSpan() const override {
@@ -95,9 +89,12 @@ class GenericTraceActivity : public TraceActivity {
   int32_t resource{0};
   ActivityType activityType;
   std::string activityName;
-  TraceActivity* linkedAct;
-  // High-4bits represents link type.
-  uint64_t linkId;
+  struct Flow {
+    Flow(): linkedActivity(nullptr), id(0), type(0) {}
+    TraceActivity* linkedActivity; // Only set in destination side.
+    uint64_t id : 60;
+    uint64_t type : 4;
+  } flow;
 
  private:
   const TraceSpan* traceSpan_;
--- a/libkineto/src/output_json.cpp
+++ b/libkineto/src/output_json.cpp
@@ -252,12 +252,12 @@ void ChromeTraceLogger::handleGenericActivity(
 
 void ChromeTraceLogger::handleGenericLink(const GenericTraceActivity& op) {
   if (op.flow.type == kLinkFwdBwd) {
-    this->handleFwdBwdLinkStart(*((GenericTraceActivity*)op.flow.linkedActivity), "forward_backward", "fwd_bwd");
-    this->handleFwdBwdLinkEnd(op, "forward_backward", "fwd_bwd");
+    this->handleGenericLinkStart(*((GenericTraceActivity*)op.flow.linkedActivity), "forward_backward", "fwd_bwd");
+    this->handleGenericLinkEnd(op, "forward_backward", "fwd_bwd");
   }
 }
 
-void ChromeTraceLogger::handleFwdBwdLinkStart(
+void ChromeTraceLogger::handleGenericLinkStart(
     const libkineto::GenericTraceActivity& s,
     const std::string& cat,
     const std::string& name) {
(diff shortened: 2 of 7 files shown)


chaekit:
@gdankel +1 on adding extra fields in GTA after all the refactoring we have gone through. I'll add a comment in the class to warn future contributor about this. other than that @gaoteng-git thanks for the PR!

libkineto/include/GenericTraceActivity.h
```diff
@@ -87,6 +87,11 @@ class GenericTraceActivity : public TraceActivity {
   ActivityType activityType;
   std::string activityName;
 
+  int64_t seqNumber{-1};
```
gdankel, line 90: Adding four fields for this is a bit much. And it's very specific, both to PyTorch and to the way forward / backwards currently works.
At the very least this needs a big FIXME comment to explain this is a temporary and hacky implementation.

libkineto/src/ActivityProfiler.cpp
```diff
@@ -186,6 +190,37 @@ void ActivityProfiler::processCpuTrace(
...
+      // Assumption: Among all ops with same sequence number,
+      // the one with biggest start time is most likely launching backward op.
+      auto iter = tidSeq2activity_.find(key);
+      if (iter == tidSeq2activity_.end()) {
+        uint64_t uid = tidSeq2activity_.size() + 1; // Start from 1.
+        act.unique_op_id = uid;
+        tidSeq2activity_[key] = &act;
+      }
+      else {
+        if (act.startTime >= iter->second->startTime) {
```
gdankel, line 215: When does it happen that the same key is encountered multiple times?

gdankel (commented):
I think this is good for a prototype implementation but we can't land this without big comments in the code referring to an issue describing a plan for improving.

libkineto/src/ActivityProfiler.cpp
gaoteng-git: Now the sequence number is only incremented on creating a "Node" object for backward pass, by [calling "at::sequence_number::get_and_increment()"](https://github.com/pytorch/pytorch/blob/bf7d03ff1f3cded53a16b707a2db76ce25c130f7/torch/csrc/autograd/function.h#L130). 
Among all ops with same sequence number, the one with biggest startTime is the one launching backward op. So I choose this one to cache into "tidSeq2activity_" for matching with backward-op of same tid and seq number.
I've added comments explaining this.

libkineto/include/GenericTraceActivity.h
gaoteng-git: Yes, it's a little heavy. I've considered moving these columns into "args" json string, but it will need parse the string in kineto side when using them to do correlation.
I've added FIXME to it. Please feel free to update it in a better way later on.

gdankel: I think most of the code does not belong in Kineto itself but in PyTorch Kineto profiler.

libkineto/src/ActivityProfiler.cpp
```diff
@@ -187,6 +191,40 @@ void ActivityProfiler::processCpuTrace(
 
   CpuGpuSpanPair& span_pair = recordTraceSpan(cpuTrace.span, cpuTrace.gpuOpCount);
   TraceSpan& cpu_span = span_pair.first;
+  // Update act.unique_op_id to correlate forward-op and backward-op.
```
gdankel, line 194: I suggest adding all this code into profiler_kineto.cpp (as part of #62553) and call it from finalizeCPUTrace.
Then, the only thing we need to add to GenericTraceActivity is a linkedFlowId or a pointer to the linked activity that can be returned by linkedActivity().
My overall plan for flows is to encode a flow ID and a flow type in a flow field (e.g. 4 bits for flow type). Profilers can register flow types and associate them with names to make it possible to link activities across separate profilers.
In this case you're only linking activities within a single profiler so it's a bit simpler.

libkineto/src/output_json.cpp
```diff
@@ -244,6 +244,43 @@ void ChromeTraceLogger::handleGenericActivity(
       op.traceSpan()->name, op.traceSpan()->iteration, separator,
       op_metadata);
   // clang-format on
+
+  if (op.unique_op_id != 0) {
```
gdankel, line 248: Change to check for linkedActivity() != nullptr after making changes suggested above, and add handleGenericLink() instead taking two activities (start and end).

gdankel (changes requested):

libkineto/include/GenericTraceActivity.h
gaoteng-git: OK. Moved to PyTorch side.

libkineto/src/ActivityProfiler.cpp
gaoteng-git: Got it. Made following change to "GenericTraceActivity.h":
Added "TraceActivity* linkedAct;" to point to its corresponding backward or forward op. Added "uint64_t linkId;" to give a pair of forward&backward ops same value, the high-4bits is link type.
And move correlating logic to "finalizeCPUTrace".

libkineto/src/output_json.cpp
gaoteng-git: Thanks. Done.

libkineto/src/output_json.h
```diff
@@ -44,6 +44,10 @@ class ChromeTraceLogger : public libkineto::ActivityLogger {
 
   void handleGenericActivity(const GenericTraceActivity& activity) override;
 
+  void handleGenericLink(const GenericTraceActivity& activity);
```
gdankel, line 47: Move these to private

gaoteng-git: Thanks. Done.

libkineto/include/GenericTraceActivity.h
```diff
@@ -86,6 +95,9 @@ class GenericTraceActivity : public TraceActivity {
   int32_t resource{0};
   ActivityType activityType;
   std::string activityName;
+  TraceActivity* linkedAct;
+  // High-4bits represents link type.
+  uint64_t linkId;
```
gdankel, line 100: From offline discussion, how about this for now:
struct {
TraceActivity* prevActivity;
uint64_t id : 28;
uint64_t type : 4;
} flow;

prevActivity establishes order, so even if timestamps are equal we know which come first.

libkineto/include/GenericTraceActivity.h
```diff
@@ -16,6 +16,15 @@
 #include "TraceActivity.h"
 #include "TraceSpan.h"
 
+#define GENERATE_FORWARD_BACKWARD_LINK(link_id, link_type) \
```
gdankel, line 19: I think we can remove all these defines with the below bitfield proposal.

libkineto/include/GenericTraceActivity.h
```diff
@@ -16,6 +16,15 @@
...
+  ((((uint64_t)(link_type)) << 60) | (link_id))
+#define GET_LINK_TYPE(link) \
+  (((uint64_t)(link)) >> 60)
+#define GET_LINK_ID(link) \
+  ((((uint64_t)1 << 60) - 1) & (link))
+#define LINK_FORWARD_BACKWARD_HEAD 1
```
gdankel, line 25: We can use one type here I think, just to determine category and name:
constexpr unsigned int kLinkFwdBwd = 1;

libkineto/src/output_json.cpp
```diff
@@ -244,6 +244,50 @@ void ChromeTraceLogger::handleGenericActivity(
...
   // clang-format on
+
+  if (op.linkedActivity() != nullptr) {
+    handleGenericLink(op);
+  }
+}
+
+void ChromeTraceLogger::handleGenericLink(const GenericTraceActivity& op) {
+  uint64_t link_type = GET_LINK_TYPE(op.linkId);
+  if (link_type == LINK_FORWARD_BACKWARD_HEAD) {
```
gdankel, line 255: Assuming linkedActivity is set only for destination and above bitfield suggestion:

if (op.flow.linkType == kLinkFwdBwd) {
  handleGenericLinkStart(op.flow.linkedActivity(), "forward_backward", "fwd_bwd");
  handleGenericLinkEnd(op, "forward_backward", "fwd_bwd");
}

libkineto/src/output_json.cpp
```diff
@@ -244,6 +244,50 @@ void ChromeTraceLogger::handleGenericActivity(
...
+    return;
+  }
+
+  // clang-format off
+  traceOf_ << fmt::format(R"JSON(
+  {{
+    "ph": "s", "id": {}, "pid": {}, "tid": "{}", "ts": {},
+    "cat": "forward_backward", "name": "fwd_bwd"
+  }},)JSON",
+      GET_LINK_ID(s.linkId), processId(), s.resourceId(), s.timestamp());
```
gdankel, line 274: s.flow.id

gdankel (changes requested):

libkineto/include/GenericTraceActivity.h
```diff
@@ -27,7 +30,7 @@ class GenericTraceActivity : public TraceActivity {
 
   GenericTraceActivity(
       const TraceSpan& trace, ActivityType type, const std::string& name)
-      : activityType(type), activityName(name), traceSpan_(&trace) {
+      : activityType(type), activityName(name), traceSpan_(&trace), flow.linkedActivity(nullptr), linkId(0){
```
gdankel, line 33: linkId?

libkineto/src/output_json.cpp
```diff
@@ -244,6 +244,53 @@ void ChromeTraceLogger::handleGenericActivity(
...
+}
+
+void ChromeTraceLogger::handleGenericLink(const GenericTraceActivity& op) {
+  if (op.flow.type == kLinkFwdBwd) {
+    this->handleFwdBwdLinkStart(*(op.flow.linkedActivity), "forward_backward", "fwd_bwd");
+    this->handleFwdBwdLinkEnd(op, "forward_backward", "fwd_bwd");
+  }
+}
+
+void ChromeTraceLogger::handleFwdBwdLinkStart(
```
gdankel, line 260: rename to handleGenericLinkStart()?

libkineto/src/output_json.cpp
```diff
@@ -244,6 +244,53 @@ void ChromeTraceLogger::handleGenericActivity(
...
+  traceOf_ << fmt::format(R"JSON(
+  {{
+    "ph": "s", "id": {}, "pid": {}, "tid": "{}", "ts": {},
+    "cat": "{}", "name": "{}"
+  }},)JSON",
+      s.flow.id, processId(), s.resourceId(), s.timestamp(), cat, name);
+  // clang-format on
+}
+
+void ChromeTraceLogger::handleFwdBwdLinkEnd(
```
gdankel, line 278: rename?

gdankel:
LGTM, just need to pass tests and address nits.

gdankel (approved):

libkineto/src/output_json.cpp
gaoteng-git: Thanks. Done.

libkineto/src/output_json.cpp
gaoteng-git: Thanks. Done.

libkineto/include/GenericTraceActivity.h
gaoteng-git: Thanks. Done.

tb_plugin/torch_tb_profiler/profiler/multiprocessing.py
```diff
@@ -3,7 +3,7 @@
 
 
 def get_start_method():
-    return os.getenv('TORCH_PROFILER_START_METHOD')
+    return os.getenv('TORCH_PROFILER_START_METHOD', "spawn")
```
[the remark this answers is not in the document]
cloudhan: @chaekit Maybe we could do a cherry-pick to the master branch and remove the commit from this branch? What do you think?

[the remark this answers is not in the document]
cloudhan: That will create a merge conflict when we want to merge plugin/vnext to master but we can resolve it on our side.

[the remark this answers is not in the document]
cloudhan: @gaoteng-git Please consider merge this PR after https://github.com/pytorch/kineto/pull/402

chaekit (approved):

PR #383, opened 2021-08-12
add recommendation for Tensor Cores

gaoteng-git:
image


tb_plugin/torch_tb_profiler/profiler/module_parser.py
```diff
@@ -156,3 +157,4 @@ def aggregate(self, tid2tree):
         self.op_list_groupby_name_input = list(agg_result[1].values())
         self.stack_lists_group_by_name = stack_lists_group_by_name
         self.stack_lists_group_by_name_input = stack_lists_group_by_name_input
+        self.tc_eligible_ops = sum(map(lambda op : op.tc_eligible, ops))
```
guotuofeng, line 160: seems list comprehension is faster than map if using lambda? https://stackoverflow.com/questions/1247486/list-comprehension-vs-map

gaoteng-git: @guotuofeng Cool! What about "sum([1 if op.tc_eligible else 0 for op in ops])"?

gaoteng-git: @guotuofeng Compared the speed, they are nearly same(<20% difference). I'll use "sum([1 if op.tc_eligible else 0 for op in ops])" for easy to understand.

tb_plugin/torch_tb_profiler/profiler/data.py
```diff
@@ -229,6 +231,21 @@ def analyze(self):
         self._analyze_distributed_metrics()
         self._analyze_gpu_metrics()
 
+        # Tensor Cores feature is available on GPU cards with compute capability >= 7.0
+        # https://docs.nvidia.com/cuda/cuda-c-programming-guide/index.html#features-and-technical-specifications
+        if self.device_props is not None and len(self.device_props) > 0:
```
guotuofeng, line 236: if self.device_props is enough.

gaoteng-git: Thanks! Done.

guotuofeng (approved):

PR #385, opened 2021-08-13
Fix cupti timestamp issue on windows

### LINKED ISSUE #384: cupti returns time since boot instead of time since unix epoch
blocker of #356.

This can be verified with [activity_trace.zip](https://github.com/pytorch/kineto/files/6980265/activity_trace.zip)

run `./build.ps1` on windows or `make` on linux

It is not mentioned anywhere in the cupti docs.

cloudhan:
Fixes #384 

On windows, cupti's timestamps are time since boot causing it return.
Results in kernel activity being filtered out and correction issue.

### DIFF
--- a/libkineto/src/CuptiActivityPlatform.cpp
+++ b/libkineto/src/CuptiActivityPlatform.cpp
@@ -1,0 +1,31 @@
+#include <chrono>
+
+namespace chrono = std::chrono;
+
+namespace KINETO_NAMESPACE {
+
+#ifdef _WIN32
+uint64_t epochs_diff() {
+  // On Windows, steady_clock wraps the QueryPerformanceCounter function.
+  // https://docs.microsoft.com/en-us/cpp/standard-library/steady-clock-struct?view=msvc-160
+  auto steady =
+      chrono::time_point_cast<chrono::nanoseconds>(chrono::steady_clock::now());
+  auto system =
+      chrono::time_point_cast<chrono::nanoseconds>(chrono::system_clock::now());
+
+  auto time_since_unix = system.time_since_epoch().count();
+  auto time_since_boot = steady.time_since_epoch().count();
+  return time_since_unix - time_since_boot;
+}
+
+uint64_t cuptiPlatformAgnosticTs(uint64_t ts) {
+  static uint64_t diff = epochs_diff();
+  return ts + diff;
+}
+#else
+uint64_t cuptiPlatformAgnosticTs(uint64_t ts) {
+  return ts;
+}
+#endif // _WIN32
+
+} // namespace KINETO_NAMESPACE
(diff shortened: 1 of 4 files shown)


libkineto/src/CuptiActivityPlatform.cpp
```diff
@@ -0,0 +1,31 @@
...
+      chrono::time_point_cast<chrono::nanoseconds>(chrono::steady_clock::now());
+  auto system =
+      chrono::time_point_cast<chrono::nanoseconds>(chrono::system_clock::now());
+
+  auto time_since_unix = system.time_since_epoch().count();
+  auto time_since_boot = steady.time_since_epoch().count();
+  return time_since_unix - time_since_boot;
+}
+
+uint64_t cuptiPlatformAgnosticTs(uint64_t ts) {
```
gdankel, line 21: How about unixEpochTimestamp() or unixTimestamp()?

cloudhan:
@gdankel I updated the function name as suggested, much better now. The pytorch repo integration test seems to be good for now. Do you think any test is left out, since I only enabled two obvious tests in `test_profiler.py`. Those two tests are noted as flaky, should I trigger it one more time to track the latest submodule commit?

libkineto/src/CuptiActivityPlatform.cpp
```diff
@@ -0,0 +1,31 @@
...
+      chrono::time_point_cast<chrono::nanoseconds>(chrono::steady_clock::now());
+  auto system =
+      chrono::time_point_cast<chrono::nanoseconds>(chrono::system_clock::now());
+
+  auto time_since_unix = system.time_since_epoch().count();
+  auto time_since_boot = steady.time_since_epoch().count();
+  return time_since_unix - time_since_boot;
+}
+
+uint64_t unixEpochTimestamp(uint64_t ts) {
```
gdankel, line 21: Does it make sense to implement this in libkineto/include/time_since_epoch.h instead?

cloudhan: I don't think so. `time_since_epoch.h` is a public header and this function purely serve as a implementation detail to make cupti timestamp on windows have the same semantics as declared in this header. It is not meant to be used by library users.

Also, there is function static variable to avoid computing the diff repeatedly, so there will be an implementation in a cpp file anyway.

cloudhan: but I am open minded to merge it with some other cupti related files.

gdankel: OK makes sense. We can leave it here for now - we have some other platform specific stuff elsewhere (ThreadUtil.cpp), so maybe at some point it makes sense to make a PlatformAbstraction or similar.

gdankel (approved):

PR #393, opened 2021-08-20
Workaround test cases deadlock

cloudhan:
Internally, plugin use multiprocessing to do heavy lifting of event trace parse and preprocessing.

However, multiprocessing default to `fork()` are causing deadlock on the moment of `fork` is called.

1. disable grpc `atfork` hook reduce definitely deadlock on `memory_curve` branch to probably deadlock.
2. remove barrier since it is unnecessary, to remove suspicious lock usage.
3. use `spawn` as workaround, deadlock on github actions disappear at the moment.


guotuofeng:
better add more description about what we found

guotuofeng (approved):

PR #405, opened 2021-09-02
Add Conda package

malfet:
Every conda package needs a valid LICENSE file, so copy-n-pasting one
from top-level folder

Test Plan: `conda build packaging/torch_tb_profiler`

### DIFF
--- a/tb_plugin/packaging/torch_tb_profiler/meta.yaml
+++ b/tb_plugin/packaging/torch_tb_profiler/meta.yaml
@@ -1,0 +1,37 @@
+package:
+  name: torch_tb_profiler
+  version: "{{ environ.get('BUILD_VERSION', '0.2.0') }}"
+
+source:
+ path: "{{ environ.get('SOURCE_ROOT_DIR', '../..') }}"
+
+requirements:
+  host:
+    - python
+    - setuptools
+    - pytorch
+
+  run:
+    - python
+    - pandas
+    - tensorboard
+
+build:
+  string: py{{py}}
+  script: python setup.py install --single-version-externally-managed --record=record.txt
+
+test:
+  imports:
+    - torch_tb_profiler
+  source_files:
+    - test
+  requires:
+    - tensorboard
+    - pandas
+
+
+about:
+  home: https://github.com/pytorch/kineto/tree/main/tb_plugin
+  license: BSD
+  license_file: LICENSE
+  summary: 'Tensoboard Plugin that provides visualization of PyTorch profiling'
--- a/tb_plugin/LICENSE
+++ b/tb_plugin/LICENSE
@@ -1,0 +1,33 @@
+BSD License
+
+For Kineto software
+
+Copyright (c) Facebook, Inc. and its affiliates. All rights reserved.
+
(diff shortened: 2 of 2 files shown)


tb_plugin/packaging/torch_tb_profiler/meta.yaml
```diff
@@ -0,0 +1,37 @@
...
+  script: python setup.py install --single-version-externally-managed --record=record.txt
+
+test:
+  imports:
+    - torch_tb_profiler
+  source_files:
+    - test
+  requires:
+    - tensorboard
+    - pandas
```
chauhang, line 30: Why do we need pandas?

chauhang: Please see the install dependencies for torch-tb-profiler [here](https://github.com/pytorch/kineto/blob/main/tb_plugin/ci_scripts/install_env.sh#L23)

chauhang (commented):
Thanks @malfet. Please see comments inline

tb_plugin/packaging/torch_tb_profiler/meta.yaml
malfet: This is not the source of truth for packaging system, but this is:
https://github.com/pytorch/kineto/blob/eaaf2acf8811e93f0175a904e63ee864cc884162/tb_plugin/setup.py#L27-L30

And the wheel that is currently published also depends on pandas:
```
% curl -O https://files.pythonhosted.org/packages/ac/53/49685c0b0d2388959fa58a1b6827bf7c565c91e01862aadde42c1ac10ad8/torch_tb_profiler-0.2.1-py3-none-any.whl; python3 -c 'import pkginfo;print(pkginfo.get_metadata("torch_tb_profiler-0.2.1-py3-none-any.whl").requires_dist)'
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100  932k  100  932k    0     0  5585k      0 --:--:-- --:--:-- --:--:-- 5585k
['pandas (>=1.0.0)', 'tensorboard (!=2.1.0,>=1.15)', "azure-storage-blob ; extra == 'blob'", "google-cloud-storage ; extra == 'gs'", "boto3 ; extra == 's3'"]
```

guyang3532:
Thanks @malfet , but it seems we have no permission to upload package to pytorch channel of conda?

chauhang (approved):

tb_plugin/packaging/torch_tb_profiler/meta.yaml
```diff
@@ -0,0 +1,37 @@
...
+    - setuptools
+    - pytorch
+
+  run:
+    - python
+    - pandas
+    - tensorboard
+
+build:
+  string: py{{py}}
```
guyang3532, line 20: Since this package only contains python scripts, is it better to specify "noarch: python" for package compatible with all platforms?

malfet: Good point, let me do it in the followup PR.

PR #421, opened 2021-09-15
allow filter memory events by size with a slider

JouleYuan:
allow filter memory events by size with a slider


tb_plugin/torch_tb_profiler/run.py
```diff
@@ -487,6 +494,7 @@ def get_op_name_or_ctx(record: MemoryRecord):
             "metadata": {
                 "title": "Memory Events",
                 "default_device": default_device,
+                "max_size": max_size
```
guotuofeng, line 497: could we statistic the max size only from frontend?

JouleYuan: Of course. But on the backend side, we can get the "max size" in a data scan loop that already exists. If we choose to obtain "max size" from the frontend side, a new iteration is needed, which brings overhead. Though the overhead would not be too much I think.

tb_plugin/fe/src/apifix.d.ts
```diff
@@ -44,6 +44,10 @@ declare module './api/generated/api' {
     default_device: MemoryTableMetadata['defaultDevice']
   }
 
+  export interface MemoryEventsTableMetadata {
```
guotuofeng, line 47: needless to change?

JouleYuan: Yes, I've removed this in the latest commit.

guotuofeng (approved):

PR #422, opened 2021-09-15
Enhancement of tensorcores

gaoteng-git:
1. Add Tensor Cores usage ratio of each GPU to overview
image

2. Update recommendation to kernel time ratio of bottom TC eligible ops
image

TODO: Add unit test in following PR.


tb_plugin/torch_tb_profiler/profiler/module_parser.py
```diff
@@ -120,9 +120,19 @@ def __init__(self):
...
-        self.tc_eligible_ops = None
+        self.tc_eligible_ops_kernel_ratio = 0.0
 
     def aggregate(self, tid2tree):
+        def sum_self_kernel_time(ops):
+            sum_time = 0
+            for op in ops:
+                for rt in op.runtimes:
+                    # "CallTreeRoot" & "dummy" kernels are launched out of profiler step, so don't count them.
+                    if op.name != "CallTreeRoot" and rt.name != "dummy":
```
guotuofeng, line 131: the kernels under dummy node will be ignored?

gaoteng-git: Yes. Because we can't get which operators launch those kernels of dummy node, so we can't know these kernels are launched by TC-eligible ops or none-TC-eligible ops.

tb_plugin/torch_tb_profiler/profiler/module_parser.py
```diff
@@ -157,4 +167,10 @@ def aggregate(self, tid2tree):
         self.op_list_groupby_name_input = list(agg_result[1].values())
         self.stack_lists_group_by_name = stack_lists_group_by_name
         self.stack_lists_group_by_name_input = stack_lists_group_by_name_input
-        self.tc_eligible_ops = sum([1 if op.tc_eligible else 0 for op in ops])
+        ops_bottom_tc_eligible = []
```
guotuofeng, line 170: better to move all the aggregation for tensorcore into a separated class. For example, TensorCoreAggregator?

tb_plugin/torch_tb_profiler/profiler/gpu_metrics_parser.py
```diff
@@ -31,6 +32,9 @@ def __init__(self):
         self.occupancy_per_device = [[] for _ in range(consts.MAX_GPU_PER_NODE)]
         self.avg_occupancy_per_device = [None] * consts.MAX_GPU_PER_NODE
         self.occupancy_count = [0] * consts.MAX_GPU_PER_NODE
+        # For calculating Tensor Cores time ratio.
+        self.kernel_per_device = [[] for _ in range(consts.MAX_GPU_PER_NODE)]
+        self.tc_ratio = [None] * consts.MAX_GPU_PER_NODE
```
guotuofeng, line 37: do we still need keep kernel parser tc_used_ratio?

tb_plugin/torch_tb_profiler/profiler/gpu_metrics_parser.py
```diff
@@ -151,6 +155,27 @@ def calculate_occupancy(self, steps_start_time, steps_end_time):
             if total_time > 0:
                 self.avg_occupancy_per_device[gpu_id] = total_occupancy / total_time
 
+    def calculate_tc_ratio(self, steps_start_time, steps_end_time):
```
guotuofeng, line 158: could we move all the logic to calculate tensor core ratio to a consolidated place instead of spread across multiple files?

tb_plugin/torch_tb_profiler/profiler/gpu_metrics_parser.py
gaoteng-git: Yes. tc_ratio is an array of every GPU ratio. tc_used_ratio is ratio of all GPUs' kernels.
tc_used_ratio is used to show pie chart in kernel view:
image

guotuofeng (approved):

Issue #426, opened 2021-09-16
Collect GPU data on ROCm Platforms


mwootton:
ROCm systems do not have cupti available.  GPU api calls and async ops need to be collected via the roctracer api.

aaronenyeshi:
Closing since roctracer is supported.

PR #427, opened 2021-09-16
Add roctracer support

mwootton:
Issue #426

This adds a RoctracerActivityInterface and hooks it into ActivityProfiler (when a rocm build is detected).
Uses GenericTraceActivity records to avoid muddying up the waters.
Build support in pytorch is already upstreamed - https://github.com/pytorch/pytorch/pull/58401


libkineto/src/ActivityProfiler.cpp
```diff
@@ -114,7 +117,11 @@ bool ActivityProfiler::applyNetFilterInternal(const std::string& name) {
 }
 
 // This has dependence on CuptiActivityInterface
+#ifdef HAS_ROCTRACER
+ActivityProfiler::ActivityProfiler(RoctracerActivityInterface& cupti, bool cpuOnly)
```
gdankel, line 121: Can we instead create a pure virtual class IDeviceActivityProfiler, with generic function names like enableActivities() and so on? This will be resolved with my ongoing refactoring, but would be nice to at least avoid additional #ifdefs like these.

libkineto/src/init.cpp
```diff
@@ -136,6 +136,12 @@ bool libkineto_init(bool cpuOnly, bool logOnError) {
   }
 #endif // HAS_CUPTI
 
+#ifdef HAS_ROCTRACER
+  if (!cpuOnly) {
+
```
gdankel, line 141: ?

libkineto/src/output_json.cpp
```diff
@@ -257,22 +261,56 @@ void ChromeTraceLogger::handleGenericActivity(
...
-  }},)JSON",
-      toString(op.type()), traceActivityJson(op),
-      // args
-      op.id,
-      op.traceSpan()->name, op.traceSpan()->iteration, separator,
-      op_metadata);
-  // clang-format on
 
+  switch (op.type()) {
+    case ActivityType::CUDA_RUNTIME:
```
gdankel, line 266: OK this is pretty hacky, but clearly GenericActivity is not generic enough yet.
Perhaps we can live with this for a bit but let's create an issue for fixing GenericTraceActivity.

gdankel (changes requested):

libkineto/libkineto_defs.bzl
```diff
@@ -35,6 +35,25 @@ def get_libkineto_srcs(with_api = True):
         "src/output_json.cpp",
     ] + (get_libkineto_api_srcs() if with_api else [])
 
+def get_libkineto_roctracer_srcs(with_api = True):
```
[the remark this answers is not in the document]
mwootton: Yeah, I think so.  I broke myself during a merge from upstream here previously.  (i.e. a new file was added to both _cpu_only and _cupti_srcs).  Better to have new files go in in exactly one place.

libkineto/src/ActivityProfiler.cpp
mwootton: Yeah, it's pretty clear that this stage of development wasn't really done with multiple backends in mind.  I was hoping the "refactor" would make this fit together more cleanly.  I didn't want to hijack the refactor... so there is some not-so-elegant stuff in there.  IMO there shouldn't be an '#ifdef Anything' in any of the files save once per header.  That should be handled in the build.  Cleaning this up will be an effort.  As you know, there is cupti code and cupti cruft everywhere.  E.g. ActivityProfiler (used by cpu, cupti, and roctracer) has cupti code all through it.  Ideally that would be ifdef-clean and cupti/roctracer concerns would be layered on with separate source files.

We could create this proposed interface now, but it means very little until we get the cupti concerns out of ActivityProfiler.  There will still be as many ifdefs (you could change some to dynamic_cast<> checks) :)  Keep in mind I didn't add any "ifdef cupti" to that file, those are all there from the original separation of concerns problem.  So those can't go away until a well thought-out refactor takes place.   As a simple example, to generalize enableActivity() I will have to decide on and define the actual Activities enum. Then go change all the existing code that uses the cupti provided enums.

So if I do that now then undeniably the refactor is creeping into a roctracer support PR.  Honestly, I was seeing this "not so elegant" integration as a tool for you to direct your refactoring.  Having the integration in place should produce a better refactor as the issues are real instead of hypothetical.

So the full refactor should not be in this PR, clearly.  We can pull this small and incomplete part of it in right now, but I'd personally rather not.  Your call.

libkineto/src/init.cpp
mwootton: Oh, that was a placeholder, didn't need anything there.  I'll remove, thanks.

libkineto/src/output_json.cpp
mwootton: Well, it has some endearing qualities:

1. it's in the right place (file), this is a json formatting concern.
2. I didn't have to make several subclasses of TraceActivity
3. I didn't have to bloat output_json.h with methods to pass those in
4. I didn't have ifdef those methods to keep the interface generic
5. The choice of metadata to output (a gpu concern) is made in RoctracerActivityInterface (the right place) rather than in output.json which shouldn't really shouldn't be privy to those details.

End of the day this switches off TraceActivity.type, while the cupti path switches of of typeid.  Six of one... half dozen of the other.

TraceActivity defines the ActivityType so it acknowledges there are different types of activity.  This function deals with only those types, that seems like fair game.
 
Granted this is a tricky pattern to get right from an OO and SOC point of view.  We'll figure something nice out.

libkineto/src/ActivityProfiler.cpp
gdankel: Fair. I was under the impression that CUPTI and the ROC tracer APIs were very similar, hence the comments. But I want to avoid more work for no gain obviously. What's your timeline for this, do you have a special reason to get this in for 1.10 or is it just as soon as possible?

libkineto/src/output_json.cpp
gdankel: Ah, I definitely think using the generic activity makes sense, I was just talking about the switch statement. I think we can get rid of that with some small changes.

libkineto/include/GenericTraceActivity.h
```diff
@@ -26,7 +26,7 @@ constexpr unsigned int kLinkFwdBwd = 1;
 class GenericTraceActivity : public TraceActivity {
 
  public:
-  GenericTraceActivity() = delete;
+  GenericTraceActivity() : activityType(ActivityType::ENUM_COUNT), traceSpan_(NULL) {}
```
gdankel, line 29: Do we need this constructor?

libkineto/src/ActivityProfiler.cpp
mwootton: Yes, the cupti and roctracer interfaces are superficially very similar.  It's in the details that implementations diverge.  And of course all the types, constants, etc. are going to have to be generalized.  That generalization looks silly when you have a single backend (hence you skipped it) but required when you have two or more.

Oh, indeed ActivityType is already generalized (contrary to what I said in the previous comment).   Good.  I got confused with the ActivityController bits that are handling cupti-specific stuff.

We would like to get this into 1.10 because we have some big installations using it.  We've been maintaining a 1.09 fork with this enabled.  Would really like to avoid that support drag and make it easier for the users to get.  And yeah, I'll would have to create the 1.10 fork personally. :P

libkineto/include/GenericTraceActivity.h
mwootton: I am using GenericTraceActivity for the gpu_api and gpu_ops coming out of RoctracerActivityInterface.  I didn't have a source for a TraceSpan.  Also, the cupti api and op activity do not have a TraceSpan, it is NULL.  The only available constructor takes a TraceSpan&.  Setting the type to ENUM_COUNT/invalid seemed proper.

Queue MoreGenericTraceActivity.

jithunnair-amd:
@gdankel Are there any updates on the internal Phabricator run? 
cc @mwootton

libkineto/include/GenericTraceActivity.h
```diff
@@ -73,7 +73,11 @@ class GenericTraceActivity : public TraceActivity {
 
   //Encode client side metadata as a key/value string.
   void addMetadata(const std::string& key, const std::string& value) {
```
gdankel, line 75: Why not instead add addMetadata(const std::string& key,  int value) ?

libkineto/src/RoctracerActivityInterface.h
```diff
@@ -0,0 +1,178 @@
...
+             , uint32_t tid, uint64_t begin, uint64_t end
+             , const void* ptr, size_t size)
+    : roctracerRow(id, domain, cid, pid, tid, begin, end)
+    , ptr(ptr), size(size) {}
+  const void *ptr;
+  size_t size;
+};
+
+
+class RoctracerActivityInterface {
```
gdankel, line 117: How about RoctracerActivityApi to avoid overloading interface? (I can change the cupti one).

libkineto/include/GenericTraceActivity.h
```diff
@@ -73,7 +73,11 @@ class GenericTraceActivity : public TraceActivity {
...
-    metadata_.push_back(fmt::format("\"{}\": {}", key, value));
+    if (!value.empty() && (value.find_first_not_of("-0123456789") == std::string::npos
```
gdankel, line 76: Nit: I think coding standard dictates use of curly brackets.

libkineto/src/RoctracerActivityInterface.h
```diff
@@ -0,0 +1,178 @@
...
+#ifdef HAS_ROCTRACER
+#define __HIP_PLATFORM_AMD__
+#include <roctracer.h>
+#include <roctracer_hcc.h>
+#include <roctracer_hip.h>
+#include <roctracer_ext.h>
+#include <roctracer_roctx.h>
+
+#endif
+#include <functional>
```
gdankel, line 24: Nit: system / STL headers on top.

libkineto/src/RoctracerActivityInterface.h
mwootton: I am in the middle of refactoring those classes.  Can we just defer the name change until after that?

libkineto/src/RoctracerActivityInterface.h
mwootton: Yeah sure

libkineto/include/GenericTraceActivity.h
mwootton: Done

libkineto/src/RoctracerActivityInterface.cpp
```diff
@@ -0,0 +1,568 @@
...
+#include "RoctracerActivityInterface.h"
+
+#include <cstring>
+#include <chrono>
+#include <time.h>
+
+#include <cxxabi.h>
+#include <sys/syscall.h>
+
+#include "Logger.h"
```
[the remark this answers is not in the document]
mwootton: Removed.

libkineto/src/RoctracerActivityInterface.cpp
```diff
@@ -0,0 +1,568 @@
...
+#include "output_base.h"
+
+typedef uint64_t timestamp_t;
+
+static inline uint32_t getPid() { return syscall(__NR_getpid); }
```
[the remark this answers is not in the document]
mwootton: You spoil me.  Done.

libkineto/src/RoctracerActivityInterface.cpp
```diff
@@ -0,0 +1,568 @@
...
+static inline uint32_t getTid() { return syscall(__NR_gettid); }
+
+// C++ symbol demangle
+static inline const char* cxx_demangle(const char* symbol) {
```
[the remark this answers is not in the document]
mwootton: Pattern forming here.  It's like we all hit the same issues.

libkineto/src/RoctracerActivityInterface.cpp
```diff
@@ -0,0 +1,568 @@
...
+}
+
+RoctracerActivityInterface::~RoctracerActivityInterface() {
+  disableActivities(std::set<ActivityType>());
+  endTracing();
+}
+
+void RoctracerActivityInterface::pushCorrelationID(int id, CorrelationFlowType type) {
+#ifdef HAS_ROCTRACER
+  if (!singleton().m_externalCorrelationEnabled)
```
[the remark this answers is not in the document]
mwootton: Fair enough.

libkineto/src/RoctracerActivityInterface.cpp
```diff
@@ -0,0 +1,568 @@
...
+    a.addMetadata("shared size", std::to_string(item.groupSegmentSize));
+    std::snprintf(buff, 4096, "%p", item.stream);
+    a.addMetadata("stream", std::string(buff));
+
+    // Stash launches to tie to the async ops
+    m_kernelLaunches[a.id] = a;
+
+    // Stash kernel names to tie to the async ops
+    std::string name;
+    if (item.functionAddr != NULL)
```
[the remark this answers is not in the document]
mwootton: #define NULL nullptr
nah ;)

libkineto/src/RoctracerActivityInterface.cpp
```diff
@@ -0,0 +1,568 @@
...
+      name = hipKernelNameRefByPtr(item.functionAddr, item.stream);
+    else if (item.function != NULL)
+      name = hipKernelNameRef(item.function);
+    if (name.empty() == false) {
```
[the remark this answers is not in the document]
mwootton: Changed.  This is less descriptive and can be confused with the integer operator (which should not be used at all).  But is saves previous characters that we need for extra { }.   :)

libkineto/src/output_json.cpp
```diff
@@ -236,7 +240,7 @@ void ChromeTraceLogger::handleGenericInstantEvent(
 }
 
 void ChromeTraceLogger::handleGenericActivity(
-    const libkineto::GenericTraceActivity& op) {
+    const GenericTraceActivity& op) {
```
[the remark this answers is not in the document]
mwootton: Yeah, seems to be the only occurrence of namespacing in the file.  Removed.

libkineto/src/RoctracerActivityInterface.cpp
```diff
@@ -0,0 +1,568 @@
...
+    else if (item.function != NULL)
+      a.addMetadata("kernel", hipKernelNameRef(item.function));
+    std::snprintf(buff, 4096, "(%d, %d, %d)", item.gridX, item.gridY, item.gridZ);
+    a.addMetadata("grid dim", std::string(buff));
+    std::snprintf(buff, 4096, "(%d, %d, %d)", item.workgroupX, item.workgroupY, item.workgroupZ);
+    a.addMetadata("block dim", std::string(buff));
+    a.addMetadata("shared size", std::to_string(item.groupSegmentSize));
+    std::snprintf(buff, 4096, "%p", item.stream);
+    a.addMetadata("stream", std::string(buff));
+
```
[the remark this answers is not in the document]
mwootton: Cool

libkineto/src/RoctracerActivityInterface.h
```diff
@@ -0,0 +1,177 @@
...
+  std::map<activity_correlation_id_t, uint32_t> m_kernelNames;
+
+  ApiIdList m_loggedIds;
+
+  // Api callback data
+  std::deque<roctracerRow> m_rows;
+  std::deque<kernelRow> m_kernelRows;
+  std::deque<copyRow> m_copyRows;
+  std::deque<mallocRow> m_mallocRows;
+  std::map<activity_correlation_id_t, GenericTraceActivity> m_kernelLaunches;
```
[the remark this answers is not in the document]
mwootton: Yuck, but done.  Postfix (reverse) scoping is so unreadable.  Not even python messes this up, e.g. _memberVar
string::std
radius.circle
size<-this
size<-d
Someone at google was nuts. or trolling. :)

libkineto/src/RoctracerActivityInterface.cpp
```diff
@@ -0,0 +1,554 @@
...
+    GenericTraceActivity a;
+    a.startTime = (item.begin + toffset) / 1000;
+    a.endTime = (item.end + toffset) / 1000;
+    a.id = item.id;
+    a.device = item.pid;
+    a.resource = item.tid;
+    a.activityType = ActivityType::CUDA_RUNTIME;
+    a.activityName = std::string(roctracer_op_string(ACTIVITY_DOMAIN_HIP_API, item.cid, 0));
+
+    a.addMetadata("ptr", fmt::format("%p", item.ptr));
```
gdankel, line 106: I believe this will simply print "%p" ?
Think you need to use the {} and fmt::ptr syntax.

mwootton: Yeah, got copy/paste happy.  Missed a step.  Hold please.

PR #437, opened 2021-09-22
fix a pandas dataframe usage bug

gaoteng-git:
Operate on a DataFrame column needs to use "map".

### DIFF
--- a/tb_plugin/torch_tb_profiler/profiler/kernel_parser.py
+++ b/tb_plugin/torch_tb_profiler/profiler/kernel_parser.py
@@ -17,7 +17,7 @@ class KernelParser:
         events = [vars(event) for event in events if event.type == EventTypes.KERNEL]
         events = pd.DataFrame(events)
         events = events.astype({"type": "category", "name": "string"}, copy=False)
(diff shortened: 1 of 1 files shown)


tb_plugin/torch_tb_profiler/profiler/kernel_parser.py
```diff
@@ -17,7 +17,7 @@ def parse_events(self, events):
         events = [vars(event) for event in events if event.type == EventTypes.KERNEL]
         events = pd.DataFrame(events)
         events = events.astype({"type": "category", "name": "string"}, copy=False)
-        events["tc_used"] = events["name"] in TC_Allowlist
+        events["tc_used"] = events["name"].map(lambda name: name in TC_Allowlist)
```
guotuofeng, line 20: is numpy.where better than map here?

gaoteng-git: Cool!
However, it can't work in this case. Changing to following will set all "tc_used" to False.
events["tc_used"] = np.where(events["name"] in TC_Allowlist, True, False)

guotuofeng (approved):

PR #440, opened 2021-09-30
Enable windows build for benchmark binary

larryliu0820:
Summary:
As titled, we are building PyTorch lite predictor benchmark binary on Windows so that it can be used for models running on Windows. This diff adds the ability to build on Windows for a lot of dependencies of `xplat/caffe2/fb/lite_predictor:lite_predictor`. There are a few things to pay attention to:

* This diff excludes `unet_106_pt` custom ops from lite_predictor because it depends on `NNPACK` and it's not compatible with Windows. I have a separate PR to enable Windows build on `NNPACK` but there's no plan to merge it in.

* In `arvr/libraries/bolt_analyzer/bolt/analyzer/Tracer-default.cpp` I excluded `BoltPerformanceCounters.h` from include list, since it doesn't build for `is_arvr_mode() = False` which is the case here since we are building from xplat.

* In `xplat/pytorch/vision/torchvision/csrc/vision.cpp`, since `PyMODINIT_FUNC` depends on `Python.h` I added another condition to the macros.

Differential Revision: D31295975


libkineto/src/ActivityLoggerFactory.h
```diff
@@ -12,7 +12,9 @@
 #include <functional>
 #include <map>
 #include <string>
-
+#if defined(_MSC_VER)
```
malfet, line 15: Can you please explain why `cctype` is needed, as Kineto can already be built fine with VC++?

larryliu0820: That is to enable `std::transform` and `std::tolower`. I'm not sure why it builds with VC++ before.

malfet: `std::transform` is defined in algorithm, and `std::tolower` in indeed in `cctype`, but why does it need to be gated behind `#ifdef _MSC_VER`

PR #455, opened 2021-10-27
Add CUPTI Range Profiler Interface

briancoutinho:
Summary:
# CUPTI Range Profiler API Support
Adds new core logic to leverage CUPTI Range Profiler API.
Note that this adds a dependency on nvperf_host.so library that was introduced in later CUDA versions
The range based profiler allows us to calculate any number of profiler metrics for a region of the code. This region is known as a range.

## User vs Auto Range
The profiler can be enabled in an auto-range mode where each kernel automatically becomes a range and configured metrics are collected for it. The kernel executions in auto range mode are serialized.
In user range mode we can push/pop user range names and the metric is collected between each range stack. User range mode supports kernel concurrency.

User range example
```
profiler.pushRange("iter0");
....
profiler.popRange();
```

Reference : https://docs.nvidia.com/cupti/Cupti/r_main.html#r_profiler

# Implementation Details
1. CuptiProfilerAPI.h :
      CuptiRBSerssion -> this object contains all of the state and configuration data for CUPTI profiler. It performs counter configuration and returns results in the CuptiProfilerResults datastructure.
2. CuptiNvPerf.cpp : Adds helper code using NV PerfWorks APIs to configure and evaluate metrics from the range profiler.

Differential Revision: D31437227

### DIFF
--- a/libkineto/libkineto_defs.bzl
+++ b/libkineto/libkineto_defs.bzl
@@ -17,6 +17,7 @@ def get_libkineto_cupti_srcs(with_api = True):
         "src/CuptiCallbackApi.cpp",
         "src/CuptiEventApi.cpp",
         "src/CuptiMetricApi.cpp",
+        "src/CuptiRangeProfilerApi.cpp",
         "src/Demangle.cpp",
         "src/EventProfiler.cpp",
         "src/EventProfilerController.cpp",
--- a/libkineto/libkineto_defs.bzl
+++ b/libkineto/libkineto_defs.bzl
@@ -14,6 +14,7 @@ def get_libkineto_cupti_srcs(with_api = True):
         "src/CudaDeviceProperties.cpp",
         "src/CuptiActivityApi.cpp",
         "src/CuptiActivityPlatform.cpp",
+        "src/CuptiCallbackApi.cpp",
         "src/CuptiEventApi.cpp",
         "src/CuptiMetricApi.cpp",
         "src/Demangle.cpp",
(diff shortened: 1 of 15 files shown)


sophiaprincess002:
> Summary:
> 
> # CUPTI Range Profiler API Support
> Adds new core logic to leverage CUPTI Range Profiler API. Note that this adds a dependency on nvperf_host.so library that was introduced in later CUDA versions The range based profiler allows us to calculate any number of profiler metrics for a region of the code. This region is known as a range.
> 
> ## User vs Auto Range
> The profiler can be enabled in an auto-range mode where each kernel automatically becomes a range and configured metrics are collected for it. The kernel executions in auto range mode are serialized. In user range mode we can push/pop user range names and the metric is collected between each range stack. User range mode supports kernel concurrency.
> 
> User range example
> 
> ```
> profiler.pushRange("iter0");
> ....
> profiler.popRange();
> ```
> 
> Reference : https://docs.nvidia.com/cupti/Cupti/r_main.html#r_profiler
> 
> # Implementation Details
> 1. CuptiProfilerAPI.h :
>    CuptiRBSerssion -> this object contains all of the state and configuration data for CUPTI profiler. It performs counter configuration and returns results in the CuptiProfilerResults datastructure.
> 2. CuptiNvPerf.cpp : Adds helper code using NV PerfWorks APIs to configure and evaluate metrics from the range profiler.
> 
> Differential Revision: D31437227

PR #461, opened 2021-11-08
add PYTHON_FUNCTION activity type

robieta:
Add new event type for https://github.com/pytorch/pytorch/pull/67407

I patched this change into the `third_party/kineto` folder of my pytorch checkout and events still show up on the chrome trace. (With just a couple tweaks.) So we can land this and then I'll update the submodule ref in pytorch to pick it up before we land the tracer.


chaekit (approved):

gdankel (commented):
Looks good - note that it will also have to be exposed through PyTorch profiler API separately at the moment.

gdankel (approved):

PR #462, opened 2021-11-09
add module view support

guotuofeng:
Add module view event and node so that the module view data can be constructed.


chaekit:
Trying to understand the event flow here
1. so I am assuming that https://github.com/pytorch/kineto/pull/462 will generate the .pbtxt on Tensorboard launch
2. Tensorboard will try to load kineto_plugin
3. It will then use this change to convert .pbtxt --> blob that graph plugin can understand

I can see some ordering issue here where if graph plugin is loaded before kineto plugin, it won't be able to scan the converted blob? is this possible? cc: @Reubend

tb_plugin/torch_tb_profiler/profiler/module_op.py
```diff
@@ -0,0 +1,220 @@
...
+
+
+class Module:
+    def __init__(self, name, module_id):
+        self.name = name
+        self.module_id = module_id
+        self.children = []
+
+    def __hash__(self):
+        if self.children:
```
chaekit, line 14: nit: not too familiar with Python but what happens if you just `tuple([])` i.e. empty list? will it just crash?

tb_plugin/torch_tb_profiler/run.py
```diff
@@ -505,6 +507,38 @@ def get_op_name_or_ctx(record: MemoryRecord):
...
+                {"name": "Host Total Time", "type": "number", "key": "host_duration"},
+                {"name": "Host Self Time", "type": "number", "key": "self_host_duration"},
+                {"name": "Device Total Time", "type": "number", "key": "device_duration"},
+                {"name": "Device Self Time", "type": "number", "key": "self_device_duration"}
+            ],
+            "data": []
+        }
+
+        def process_modules_stats(parent, modules_stats):
+          for stats in modules_stats:
```
chaekit, line 525: indentation?

tb_plugin/torch_tb_profiler/profiler/module_op.py
```diff
@@ -0,0 +1,220 @@
...
+            return False
+
+        return (self.name == o.name and
+                self.module_id == o.module_id and
+                self.children == o.children)
+
+class ModuleStats:
+    def __init__(self, name, module_id):
+        self.name = name
+        self.module_id = module_id
```
chaekit, line 30: `module_id` should be distinct so not sure how you can have multiple `occurences`? i.e. you'll see one `Linear_0` and not like 3 `Linear_0` s

tb_plugin/torch_tb_profiler/profiler/module_op.py
```diff
@@ -0,0 +1,220 @@
...
+        self.occurences = 0
+        self.operators = 0
+        self.host_duration = 0
+        self.device_duration = 0
+        self.self_host_duration = 0
+        self.self_device_duration = 0
```
chaekit, line 36: oh wow didn't know that we can get the corresponding device durations as well! 😮

tb_plugin/torch_tb_profiler/profiler/module_op.py
```diff
@@ -0,0 +1,220 @@
...
+
+class ModuleStats:
+    def __init__(self, name, module_id):
+        self.name = name
+        self.module_id = module_id
+        self.occurences = 0
+        self.operators = 0
+        self.host_duration = 0
+        self.device_duration = 0
+        self.self_host_duration = 0
```
chaekit, line 35: nit: i'd name these `self_total_host_duration` since its an accumulated value

tb_plugin/torch_tb_profiler/profiler/module_op.py
```diff
@@ -0,0 +1,220 @@
...
+        self.self_device_duration = 0
+
+    @property
+    def avg_host_duration(self):
```
chaekit, line 39: don't matter for this PR but would be good to have full list of basic stats like min, max, std dev, median in the future 👍

tb_plugin/torch_tb_profiler/profiler/module_op.py
```diff
@@ -0,0 +1,220 @@
...
+    if modules and roots:
+        return _process_module_statistics(modules, roots)
+    else:
+        return None
+
+def _build_module_hierarchy(events):
+    '''Get the module hierarchy from the chome trace events
+    '''
+    python_events = [e for e in events if e.type in (EventTypes.PYTHON_FUNCTION, EventTypes.MODULE)]
+    python_events.sort(key=lambda e: e.python_id)
```
chaekit, line 59: not sure if we need to sort these by id's first?

tb_plugin/torch_tb_profiler/profiler/module_op.py
```diff
@@ -0,0 +1,220 @@
...
+        children.setdefault(e_parent_id, [])
+        children[e_parent_id].append(e_id)
+    function_leaves = [k for k, v in children.items() if not v]
+
+    # Convert Python function topology to Module topology.
+    #   This is a simple O(n) tree walking algorithm where we start from the leaves
+    #   and walk up, discarding any nodes which are not Module nodes.
+    module_parent_map = {}
+    seen = set()
+    for i in function_leaves:
```
chaekit, line 77: nit: usually good to have some tests around testable algorithms like this. not a blocker

tb_plugin/torch_tb_profiler/profiler/trace.py
```diff
@@ -23,6 +23,8 @@ class EventTypes(object):
     MEMSET = "Memset"
     PYTHON = "Python"
     MEMORY = "Memory"
+    PYTHON_FUNCTION = "python_function"
```
chaekit, line 26: so whats the difference between `Python" on line 24 and our `python_function` ? the names could be very confusing to those new to this code

tb_plugin/torch_tb_profiler/profiler/node.py
```diff
@@ -224,6 +250,7 @@ def create(cls, event):
             kwargs["device_id"] = event.device_id
         return cls(**kwargs)
 
+
```
chaekit, line 253: I know its draft but empty line 😃

tb_plugin/torch_tb_profiler/profiler/module_op.py
```diff
@@ -0,0 +1,220 @@
+from copy import copy
+
+from .node import ModuleNode, ProfilerStepNode
+from .trace import EventTypes
+
+
+class Module:
```
chaekit, line 7: do we need a different data structure for the ops themselves?

tb_plugin/torch_tb_profiler/profiler/module_op.py
guotuofeng: if the tuple is empty list, then the bool clause will be false

tb_plugin/torch_tb_profiler/run.py
guotuofeng: good catch

tb_plugin/torch_tb_profiler/profiler/module_op.py
guotuofeng: the occurences happens across multiple steps. For example, the Sequential_0 repeat 10 times.

tb_plugin/torch_tb_profiler/profiler/module_op.py
guotuofeng: What do you mean id? should not it be the python_id?

tb_plugin/torch_tb_profiler/profiler/module_op.py
guotuofeng: will add testcases when it is ready to review

tb_plugin/torch_tb_profiler/profiler/trace.py
guotuofeng: the Python in 24 is used internally in tb_plugin. maybe we need some better name.

tb_plugin/torch_tb_profiler/profiler/node.py
guotuofeng: https://www.python.org/dev/peps/pep-0008/#blank-lines

tb_plugin/torch_tb_profiler/profiler/module_op.py
guotuofeng: sorry, I don't quite understand your question.

tb_plugin/torch_tb_profiler/profiler/module_op.py
guotuofeng: the self_host_duration is the duration not contains any children ops. the host_duration is for total host time. 
So the self and total is contrary words.

chaekit (approved):

tb_plugin/torch_tb_profiler/profiler/module_op.py
guotuofeng: will remove the sort.

PR #471, opened 2021-11-19
Remove a couple of noisy warnings

gdankel:
Summary:
Libkineto is currently logging too many messages. We should change some INFO and WARNING to VLOG.
In some cases we should also suppress common errors that we handle correctly.

Reviewed By: aaronenyeshi

Differential Revision: D31665028


aaronenyeshi:
Closing, and will re-open on fixing merge issues.

PR #473, opened 2021-11-19
Don't reset verbose level

gdankel:
Summary: If setting verbose log level via on-demand request, just leave it on permanently and leave it to the user to turn it off again. It can be annoying and confusing that it's automatically disabled after some time.

Reviewed By: briancoutinho

Differential Revision: D31665029


aaronenyeshi:
Closing, and will re-open on fixing merge issues.

PR #479, opened 2021-12-02
Make sample program compile

r-barnes:
The sample program didn't have any compilation stuff associated with it.

This PR fixes that; however, the sample program segfaults when it runs. I think it needs some attention and probably needs to be in a contbuild or some such.


aaronenyeshi:
Closing as sample_program was updated in https://github.com/pytorch/kineto/pull/812.

PR #487, opened 2021-12-13
fix memoryview bug that UI not refreshed

guotuofeng:
fix memoryview bug that UI not refreshed when switching span after selecting a range


cloudhan:
LGTM!

PR #512, opened 2022-02-01
Simplify Logger observers

swolchok:
Summary: This seems like a simpler way to set up the kineto Logger observers. Unlike a previous attempt to do it this way, we 1) suppress the -Wglobal-constructors diagnostic and 2) work around SIOF by putting our global singleton set on the heap.

Reviewed By: aaronenyeshi

Differential Revision: D33690834


aaronenyeshi (approved):
LGTM, thanks for the improvements in this patch, and checking tests pass on Github Actions.

PR #513, opened 2022-02-01
kineto] enable iteration tracking and trace trigger using it

briancoutinho:
Summary:
## Overview

Kineto traces can be triggered asynchronously using Config files. This method only supports specifying a trace start time and trace duration.
Enabling tracing based on the iteration count rather than a combination of {start time, duration} has several added advantages
1. Tools triggering traces externally can obtain traces containing whole iterations i.e  they do not need to tune the duration based on the iteration time
2. This provides feature parity between the pytorch profiler and external interface to Kineto.

## Implementation

We add new Config options for profiling using iteration counts
```
PROFILE_START_ITERATION=200   // iteration to start profile
ACTIVITIES_WARMUP_ITERATIONS=1   // iterations to warmup BEFORE the profiler start iteration above
ACTIVITIES_ITERATIONS=3  // number of iterations to profile
```

Further, we add a 'step()' method into Kineto's public interface. This will do two things :
1) increment the iteration count internally.
2) if an external config is present it can evaluate if we need to start/warmup/complete the traces.

The step method can then be inserted into the AI framework layer such as within the pytorch profiler step() implementation.

## Additional Details

What happens if **current iteration  > (start - warmup iters)**?  In this case, the behavior is to move the start to the next iteration boundary. Profiler then does the warmup and profile capture.

**Avoiding Race conditions:**
Since we are introducing a second method to trigger traces we need to update the state of profiler in both paths (step() method and the periodic profiling thread). So far we have ensured there are no races between two threads updating the state, thus avoiding a lock. Depending on if the config uses iteration count or profile start time state transitions are mainly done via one of the two paths only. Let me know if that makes sense.

Differential Revision: D33027010

### DIFF
--- a/libkineto/include/ActivityProfilerInterface.h
+++ b/libkineto/include/ActivityProfilerInterface.h
@@ -58,6 +58,10 @@ class ActivityProfilerInterface {
     return nullptr;
   }
 
+  // Re-evaluate internal state to allow for triggering operations based
+  // on number of iteration. each implicitly increments the iteration count
+  virtual void step() {}
+
   // *** TraceActivity API ***
   // FIXME: Pass activityProfiler interface into clientInterface?
   virtual void pushCorrelationId(uint64_t id){}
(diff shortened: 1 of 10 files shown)


aaronenyeshi (approved):
LGTM!

PR #527, opened 2022-02-14
Fix Config update logic in prepareTrace()

briancoutinho:
Summary:
This patch tries to retain the logic as it was before
https://github.com/pytorch/kineto/pull/521/commits/fae58ea950489a3d2e31e0030e09cb97ee0bc9f7

In other words, once we receive a synchronous request
1. Create a Config object
2. Set client defaults
3. set selected activties
4. Validate the config

Previous patch moved the validate step above 2 and 3, it is possible this caused some segfaults due to a hidden assumption.

Differential Revision: D34220877


aaronenyeshi (approved):
LGTM, thanks for fixing the bug in the CI

PR #540, opened 2022-02-18
Fix roctracerActivityBuffer activity flow id to be correlation_id

aaronenyeshi:
Summary: This looks like a typo from copy and paste, since item does not exist. The GenericTraceActivity's flow id should be the record's correlation_id saved in the buffer.

Reviewed By: briancoutinho

Differential Revision: D34321823

Pulled By: aaronenyeshi


aaronenyeshi:
This is landed, not sure why it didn't close.

PR #544, opened 2022-02-23
Enable CUPTI Profiler API to track multiple threads/context

briancoutinho:
Summary:
# Summary

The CUPTI Ranger profiler API is confined to each CUDA context. This makes it harder to use in applications that are multi-threaded, such as pytorch models
with different threads for forward and backward pass computation.

In this patch we leverage CUPTI Callback API to track the CUDA contexts and kernel launches.
* Also, introduce an asynchronous mode to start and stop the CUPTI Range profiling API that is triggered using Kernel launch callbacks
* Track the name of kernels to use with auto-range model as CUPTI only provides a number for range ID.

## Testing
1. Enabled mocking the callbacks for Cuda context create/destroy and kernel launches.
2. Added a unit test to check callback triggered logic and asynchronous start and stop of ranges.

Differential Revision: D32008907

### DIFF
--- a/libkineto/src/CuptiCallbackApi.cpp
+++ b/libkineto/src/CuptiCallbackApi.cpp
@@ -8,7 +8,9 @@
 #include <mutex>
 #include <shared_mutex>
 
+#ifdef HAS_CUPTI
 #include "cupti_call.h"
+#endif
 #include "Logger.h"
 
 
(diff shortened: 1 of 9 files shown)


aaronenyeshi (approved):
LGTM, see changes internal

PR #545, opened 2022-02-23
Enable rounding up profile start iteration

briancoutinho:
Summary:
## Summary
In cases where you need to remotely trigger a trace having a fixed profiler start iteration can be difficult.
You may want multiple distributed ranks (GPUs/servers) to capture same set of iterations. One idea to achieve synchrony
is to have profiler iterations that are a multiple of some integer. This reduces the probability of network delays causing mismatch in profiled iterations.

## Details
1. Add PROFILE_START_ITERATION_ROUNDUP config parameter.
2. If current iterations > start iteration, round up to nearest multiple of PROFILE_START_ITERATION_ROUNDUP

Reviewed By: aaronenyeshi

Differential Revision: D34248399

### DIFF
--- a/libkineto/include/Config.h
+++ b/libkineto/include/Config.h
@@ -226,8 +226,20 @@ class Config : public AbstractConfig {
     return profileStartIteration_ >= 0 && activitiesRunIterations_ > 0;
   }
 
-  void setProfileStartIteration(int64_t iter) {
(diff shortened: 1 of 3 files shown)


aaronenyeshi (approved):
LGTM! See comments internal

PR #553, opened 2022-03-04
Change output_membuffer to copy and own GenericTraceActivity objects

mwootton:
This is a fix for a break at https://github.com/pytorch/kineto/pull/464
This change was adding pointers to temporaries (stack addresses) to the list of output events.  Specifically it was taking a const &, assuming it was referring to a heap object, deciding it owned it now, and ignoring any life-cycle concerns.

This change mirrors the handling used for GpuActivity.
1) change ActivityLogger::handleGenericActivity() to actually take a GenericTraceActivity instead of the abstract base.
2) concrete GenericTraceActivity can now be used with make_unique<>.
3) use MemoryTraceLogger::addActivityWrapper (as GpuActivity does), which copies the object and maintains lifecycle via wrappers_
4) CLEANUP: added a ActivityLogger::handleActivity(cont ITraceActivity) in case someone actually wants to log an abstract base event (at their peril)

### DIFF
--- a/libkineto/src/output_base.h
+++ b/libkineto/src/output_base.h
@@ -68,8 +68,10 @@ class ActivityLogger {
 
   virtual void handleTraceSpan(const TraceSpan& span) = 0;
 
-  virtual void handleGenericActivity(
+  virtual void handleActivity(
       const libkineto::ITraceActivity& activity) = 0;
+  virtual void handleGenericActivity(
+      const libkineto::GenericTraceActivity& activity) = 0;
 
 #ifdef HAS_CUPTI
   virtual void handleGpuActivity(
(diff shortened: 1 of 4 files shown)


mwootton:
This is needed for https://github.com/pytorch/pytorch/issues/73588

aaronenyeshi:
Thanks for the patch @mwootton ! Importing internally to run full CI tests.

aaronenyeshi:
Part of the future refactoring patches, looks like all CUPTI activities are moving to handleActivity(const ITraceActivity). At the current head, only RuntimeActivity and OverheadActivity and not yet GPUActivity.

Issue #556, opened 2022-03-08
GitHub detected vulnerabilities in tb_plugin files


aaronenyeshi:
Please help take a look at these automatically detected vulnerabilities.

pytorch/kineto: GitHub detected a vulnerability in the node-forge dependency
```
GitHub has detected that a package defined in the tb_plugin/fe/yarn.lock file of the pytorch/kineto repository contains a security vulnerability.

Package name: node-forge
Affected versions: < 1.0.0
Fixed in version: 1.0.0
Severity: LOW

Identifier(s):
GHSA-gf8q-jrpm-jvxq

Reference(s):
https://github.com/digitalbazaar/forge/security/advisories/GHSA-gf8q-jrpm-jvxq
https://github.com/advisories/GHSA-gf8q-jrpm-jvxq
```

pytorch/kineto: GitHub detected a vulnerability in the ansi-regex dependency
```
GitHub has detected that a package defined in the tb_plugin/fe/yarn.lock file of the pytorch/kineto repository contains a security vulnerability.

Package name: ansi-regex
Affected versions: > 2.1.1, < 5.0.1
Fixed in version: 5.0.1
Severity: MODERATE

Identifier(s):
GHSA-93q8-gq69-wqmw
CVE-2021-3807

Reference(s):
https://nvd.nist.gov/vuln/detail/CVE-2021-3807
https://github.com/chalk/ansi-regex/commit/8d1d7cdb586269882c4bdc1b7325d0c58c8f76f9
https://huntr.dev/bounties/5b3cf33b-ede0-4398-9974-800876dfd994
https://github.com/chalk/ansi-regex/issues/38#issuecomment-924086311
https://app.snyk.io/vuln/SNYK-JS-ANSIREGEX-1583908
https://github.com/chalk/ansi-regex/issues/38#issuecomment-925924774
https://github.com/advisories/GHSA-93q8-gq69-wqmw
```

pytorch/kineto: GitHub detected a vulnerability in the ansi-html dependency
```
GitHub has detected that a package defined in the tb_plugin/fe/yarn.lock file of the pytorch/kineto repository contains a security vulnerability.

Package name: ansi-html
Affected versions: <= 0.0.7
Fixed in version: (No fix version known)
Severity: HIGH

Identifier(s):
GHSA-whgm-jr23-g3j9
CVE-2021-23424

Reference(s):
https://nvd.nist.gov/vuln/detail/CVE-2021-23424
https://github.com/Tjatse/ansi-html/issues/19
[...]

PR #557, opened 2022-03-08
fix yarn.lock vulnerabilities

guotuofeng:
Fix https://github.com/pytorch/kineto/issues/556


aaronenyeshi:
Thank you for the quick update, will keep you posted!

aaronenyeshi:
Landed!

Issue #504, opened 2022-01-19
torch_tb_profiler memory_parser throws an error


jyyon:
When running model training with deepspeed and pytorch profiler, the training runs without exception, but when I try to analyze the profiling result with tensorboard, it throws `assert prev_ts < r.ts` error from memory parser. I didn't use deepspeed model parallel flag and ran it on single machine with 4 GPUs.

deepspeed version = 0.5.9
torch version = 1.10.1
torch_tb_profiler version = 0.3.1

Training code:
```
def train_shard(model, dataset_provider, profile):
    ...
    dataset_iterator, total_length = dataset_provider.get_shard()
    for batch_num, batch in enumerate(tqdm(dataset_iterator)):
        ...
        # Wherever model.step is called, call profiler.step. The model is deepspeed initialized.
        model.step()
        profiler.step()

def main():
    ...
    with torch.profiler.profile(
            activities=[ProfilerActivity.CPU, ProfilerActivity.CUDA],
            schedule=torch.profiler.schedule(
                skip_first=2,
                wait=1,
                warmup=1,
                active=3,
                repeat=2
            ),
            on_trace_ready=tensorboard_trace_handler('path/log/fine_tune_logs'),
            profile_memory=True,
            with_stack=True,
            with_flops=True,
    ) as prof:
        while True:
            complete = train_shard(model, dataset_provider, prof)
            if complete:
                break
```

Error thrown when I ran tensorboard:
```
W0112 16:20:05.917469 4567412160 loader.py:100] Failed to parse profile data for Run fine_tune_logs on ip-x-x-x-x.ec2.internal. Exception=
Traceback (most recent call last):
  File "/Users/user/Library/Python/3.8/lib/python/site-packages/torch_tb_profiler/profiler/loader.py", line 86, in _process_data
    data, trace_path = RunProfileData.parse(worker, span, local_file)
  File "/Users/user/Library/Python/3.8/lib/python/site-packages/torch_tb_profiler/profiler/data.py", line 77, in parse
    profile = RunProfileData.from_json(worker, span, trace_json)
  File "/Users/user/Library/Python/3.8/lib/python/site-packages/torch_tb_profiler/profiler/data.py", line 97, in from_json
    profile.process()
  File "/Users/user/Library/Python/3.8/lib/python/site-packages/torch_tb_profiler/profiler/data.py", line 217, in process
    self.memory_parser = MemoryParser(self.tid2tree, memory_events)
[...]

PR #560, opened 2022-03-11
Fix memory analysis crash when there is multiple free with same address

guotuofeng:


Youcantstopme2744:
> Fix https://github.com/pytorch/kineto/issues/504#issuecomment-1062014641
> 
> Root cause: there are multiple freed memory with same address. With this fix, the last freed memory would win when pairing with allocation.

Youcantstopme2744:
Gix

PR #571, opened 2022-03-30
fix : add default activity types list for configs

briancoutinho:
Summary:
Currently the default activity list includes all activity types. This can be an issue if we want optional or experimental activity types.
This changes updates the Activity type enum to have a default and an optional section.
* Adds a `defaultActivityTypes()` function as well.
* Fix child profiler message so we know if it actually was initialized.

Reviewed By: aaronenyeshi

Differential Revision: D35218132

### DIFF
--- a/libkineto/include/ActivityType.h
+++ b/libkineto/include/ActivityType.h
@@ -7,7 +7,10 @@
 
 namespace libkineto {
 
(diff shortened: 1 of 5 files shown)


aaronenyeshi (approved):
LGTM, thank you!

PR #572, opened 2022-03-31
Fix potential Roctracer thread safety issue

mwootton:
I found some evidence the api callback function is not being serialized by libroctracer as was initially expected.
Protected the data structures.


aaronenyeshi:
Looks good, importing this patch

aaronenyeshi (approved):
LGTM, landing

PR #575, opened 2022-04-04
[kineto] Add fixes for pytorch build CI

briancoutinho:
Adding fixes for pytorch CI. Overall this add new files for CUPTI Range Profiler but disables out some functionality.

## Fixe also include
1. Update structure initialization to be complete, fails on gcc compilers
2. Remove unnecessary header <unistd> that fails on windows
3. Add USE_CUPTI_RANGE_PROFILER define to prevent some code from being included

In the future, we will need to be able to link nvperf_host.so in the build , this is required for range profiler to work.


aaronenyeshi (approved):
LGTM, thanks a lot for fixing the CI build failures.

PR #593, opened 2022-04-27
Use lowercase for EventTypeMap keys and checks in TB Plugin

aaronenyeshi:
Recently there was a patch that changed the trace format. It was previously Kernel, Memcpy, and Memset, but these were changed to kernel, gpu_memcpy and gpu_memset (following https://github.com/pytorch/kineto/blob/cdfa44f8d8f2f2b5c6f2efa83322977551e05d92/libkineto/src/ActivityType.cpp#L20). There are also a few new fields added, offending diff: https://github.com/pytorch/kineto/commit/1679e50eebaf47d7d02903ae2069faee0a5d2db6 . To support kernel, we will lowercase all the EventTypeMap keys and the check.

### DIFF
--- a/tb_plugin/torch_tb_profiler/profiler/trace.py
+++ b/tb_plugin/torch_tb_profiler/profiler/trace.py
@@ -33,17 +33,17 @@ class EventTypes(object):
 
 
 EventTypeMap = {
-    'Trace': EventTypes.TRACE,
+    'trace': EventTypes.TRACE,
     'cpu_op': EventTypes.OPERATOR,
-    'Operator': EventTypes.OPERATOR,
-    'Runtime': EventTypes.RUNTIME,
-    'Kernel': EventTypes.KERNEL,
-    'Memcpy': EventTypes.MEMCPY,
+    'operator': EventTypes.OPERATOR,
+    'runtime': EventTypes.RUNTIME,
+    'kernel': EventTypes.KERNEL,
+    'memcpy': EventTypes.MEMCPY,
     'gpu_memcpy': EventTypes.MEMCPY,
-    'Memset': EventTypes.MEMSET,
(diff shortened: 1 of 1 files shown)


aaronenyeshi:
@guotuofeng , can you help check that the new fields didn't break anything on your side?

guotuofeng:
We should add both "kernel" and "Kernel" to keep back-compatibility. Alternatively, we can change all the keys to lower case and then change the create_trace_event like the following:
```python
     event_type = EventTypeMap.get(category.lower())
```

aaronenyeshi:
> We should add both "kernel" and "Kernel" to keep back-compatibility. Alternatively, we can change all the keys to lower case and then change the create_trace_event like the following:
I've added the change to make it category.lower()

guotuofeng:
> > We should add both "kernel" and "Kernel" to keep back-compatibility. Alternatively, we can change all the keys to lower case and then change the create_trace_event like the following:
> 
> I've added the change to make it category.lower(). I'm not sure which other files to change (and where we check keys in TB), do you want to take over this patch?

We need lower case all the keys in EventTypeMap dictionary as well.

aaronenyeshi:
I've update the diff to your review, thanks! @guotuofeng

guotuofeng:
LGTM.

aaronenyeshi:
@guotuofeng - do you know the release cadence of tensorboard plugin? Will this make it into pytorch 1.13? CC @malfet

guotuofeng:
> @guotuofeng - do you know the release cadence of tensorboard plugin? Can we include this in the next tensorboard shipment? It is causing missing GPU Summary and Util information for some users CC @malfet

I think it is safe to release this PR with next shipment since the change is safe.

PR #598, opened 2022-05-17
Allow external pytorch libraries to see kineto

astachowiczhabana:
Summary: External activity profilers needs access to libkineto::api() to register.
kineto_api uses hidden CXX_VISIBILITY_PRESET. This blocks using
libkineto::api() symbol in scenario where kineto is linked statically
to pytorch (default case) and the external profiler is inside pytorch plugin.


aaronenyeshi (approved):
LGTM! Thanks

PR #599, opened 2022-05-19
Avoid deprecation warnings due to header file location change

mwootton:
Roctracer include files are moving in the future.  The current location will soon emit unsightly deprecation warnings.

- Change include path in cmake
- Explicitly include mutex.h in RoctracerActivityApi.h


aaronenyeshi (approved):
LGTM, which ROCm version does the include file exist at rocm-X.X.X/include/roctracer?

mwootton:
Well, the include files are in both places in all versions right now.  The files are duplicated, not linked.  Looks like someone is addressing this.  As of ROCm5.3 the copies in /opt/rocm/roctracer/include include a deprecation #pragma message.  I assume they will be gone altogether within a few releases.

Also, I missed the "#include <mutex>" previously and it was getting included indirectly from someplace.

malfet (approved):

PR #607, opened 2022-05-31
add device to RunProfile

krystianMoras:
It could be useful in future to have in one place and cleaner than checking every time


aaronenyeshi:
This is a tb_plugin change, CC @guotuofeng and others to help review.

aaronenyeshi (approved):

PR #611, opened 2022-06-01
Add HPU Activity Type

astachowiczhabana:
Summary: This change is part of Habana Device enablement in PyTorch.
Allows user to select HPU as profiling activity.

### DIFF
--- a/libkineto/include/ActivityType.h
+++ b/libkineto/include/ActivityType.h
@@ -22,6 +22,7 @@ enum class ActivityType {
     CPU_INSTANT_EVENT, // host side point-like events
     PYTHON_FUNCTION,
     OVERHEAD, // CUPTI induced overhead events sampled from its overhead API.
+    HPU_OP,
 
     // Optional Activity types
     GLOW_RUNTIME, // host side glow runtime events
--- a/libkineto/include/ActivityType.h
+++ b/libkineto/include/ActivityType.h
@@ -22,11 +22,11 @@ enum class ActivityType {
     CPU_INSTANT_EVENT, // host side point-like events
     PYTHON_FUNCTION,
     OVERHEAD, // CUPTI induced overhead events sampled from its overhead API.
-    HPU_OP,
 
     // Optional Activity types
     GLOW_RUNTIME, // host side glow runtime events
     CUDA_PROFILER_RANGE, // CUPTI Profiler range for performance metrics
+    HPU_OP,  //HPU host side runtime event
 
     ENUM_COUNT, // This is to add buffer and not used for any profiling logic. Add your new type before it.
     OPTIONAL_ACTIVITY_TYPE_START = GLOW_RUNTIME,
--- a/libkineto/src/output_json.cpp
+++ b/libkineto/src/output_json.cpp
@@ -186,9 +186,6 @@ void ChromeTraceLogger::handleTraceSpan(const TraceSpan& span) {
   if (!traceOf_) {
     return;
   }
-  if (span.startTime == 0 && span.endTime == 0 && span.iteration == -1) {
-    return;
-  }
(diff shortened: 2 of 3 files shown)


wconstab:
@robieta do we have a reviewer in mind for this, is it you or we need to tag someone else?

libkineto/include/ActivityType.h
```diff
@@ -22,6 +22,7 @@ enum class ActivityType {
     CPU_INSTANT_EVENT, // host side point-like events
     PYTHON_FUNCTION,
     OVERHEAD, // CUPTI induced overhead events sampled from its overhead API.
+    HPU_OP,
```
briancoutinho, line 25: please add a comment

briancoutinho: Oh also do you intend for this to be always on by default, if not you can move it below the GLOW_RUNTIME, this will make it an optional activity type

libkineto/src/output_json.cpp
```diff
@@ -186,6 +186,9 @@ void ChromeTraceLogger::handleTraceSpan(const TraceSpan& span) {
   if (!traceOf_) {
     return;
   }
+  if (span.startTime == 0 && span.endTime == 0 && span.iteration == -1) {
```
briancoutinho, line 189: confused why this is here? is related to some other change.

briancoutinho (changes requested):

libkineto/include/ActivityType.h
astachowiczhabana: Hi @briancoutinho. Can you please explain what it means for an activity to be on by default? I expect to always see HPU_OP activity type when user selects torch.profiler.ProfilerActivity.HPU activity on HPU-enabled environment. But without this it shouldn't be produced. Does this mean the activity is on by default?

libkineto/src/output_json.cpp
astachowiczhabana: it's my mistake, removing it

libkineto/include/ActivityType.h
briancoutinho: sorry about the delay:( oh the default activities are in case the trace is captured asynchronously using a configuration file (on-demand profiling mode). The torch.profiler version should not be dependent on this

briancoutinho (approved):
looks good!

wconstab:
@briancoutinho are you planning to merge this soon? I am assuming pytorchbot is not available on kineto repo?

libkineto/include/ActivityType.h
```diff
@@ -26,6 +26,7 @@ enum class ActivityType {
     // Optional Activity types
     GLOW_RUNTIME, // host side glow runtime events
     CUDA_PROFILER_RANGE, // CUPTI Profiler range for performance metrics
+    HPU_OP,  //HPU host side runtime event
```
aaronenyeshi, line 29: nit: space after comment
`HPU_OP,  // HPU host side runtime event`

aaronenyeshi (commented):
This looks good to me, we can merge it. This may require additional logic internally to handle the new type. 

Also, if you plan to use TensorBoard to view your traces and summary, you may want to also add this ActivityType, HPU_OP to tb_plugin here: https://github.com/pytorch/kineto/blob/main/tb_plugin/torch_tb_profiler/profiler/trace.py#L47

aaronenyeshi (approved):

astachowiczhabana:
@aaronenyeshi thx, is there anything I can do with failing CI at "Facebook Internal - Builds & Tests" and "Facebook Internal - Linter "?

aaronenyeshi:
I've fixed the internal failing CI issues, and landed this patch.

PR #612, opened 2022-06-03
Fix duplicate symbols when using ROC-tracer

astachowiczhabana:
Summary: Kineto source list contains duplicate entries when
ROC-tracer is enabled. This leads to multiple definition errors
when kineto library is linked with --whole-archive option.

This change is required for:
https://github.com/pytorch/pytorch/pull/77746


aaronenyeshi (approved):
LGTM!

Issue #604, opened 2022-05-25
`tb_plugin/examples/datapipe_example.py` outdated


msaroufim:
There's a few lines in this file that are now outdated

`import torchvision.prototype.models as models` does not exist

`    trainset = Cifar10().to_datapipe(root='./data', decoder=pdatasets.decoder.raw)` where `Cifar10()` expects to be passed in a `root` argument directly

Could be a couple more too

aaronenyeshi:
Thank you for catching this, please feel free to submit a patch and we can help merge, or we will get back to you.

msaroufim:
I chatted with @NicolasHug about this offline, given that this example leverages `prototype` features there is no expectation of backwards compatibility so this example will continue breaking until datapipes are promoted out of prototype. So for now I guess we can add a notice at the top of the code. I'll write something now

aaronenyeshi:
Sounds good to me!

PR #614, opened 2022-06-03
Deleted broken datapipe example

msaroufim:
This example uses experimental `torchvision` features which have no expectation of being supported yet. We can keep this example since it still has some educational value but it will break in many more ways soon with no clear path to make it work.

Related to #604


NicolasHug (approved):
Thanks Mark, LGTM.
I'll approve but IMHO it would make sense to remove completely this example. It will never work as-is and further breakage is expected, e.g. from the use of `torch.utils.data.dataloader_experimental` which will be removed soon.
So it might add unnecessary confusion to reader

msaroufim:
It would be possible to replace this with a toy example with a toy dataset, unfortunately all the domain examples are still experimental and will be for the foreseeable future. Data Loader 2 is also experimental so this example just came a bit too early. I'm on PTO for half of next week but I can produce a working toy example in the week that follows with Data Loader 1 that we'll have to update when Data Loader 2 comes out. Let me know what you prefer.

aaronenyeshi:
Let's add the new Data Loader 2 example when it comes out. Thanks!

aaronenyeshi (approved):

PR #619, opened 2022-06-14
fix JSON output

shintaro-iwasaki:
Summary:
This patch fixes JSON format by quoting hex pointer values.

For example,
```
# Wrong since a hex value (for example 0x7f162bc09600) is not a number in JSON
  {
    "ph": "X", "cat": "cuda_runtime", "name": "hipMemcpyWithStream", "pid": 1468712, "tid": 1468712,
    "ts": 1655165458323941, "dur": 32,
    "args": {
      "External id": 30,
      "src": 0x7f162bc09600, "dst": 0x7f1c47af8c40, "size": 8, "kind": 2, "stream": 0x0
    }
  },
```
->
```
# Correct since hex values are quoted.
  {
    "ph": "X", "cat": "cuda_runtime", "name": "hipMemcpyWithStream", "pid": 1468712, "tid": 1468712,
    "ts": 1655165458323941, "dur": 32,
    "args": {
      "External id": 30,
      "src": "0x7f162bc09600", "dst": "0x7f1c47af8c40", "size": 8, "kind": 2, "stream": "0x0"
    }
  },
```
I could not find "ptr" in the log, but I believe "ptr" takes a hex value.

Differential Revision: D37128234

### DIFF
--- a/libkineto/src/RoctracerActivityApi.cpp
+++ b/libkineto/src/RoctracerActivityApi.cpp
@@ -105,7 +105,7 @@ int RoctracerActivityApi::processActivities(
     a.flow.type = kLinkAsyncCpuGpu;
     a.flow.start = true;
 
-    a.addMetadata("ptr", item.ptr);
+    a.addMetadataQuoted("ptr", fmt::format("{}", item.ptr));
     if (item.cid == HIP_API_ID_hipMalloc) {
       a.addMetadata("size", item.size);
     }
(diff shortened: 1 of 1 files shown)


shintaro-iwasaki:
@mwootton I would appreciate it if you could review it.

mwootton:
@shintaro-iwasaki I seem to remember sorting through this problem before; I'm pretty sure I added that second addMetadata method to deal with this.  Clearly it is not okay now.

Yes, the pointer values need to be using AddMetadataQuoted.  The explicit fmt call is fine.

Looks good, thanks.

shintaro-iwasaki:
Thanks for your review, @mwootton! I will merge it.

aaronenyeshi (approved):

PR #621, opened 2022-06-20
Set flow end for kernel ops

mwootton:
Explicitly set the flow end on kernel ops from roctracer.  This was taking the default value.


libkineto/src/RoctracerActivityApi.cpp
```diff
@@ -238,6 +238,7 @@ int RoctracerActivityApi::processActivities(
         a.activityName = std::string(name);
         a.flow.id = record->correlation_id;
         a.flow.type = kLinkAsyncCpuGpu;
+        a.flow.start = false;
```
aaronenyeshi, line 241: LGTM, the default value should have been 0, but it's not being newly constructed every time. In this case, while iterating through the records, after the first `a.flow.start = true`, all subsequent records became true too.

aaronenyeshi (approved):
Thanks for this patch!

PR #628, opened 2022-07-05
Add python stack tracing option on on-demand flow

slgong-fb:
Summary:
Changes:
1. add an option in Config; can use 'PYTHON_STACK_TRACE=true' option (via .conf)
2. deliver PYTHON_STACK_TRACE value to kineto_client_interface start()
3. abstract class also changed.

Differential Revision: D37410204

### DIFF
--- a/libkineto/include/ClientInterface.h
+++ b/libkineto/include/ClientInterface.h
@@ -12,7 +12,7 @@ class ClientInterface {
   virtual ~ClientInterface() {}
   virtual void init() = 0;
   virtual void warmup(bool setupOpInputsCollection) = 0;
-  virtual void start() = 0;
+  virtual void start(bool withStack) = 0;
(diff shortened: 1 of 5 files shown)


slgong-fb:
Has the same issue discussed in https://github.com/pytorch/kineto/pull/631

robieta:
@slgong-fb This is a BC breaking change and is preventing us from updating the pinned version of Kineto in PyTorch. Can we revert?

robieta:
Specifically:
```
/var/lib/jenkins/workspace/torch/csrc/profiler/kineto_client_interface.cpp:21:8:error: ‘void torch::profiler::impl::{anonymous}::LibKinetoClient::start()’ marked ‘override’, but does not override
   void start() override {
^~~~~
```

https://github.com/pytorch/pytorch/pull/80974

PR #631, opened 2022-07-19
Fix CuptiActivity name from tpp to cpp

aaronenyeshi:
Summary: There was a typo in the naming of this file when it was created causing a suffix .tpp. Fix this to .cpp.

Reviewed By: slgong-fb

Differential Revision: D37967514

Pulled By: aaronenyeshi

PR #632, opened 2022-07-19
Disable test_profiler_api_without_gpu in TB CI

aaronenyeshi:
As discussed with Taylor in https://github.com/pytorch/kineto/pull/631, test_profiler_api_without_gpu should be disabled in TB_Plugin_CI, and if we want this to exist, it should exist in PyTorch CI.


aaronenyeshi:
Using CI tests to validate this patch.

robieta (approved):
Thanks!

PR #637, opened 2022-07-28
Add ipcfabric for receiving on-demand tracing and base config options

briancoutinho:
Summary:
## Summary
This change to enable communication to a daemon in OSS, it is a preparatory change before we open source the daemon.

## How it works?
Each kineto instance registers itself with the deamon, this should work in a multi-processing use-case as well
1. Kineto registers pid and
2. Kineto periodically polls

## IPCFabric
This provides a great abstract for inter process communication on the system. It uses Linux named sockets under the hood.
Ideally ipcfabric will be open sourced separately but for now we are adding it inside kineto in ipcfabric/ directory.

## Changes
1. Rename DaemonConfigLoader to an interface IDaemonConfigLoader
2. Add a DaemonConfigLoader implementation that leverages IpcFabricClient
3. Add the IpcFabricClient leverages IPC Fabric to do registry and polling with a daemon.

Differential Revision: D37705980


aaronenyeshi (approved):
LGTM, thanks for adding this feature!

PR #638, opened 2022-07-29
Fix undefined reference

mosout:
Fix:
```
error: _deps/kineto-build/libkineto.so: undefined reference to vtable for libkineto::ActivityTrace
error: _deps/kineto-build/libkineto.so: undefined reference to libkineto::isGpuAvailable()
```


robieta:
It looks like this was independently discovered in https://github.com/pytorch/kineto/pull/639 (plus another fix), so I'm going to close this one. Thanks though!

PR #641, opened 2022-08-04
Exit `profilerLoop` after sleep when requested

Flamefire:
I see a SEGFAULT on application shutdown as the system is shutting down but the profiler thread is sleeping and then tries to report a last sample but the loggers are already gone.
So check the flag after the sleep to mitigate this.

Actually this is an example of static deinit fiasco. But fixing this is really hard...


aaronenyeshi (approved):
LGTM!

PR #642, opened 2022-08-05
Fix crash during shutdown when profiler was used

Flamefire:
The EventProfilerController uses the loggers, so when they get destroyed before the controller you'll get a segfault.
Currently the creation / function calls are in this order:
- profilerMap
- loggers
- makeLoggers

This means the loggers will be destroyed first (reverse order of construction).
This changes puts the profilerMap call to the end destroying that first, then the loggers.


robieta (approved):
Thank you not only for running down this issue (which I'm sure was tremendously unpleasant), but also for the lovely comment explaining it.

PR #646, opened 2022-08-11
guard against negative durations in Chrome trace generation

chaekit:
Summary: the profiler may have 0 end time stamp which was recently fixed in D38426342. even though thats the source of negative duration, we should be extra cautious with negative duration in chrome trace generation as thats where the side effect is the worst: super zoomed out trace that is annoying

Reviewed By: aaronenyeshi

Differential Revision: D38607705


aaronenyeshi (approved):
LGTM!

PR #648, opened 2022-08-17
add prototype for isGpuAvailable()

chaekit:
Summary: This is a generic util function that checks if GPU is available on the system but the current impl is not compatible for mobile build CIs. due to missing prototype error.

Reviewed By: aaronenyeshi

Differential Revision: D38803286


aaronenyeshi (approved):
LGTM! Reviewed internally.

PR #651, opened 2022-08-25
Remove guards from Python stack tracing OD flow

slgong-fb:
Summary:
Diff for removing #ifdef guard from Python stack tracing option on on-demand flow.
1. Point a new Kineto submodule to compile in CIs with kineto.submodule.txt
2. The new Kineto submodule has 'withStack' inside #ifdef USE_KINETO_MIN_CHANGE
3. This diff removed #ifdef from codes
4. but we still need another diff to deprecate 'USE_KINETO_MIN_CHANGE' from TARGETS and .bzl files.

Differential Revision: D38963630


aaronenyeshi (approved):

aaronenyeshi (changes requested):
Closing this, outdated pr

Issue #647, opened 2022-08-15
[RFC] Make the kineto extendable for other runtime than CUDA


xunsongh:
## 🚀 Feature
This RFC proposes to add a profile abstract shim layer to Kineto for plug-in to extend. They align with CUPTI. The common structure of the profiling information can be shared. And processes the profiling information with same post-analysis . To enable new language profiling, minimal changes are required in Kineto packages.

## Motivation
The Pytorch uses the Kineto as the new profiler middle ware. But the CUPTI (ROCm corresponding for AMD GPU) is the only supported backend for Pytorch CUDA in Kineto. As the Pytorch supports more and more new devices other than CUDA, like XPU for Intel's GPU which is based on Intel's SYCL, it becomes a problem that the Kineto is coupled CUPTI tightly and hard to support new device type in Pytorch. The easy extendable Kineto is required for collection profiling information on different runtime.

## Pitch
An extendable Kineto should support those extensions of other languages which operates like a plug-in, such as SYCL on XPU backend. By abstracting the CuptiActivityProfiler class to offer a registrable interface for backends to override, the Kineto could support different backends other than CUDA/ROCm.  PyTorch can pass the device message (aka. DeviceType) to Kineto and let Kineto choose different path for each backend. This puts requirements for the universality of some core methods such as startTrace/stopTrace/processTrace. Seperately, for XPU path, we could override those API functions at our Intel extension side outside of Kineto or PyTorch, and for CUDA/ROCm path, all things will remain as normal. For further, this abstraction won't affect all the build. That means, with or without XPU, the build will always pass as usual. The only new feature is, without XPU build in, the profiler will offer a warning for XPU users.

To describe the above design, we offer a suspected structure in this picture.
![image](https://user-images.githubusercontent.com/74238031/184579140-9c173b93-e882-4dc2-bd16-b7d4ec22c1d7.png)

For this design, we need:
1. Abstract the middle class - ActivityProfilerBase
2. Add DeviceType pass as an argument(not show in the picture)
3. Abstract core functions within ActivityProfilerBase class
4. Bypass XPU path with existed paths within CuptiActivityProfiler class

### Open issues:
Discuss and choose a proper way offered above to support Intel extension for PyTorch

xunsongh:
@gdankel Would you please help review this RFC? I think you should be an expert on this structure because you contributed most of the code on the CuptiActivityProfiler class. Would you please provide your suggestions on this RFC?

gujinghui:
[...]

PR #662, opened 2022-09-30
To support extend Profiler

xunsongh:
With this patch, the kineto will be able to support extend profiler to register themselves and co-work with kineto's tracing strategy. In this patch, we made three important changes in kineto.

1. The middle level has been abstracted as an abstract class - ActivityProfilerBase. All profiler instantiations inside or outside the kineto will inherit this class with overriding necessary APIs. Detailed design for this structure please refer to this [RFC](https://github.com/pytorch/kineto/issues/647)

2. The DeviceType enum class has been added to include/ as same as ActivityType. This enum class will make the profiler controller be enabled to choose integrated or extend path for instantiating the profiler_. This argument will be passed via front APIs such as initProfilerIfRegistered, which called by PyTorch.

3. A simple hook has been added into include/libkineto.h, with the hook function - registerExtendProfiler. This hook enables extend plug-ins such as Intel® Extension for PyTorch* to register extend profiler into kineto and override the constructor of profiler_ in controller which inherit the abstract class - ActivityProfilerBase described above.

* Any questions with this patch please contact Xunsong, Huang (xunsong.huang@intel.com)

Signed-off-by: Xunsong, Huang <xunsong.huang@intel.com>

### DIFF
--- a/libkineto/include/ActivityProfilerBase.h
+++ b/libkineto/include/ActivityProfilerBase.h
@@ -1,0 +1,86 @@
+#pragma once
+
+#include <chrono>
+
+#include "IActivityProfiler.h"
+#include "ITraceActivity.h"
+
+namespace libkineto {
+
+class ActivityProfilerBase {
+
+ public:
+  virtual ~ActivityProfilerBase() = default;
+
+  virtual bool isActive() const {
+    fail();
+    return false;
+  }
+
+  virtual const std::chrono::time_point<std::chrono::system_clock> performRunLoopStep(
+      const std::chrono::time_point<std::chrono::system_clock>& now,
+      const std::chrono::time_point<std::chrono::system_clock>& nextWakeupTime,
+      int64_t currentIter = -1) {
+    fail();
+    return nextWakeupTime;
+  }
+
+  virtual void setLogger(ActivityLogger* logger) {
+    fail();
+  }
+
+  virtual void startTrace(
+      const std::chrono::time_point<std::chrono::system_clock>& now) {
+    fail();
+  }
+
+  virtual void stopTrace(
+      const std::chrono::time_point<std::chrono::system_clock>& now) {
+    fail();
+  }
+
+  virtual void processTrace(ActivityLogger& logger) {
+    fail();
+  }
+
+  virtual void reset() {
+    fail();
+  }
+
+  virtual void configure(
+      const Config& config,
+      const std::chrono::time_point<std::chrono::system_clock>& now) {
+    fail();
+  }
+
+  virtual void transferCpuTrace(
+      std::unique_ptr<CpuTraceBuffer> cpuTrace) {
+    fail();
+  }
+
+  virtual const Config& config() {
+    fail();
+    Config* config_;
+    return *config_;
+  }
+
+  virtual inline void recordThreadInfo() {
+    fail();
+  }
+
+  virtual void addMetadata(const std::string& key, const std::string& value) {
+    fail();
+  }
+
+  virtual void addChildActivityProfiler(
+      std::unique_ptr<IActivityProfiler> profiler) {
+    fail();
+  }
+
+ private:
+  void fail() const {
+    throw std::runtime_error("KINETO: ActivityProfilerBase class not instantiated.");
+  }
+};
+
+}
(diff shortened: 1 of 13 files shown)


gujinghui:
@aaronenyeshi  @chaekit @gdankel @robieta @briancoutinho

Could you help review and give suggestions? Thanks.

aaronenyeshi:
This looks good, can you please help rebase on main, and I will run some internal checks?

briancoutinho:
@xunsongh  there already is an activity profiler and activity profiler session base class. Can we please use that instead of introducing another way to add plugins??? 
https://github.com/pytorch/kineto/blob/main/libkineto/include/IActivityProfiler.h#L73-L117

This will likely break any other use-cases of the other activity profiler plugin extension interface.

xunsongh:
> [xunsongh](https://github.com/xunsongh) You are right actually, it won't break any existing features 👍 Although, I think this leads to two ways to extend Kineto activity profiler - one using IActivityProfiler that is a plugin based , and another using ActivityProfilerBase.
> 
> Currently the codes somewhere between a refactor, @gdankel was adding a change to split up ActivityProfiler into two - one core composite profiler with all management logic and a set of plugins - like CuptActivityProfiler, CuptiRangeProfiler. (Cupti Range profiler is a plugin example - https://github.com/pytorch/kineto/blob/main/libkineto/src/CuptiRangeProfiler.h)
> 
> We can use the current approach and eventually merge ActivityProfilerBase with IactivityProfilerSession, would that be okay?

That's a good news to know that someone is working on decoupling the ActivityProfiler. And I think the purposes of us are similar to each other. There is only one suggestion or requirement from me that we wish more abstract API in high level abstract class (IActivitySession) besides `start`, `stop` and `processTrace`. For example, we are looking forward to changing runtime configures for each tracer. And also, we suppose the kineto could handle the out-of-queue scheme. We are still discussing these requirements inside Intel and will offer a complete requirements list to you.

And then the most necessary thing is, I need to be noticed when your changes of spliting up ActivityProfiler will done and which branch to trace those changes. I wish to exchange any information on this topic with you in time.

gdankel:
> That's a good news to know that someone is working on decoupling the ActivityProfiler. And I think the purposes of us are similar to each other. There is only one suggestion or requirement from me that we wish more abstract API in high level abstract class (IActivitySession) besides `start`, `stop` and `processTrace`. For example, we are looking forward to changing runtime configures for each tracer. And also, we suppose the kineto could handle the out-of-queue scheme. We are still discussing these requirements inside our team and will offer a complete requirements list to you.

While Kineto supports arbitrary configuration, it is not currently exposed to PyTorch. What sort of configuration did you have in mind?

What do you mean by out-of-queue scheme?

> And then the most necessary thing is, I need to be noticed when your changes of spliting up ActivityProfiler will done and which branch to trace those changes. I wish to exchange any information on this topic with you in time.

#698 lays the groundwork for the splitting up of ActivityProfiler, but the actual split is for a later patch. And there are three things to split out: CUPTI specific, ROCm specific and PyTorch specific. The refactoring mostly focuses on splitting out the PyTorch specific code. But that should still help with adding support for new devices.

aaronenyeshi (changes requested):
not sure where to go from here, refactoring #698 is too risky as it may cause breakage internally. Maybe we should support XPU the same way we support AMD? cc @gdankel @briancoutinho

xunsongh:
> not sure where to go from here, refactoring #698 is too risky as it may cause breakage internally. Maybe we should support XPU the same way we support AMD? cc @gdankel @briancoutinho

Hi, @aaronenyeshi 

I apologize for not having updated or kept track of this PR until now. The reason I have not closed this PR is that there is still a requirement on our end to enable Kineto support for the XPU backend. However, based on in-depth discussions with the Kineto developers and drawing from my personal development experience over the past months, I am planning to approach the integration of XPU backend support in a different manner than what was proposed in the original PR.

My colleagues are currently developing a standalone PTI-sdk library, and we hope that in the future, Kineto could integrate our library in a similar fashion to how it integrates CUPTI or Roctracer. This would ideally allow for full support of the XPU backend with minimal modifications. In approximately one to two months, I intend to submit a new RFC and accompanying PR with revised code for discussion.

For now, I will close this PR to prevent any confusion.

Thank you for your understanding.

PR #687, opened 2022-11-17
Kineto even logging: make log file write atomic

dzhulgakov:
It's useful for letting an external script know when logging is done.

Also recognize `.pt.trace.json` extension when appending the pid - it's a convention used in some visualizations (we probably should make {pid} a template instead)

### DIFF
--- a/libkineto/src/Config.cpp
+++ b/libkineto/src/Config.cpp
@@ -344,9 +344,14 @@ bool Config::handleOption(const std::string& name, std::string& val) {
   } else if (!name.compare(kActivitiesLogFileKey)) {
     activitiesLogFile_ = val;
     activitiesLogUrl_ = fmt::format("file://{}", val);
-    size_t jidx = activitiesLogUrl_.find(".json");
+    size_t jidx = activitiesLogUrl_.find(".pt.trace.json");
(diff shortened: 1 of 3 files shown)


aaronenyeshi (approved):
LGTM, please rebase on main, and I can help import/land.

aaronenyeshi:
Sorry, this patch is causing a few internal test issues, related to the file logic. We may need to revert this patch. We can reland this patch as separated patches, with a fix for the following issue: 
```
ERROR:2022-12-05 09:25:46 76568:76568 output_json.cpp:438] Failed to rename /tmp/libkineto_activities_tmp.json.drAKuH to /mnt/remote_execution/execution/1/f37b6720f9704e3bbfa7baf1c9e778a7/tmp/tmp_r3do2sk.json : Invalid cross-device link
```

dzhulgakov:
Ah, good catch regarding the temp file - moving cross file-systems with `rename` doesn't work. Also my usage of `mkstemp` seem to have broken Windows build: https://github.com/pytorch/pytorch/pull/89846 (see windows build failures).

Let me send a PR that just makes files have `.tmp` suffix before they are finished. How about that?

PR #689, opened 2022-11-18
Fixed test_profiler_tree.py unit testing

BLOrange-AMD:
Fixed test_profiler_tree.py TestProfilerTree.test_profiler_experimental_tree

### DIFF
--- a/libkineto/src/RoctracerLogger.cpp
+++ b/libkineto/src/RoctracerLogger.cpp
@@ -272,6 +272,7 @@ void RoctracerLogger::startLogging() {
     loggedIds_.add("hipPeekAtLastError");
     loggedIds_.add("hipModuleGetFunction");
     loggedIds_.add("hipEventCreateWithFlags");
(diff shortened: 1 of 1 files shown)


jithunnair-amd (approved):
Excluding this API to match Kineto profiling behaviour on CUDA, since the unit test mentioned looks for an exact match of profiling output. Ideally, the pytorch unit test would not be so fragile, and should exclude the cuda API calls from the profiling output (not at the Kineto level) if they are not relevant to the unit test.
https://github.com/pytorch/pytorch/blob/ba605c3b0439fd5dfe062f42e60b990c88c061d4/test/profiler/test_profiler_tree.py#L251

jithunnair-amd:
cc @aaronenyeshi Please review

aaronenyeshi (approved):
LGTM, please rebase on main. And also feel free to create PRs for changes to unit tests in pytorch/pytorch, there you can add tag oncall: profiler.

PR #691, opened 2022-11-18
Make on demand update duration configurable

dzhulgakov:
What title says. We read it from the main config and apply it to the next iteration of ondemand config loading. It also refactors the background thread loop to do time accounting properly if the update intervals are not divisible

### DIFF
--- a/libkineto/src/Config.cpp
+++ b/libkineto/src/Config.cpp
@@ -42,6 +42,7 @@ constexpr int kDefaultSamplesPerReport(1);
 constexpr int kDefaultMaxEventProfilersPerGpu(1);
 constexpr int kDefaultEventProfilerHearbeatMonitorPeriod(0);
 constexpr seconds kMaxRequestAge(10);
+constexpr seconds kDefaultOnDemandConfigUpdateIntervalSecs(5);
 // 3200000 is the default value set by CUPTI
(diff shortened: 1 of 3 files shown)


libkineto/src/Config.cpp
```diff
@@ -42,6 +42,7 @@ constexpr int kDefaultSamplesPerReport(1);
 constexpr int kDefaultMaxEventProfilersPerGpu(1);
 constexpr int kDefaultEventProfilerHearbeatMonitorPeriod(0);
 constexpr seconds kMaxRequestAge(10);
+constexpr seconds kDefaultOnDemandConfigUpdateIntervalSecs(5);
```
aaronenyeshi, line 45: Why is this a different default value than the constexpr in configLoader.cpp:39 below? (ie. 5 vs 1)

aaronenyeshi (approved):
LGTM, may want to add a comment in the update thread loop. I am good with the ability to change config update internal on-demand.

libkineto/src/Config.cpp
dzhulgakov: The one below is not even used, let me clean it up

dzhulgakov:
Add some comments too

robieta:
Looks like the merge conflict here is real.

dzhulgakov:
Rebased!

aaronenyeshi (approved):

PR #692, opened 2022-11-18
Add option to use domain sockets instead of abstract sockets

dzhulgakov:
Domain sockets are part of file system and, thus, are easier to share in containerized environments. This adds `KINETO_IPC_SOCKET_DIR` env var to control the behavior

### DIFF
--- a/libkineto/ipcfabric/Endpoint.h
+++ b/libkineto/ipcfabric/Endpoint.h
@@ -9,6 +9,7 @@
 #include <asm/unistd.h>
 #include <stdlib.h>
 #include <sys/socket.h>
+#include <sys/stat.h>
 #include <sys/un.h>
 #include <unistd.h>
 #include <stdexcept>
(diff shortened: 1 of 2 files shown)


aaronenyeshi (commented):
Overall looks good to me, but will leave approval to @briancoutinho , since this code comes from his team.

briancoutinho:
@dzhulgakov thanks for adding this, i'll merge this and also make sure dynolog ipcfabric is synced with this

Also (@aaronenyeshi ) this issue tracks making dynolog a submodule of kineto so we do not have two copies of ipcfabric
https://github.com/facebookincubator/dynolog/issues/57

aaronenyeshi (approved):
LGTM, approving for landing

libkineto/ipcfabric/Endpoint.h
```diff
@@ -163,8 +174,21 @@ class EndPoint final {
...
   }
 
-  const char* getName(TCtxt const& ctxt) const noexcept {
-    return ctxt.msg_name.sun_path + 1;
+  const char* getName(TCtxt const& ctxt) const {
+    const char* socket_dir = getenv("KINETO_IPC_SOCKET_DIR");
+    bool is_domain_socket = socket_dir && socket_dir[0];
+    if (is_domain_socket) {
+      int prefix_len = strlen(socket_dir) + 1;
+      if (strncmp(socket_dir, ctxt.msg_name.sun_path, prefix_len) == 0) {
```
briancoutinho, line 182: @dzhulgakov is this meant to be '!=' 0 instead of == 0

dzhulgakov: It should be, my bad! It didn't trigger because I put `prefix_len` to be +1. C strings are hard :)

briancoutinho:
Thanks for the update @dzhulgakov 
Syncing this up in dynolog in  https://github.com/facebookincubator/dynolog/pull/62

PR #693, opened 2022-11-18
FabricManager.h fixes to sync with dynolog repo

dzhulgakov:
`ipcfabric` is replicated across two repos (see https://github.com/facebookincubator/dynolog/issues/57). Let's handle C++17's fixes better (until the TODO is done) to make the two places identical

### DIFF
--- a/libkineto/ipcfabric/FabricManager.h
+++ b/libkineto/ipcfabric/FabricManager.h
@@ -17,6 +17,8 @@
 #include "Endpoint.h"
 #include "Utils.h"
 
+// If building inside Kineto, use its logger, otherwise use glog
+#ifdef KINETO_NAMESPACE
 // We need to include the Logger header here for LOG() macros.
 // However this can alias with other files that include this and 
 // also use glog. TODO(T131440833)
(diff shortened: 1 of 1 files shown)


libkineto/ipcfabric/FabricManager.h
```diff
@@ -91,7 +97,7 @@ struct Message {
   std::string src;
 };
 
-#ifdef ENABLE_IPC_FABRIC
+#if !defined(KINETO_NAMESPACE) || defined(ENABLE_IPC_FABRIC)
```
aaronenyeshi, line 100: Why is this not defined for Kineto Namespace?

aaronenyeshi (commented):
Added a question, overall looks good

libkineto/ipcfabric/FabricManager.h
dzhulgakov: I just wanted to make files in dynolog and kineto the same for convenience. So !defined(KINETO_NAMESPACE) is just a check for "compiling inside dynolog". Alternatively I can define ENABLE_IPC_FABRIC in dynolog's build if you prefer that

aaronenyeshi (approved):

PR #696, opened 2022-11-26
Fix Config's destruction order issue

dzhulgakov:
The previous trick with `weak_ptr` didn't work on its own and led to crashes like in https://github.com/pytorch/pytorch/pull/89174 . This seems to be sufficient to guarantee lifetime even though it still accesses destroyed `weak_ptr` instance which is technically UD. A more proper fix would probably be to pass FactoryMap as Config constructor, but it's a more involved refactoring.

### DIFF
--- a/libkineto/include/Config.h
+++ b/libkineto/include/Config.h
@@ -333,6 +333,12 @@ class Config : public AbstractConfig {
 
   void print(std::ostream& s) const;
 
+  // Config relies on some state with global static lifetime. If other
+  // threads are using the config, it's possible that the global state
+  // is destroyed before the threads stop. By hanging onto this handle,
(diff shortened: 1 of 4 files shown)


robieta (approved):
LGTM

aaronenyeshi (approved):
Thank you very much for this fix! Looks like it was not common for the config update thread to wake and reference a destroyed config factory until nvfuser unit tests. Let's land it! :)

aaronenyeshi:
Since this is merged, is there a submodule update in pytorch/pytorch, you'd like us to accept?

dzhulgakov:
Looks like @robieta is on it with https://github.com/pytorch/pytorch/pull/89846 (and automated submodule updates are back)

PR #698, opened 2022-12-01
Refactor for extensibility 6/n (#697)

aaronenyeshi:
Summary:
Pull Request resolved: https://github.com/pytorch/kineto/pull/697

Implements API described in https://fb.quip.com/aZmvA9sQqMqp.
The next diff will merge the LibkinetoObserver with the profiler_kineto and properly separate it from the CuptiActivityProfiler.

Differential Revision: D41641206

Pulled By: aaronenyeshi


aaronenyeshi:
cc - @gdankel , @briancoutinho

gdankel:
Thanks for posting @aaronenyeshi.

Since the linked document is Meta-internal I will add some quick context and notes here.

This is a big change, and is WIP. If we want to land this then probably it will be better to split it up into smaller PRs.
But It might be best to review the whole thing at a high level.

Goals of the refactoring diffs as a whole:
* Facilitate adding and controlling additional profilers and exposing them to PyTorch profiler by means of additional activity types
* Split up and simplify existing profilers to focus on a single task - in particular separate the PyTorch ops profiler from the CUPTI profiler and de-duplicate Kineto and PyTorch code
* Enable control and integration of external profilers, e.g. eBPF etc
* Make it easier to integrate and connect data from multiple profilers

Secondary goals:
* Refactor and simplify control logic, including synchronous vs. asynchronous API
* Remove existing ActivityProfilerProxy

This is not the last patch in the series of patches needed to reach the above goals, so consider it Work In Progress.

High level design:
* Multiple profilers are supported via Composite and Factory design patterns.
* Synchronized start and stop by including separate initialization step.
* PyTorch profiler interface remains the same: profilers advertise a list of supported events and configuration options that users can specify. The composite profiler will instantiate and configure the appropriate profilers given a list of events to profile.
* The intent is that this design also supports profiler adapters / controllers - e.g. an adapter can expose functionality of an existing system-level profiler to PyTorch, making it possible to configure and control operation of external profilers.

Profiler API: libkineto/include/IActivityProfiler.h
Composite profiler: libkineto/include/ICompositeProfiler.h

PR #702, opened 2022-12-07
Filter roctracer events to active timespan

mwootton:
When kineto is in "warmup" mode, both cupti and roctracer are active, logging events.  Output events should be limited to the "active" range (and not include the "warmup" events).

CuptiActivityProfiler.cpp provides this filtering for cupti.  Added equivalent filtering to RoctracerActivityApi.cpp.

Emulating the non-optimal behavior of the cupti path where these warmup events are fully logged and stored.  Events could be suppressed at collection-time rather than output-time.

### DIFF
--- a/libkineto/src/RoctracerActivityApi.cpp
+++ b/libkineto/src/RoctracerActivityApi.cpp
@@ -80,6 +80,8 @@ int RoctracerActivityApi::processActivities(
   clock_gettime(CLOCK_REALTIME, &t00);
 
   const timestamp_t toffset = (timespec_to_ns(t0) >> 1) + (timespec_to_ns(t00) >> 1) - timespec_to_ns(t1);
+  // Our stored timestamps (from roctracer and generated) are in CLOCK_MONOTONIC domain (in ns).
+  // Convert the guards rather than each timestamp.
   startTime = (startTime * 1000) - toffset;
   endTime = (endTime * 1000) - toffset;
 
--- a/libkineto/src/RoctracerActivityApi.cpp
+++ b/libkineto/src/RoctracerActivityApi.cpp
@@ -64,8 +64,13 @@ void RoctracerActivityApi::setMaxBufferSize(int size) {
   //maxGpuBufferCount_ = 1 + size / kBufSize;
 }
 
+inline bool inRange(int64_t start, int64_t end, int64_t stamp) {
+    return ((stamp > start) and (stamp < end));
+}
+
 int RoctracerActivityApi::processActivities(
-    ActivityLogger& logger, std::function<const ITraceActivity*(int32_t)> linkedActivity) {
+    ActivityLogger& logger, std::function<const ITraceActivity*(int32_t)> linkedActivity,
+    int64_t startTime, int64_t endTime) {
   // Find offset to map from monotonic clock to system clock.
   // This will break time-ordering of events but is status quo.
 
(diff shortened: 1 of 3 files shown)


libkineto/src/RoctracerActivityApi.cpp
```diff
@@ -75,6 +80,8 @@ int RoctracerActivityApi::processActivities(
   clock_gettime(CLOCK_REALTIME, &t00);
 
   const timestamp_t toffset = (timespec_to_ns(t0) >> 1) + (timespec_to_ns(t00) >> 1) - timespec_to_ns(t1);
+  startTime = (startTime * 1000) - toffset;
+  endTime = (startTime * 1000) - toffset;
```
aaronenyeshi, line 84: should this be endTime? 
```
endTime = (startTime * 1000) - toffset;
```

libkineto/src/RoctracerActivityApi.cpp
```diff
@@ -75,6 +80,8 @@ int RoctracerActivityApi::processActivities(
   clock_gettime(CLOCK_REALTIME, &t00);
 
   const timestamp_t toffset = (timespec_to_ns(t0) >> 1) + (timespec_to_ns(t00) >> 1) - timespec_to_ns(t1);
+  startTime = (startTime * 1000) - toffset;
```
aaronenyeshi, line 83: Could you add a small comment about the time conversion? Also, we should probably move the timestamps in Kineto to TSC clock in the future.

aaronenyeshi (changes requested):
Looks good, please check few small comments

libkineto/src/RoctracerActivityApi.cpp
mwootton: Good eye, thanks. Did a copy/paste back out of kineto@pytorch_1.13 branch I was testing in.  Update momentarily.

libkineto/src/RoctracerActivityApi.cpp
mwootton: I will.  Generally you would use CLOCK_MONOTONIC.  It always moves forward so it does not invert the order of any two events.  Used by e.g.: roctracer, all my tools, and viztracer.

Cupti is pushing out CLOCK_REALTIME timestamps.  So you will see events invert whenever an NTP update happens.  So you will see a kernel execute before it was queued. :o

By TSC I think you effectively mean CLOCK_MONOTONIC_RAW.  That will drift over time (months of uptime) as it makes no adjustment.  This will become a problem when doing multi-node profiling.  You will have to implement a clock synchronization protocol that runs at collection time.  CLOCK_MONOTONIC on NTP systems may be close enough for government work.

And while we are on the topic: pytorch and kineto are using usec timestamps which are course enough to cause collisions.  This leads to things like: 1) events having zero duration and 2) calling order (stack frames) being ambiguous.

Oh last bit: add a libkineto::getTimestamp() function.  In that above code I am using CLOCK_REALTIME empirically to convert the timestamps.  I should really be asking kineto what the current timestamp is and adjusting to that.  Current code will be broken if you convert away from the current timestamp domain.

mwootton:
cc @jeffdaily @sunway513 @jithunnair-amd

jeffdaily (approved):
In Michael we trust.  LGTM.

aaronenyeshi (approved):
LGTM!

PR #704, opened 2022-12-08
Fix atomic file write on Windows (again)

dzhulgakov:
Sent https://github.com/pytorch/pytorch/pull/90446 . It confirms that Windows tests pass!


briancoutinho (approved):
thanks for adding this fix @dzhulgakov 👍

PR #707, opened 2022-12-13
[kineto] Add dynolog as third party dep to obtain ipcfabric code

briancoutinho:
Add dynolog as a submodule

## Test
```
mkdir build; cd build;
cmake -G=Ninja ..
cmake --build .
```

Build for static library works:
```
[23/23] Linking CXX static library libkineto.
```

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -4,3 +4,6 @@
 [submodule "libkineto/third_party/fmt"]
 	path = libkineto/third_party/fmt
 	url = https://github.com/fmtlib/fmt.git
+[submodule "libkineto/third_party/dynolog"]
+	path = libkineto/third_party/dynolog
+	url = https://github.com/facebookincubator/dynolog.git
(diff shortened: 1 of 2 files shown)


briancoutinho:
@dzhulgakov  thanks for adding the C++17/14 defs, pulled those changes into dynolog. This PR should remove the duplication of ipcfabric.
I'll add a cleanup PR to delete the files later.

aaronenyeshi (approved):

.gitmodules
```diff
@@ -4,3 +4,6 @@
 [submodule "libkineto/third_party/fmt"]
 	path = libkineto/third_party/fmt
 	url = https://github.com/fmtlib/fmt.git
+[submodule "libkineto/third_party/dynolog"]
```
dzhulgakov, line 7: maybe minor - but why include dynolog in kineto instead of the other way around? There would be less to checkout when building pytorch.

briancoutinho: yea that is a fair point. Not a great answer but we are maintaining ipcfabric inside dynolog project. There are some in-house use-cases for it outside of Kineto too. 
-  In terms of impact  on pytorch at least we are not building dyno files at all, we just need to use ipcfabric directory that is a header only library :)
- it sill will checkout the dynolog files however.
- i'll look into some way to do a shallow checkout on submodules.

PR #714, opened 2023-01-19
Env variable to set log level

dzhulgakov:
Small improvement for debugging. Env var is better than the config since it's applied during initialization too. Also it makes suppressLibKinetoLogMessage a noop if env var was provided. Pytorch always calls it thus kind of defeating the purpose for configuring the log level


aaronenyeshi (approved):
LGTM! Thanks for the patch.

PR #715, opened 2023-01-19
Fix uuid generation for IPC fabric

dzhulgakov:
PyTorch calls IPCFabric initialization from static initializer. So it might be called before global variables in this translation unit are initialized. That was happening before (at least in my build) - all uuids generated had all zeroes in them since the random generator was in a bad state.


aaronenyeshi (approved):
LGTM, thanks for the patch!

PR #716, opened 2023-01-19
Snappier start time (10s->2s) when initial timestamp is not specified

dzhulgakov:
I think it's sufficient and gives slightly better UX


aaronenyeshi:
Will test this internally in CI. Have you tried this change on a larger distributed profile? Such as over 128 ranks or 512 ranks?

gdankel:
I usually used ~2s myself and I believe it's fine in the majority of cases, but it was set to 10s to handle larger jobs as @aaronenyeshi alluded to...

PR #718, opened 2023-01-20
enable ipc fabric using env variable

briancoutinho:
Summary: This will remove the need to explicitly enable IPC fabric in a config.

Differential Revision: D42646533


aaronenyeshi (approved):
LGTM! Wondering if the getenv function need to be guarded for __linux__ only?  Maybe a shared function that checks the guard to linux, and defaults to false ?

PR #725, opened 2023-02-12
Update static assert to cpp14 compat syntax and upgrade fmtlib

Skylion007:
* This make fmtlib match the version for PyTorch version
* Fixes a static assert so it compiles just fine in C++14 as well.


Skylion007:
@briancoutinho Mind taking a quick look?

aaronenyeshi (approved):
Thanks for the patch, LGTM!

PR #746, opened 2023-03-24
make input shape collection opt-in for on-demand tracing

chaekit:
Summary: Make input shape collection opt-in as we are re-rolling out the feature for on-demand tracing. Making this default right away could cause bloated trace size for inferences

Differential Revision: D44377341


aaronenyeshi (approved):
LGTM, but this confit may have changed the bool name.

PR #755, opened 2023-04-28
Update config to send all options and turn ON record_shapes by default

aaronenyeshi:
Reviewed By: leitian

Differential Revision: D45403874

Pulled By: aaronenyeshi


aaronenyeshi:
Landed in another diff.

PR #772, opened 2023-06-19
Roctracer: Enable individual activity types

mwootton:
RoctracerActivityApi was logging api, op, and correlation data whenever enabled, i.e. all-or-nothing.  This caused failures in some unit tests which wanted to see that kernels were enqueued (CUDA_RUNTIME) but insisted it didn't see them run (CONCURRENT_KERNEL).

Now we are correctly identifying GPU_MEMCPY, GPU_MEMSET, and CONCURRENT_KERNEL activity.  This is done via replicating an enum that was previously internal but stable.  It will be public in future rocm releases.  The names will not collide.   

Also, Roctracer reports barrier/marker ops that run on the gpu while CUPTI does not.  We now suppress these as they are pretty noisy with very little value.

Addresses https://github.com/pytorch/pytorch/issues/97167


mwootton:
@sunway513 @jithunnair-amd @aaronenyeshi

libkineto/src/RoctracerActivityApi.cpp
```diff
@@ -292,7 +321,35 @@ int RoctracerActivityApi::processActivities(
...
+          default:
+            if (!isLogged(ActivityType::CONCURRENT_KERNEL))
+              filtered = true;
+            if (record->op == HIP_OP_ID_BARRIER)  // Don't record barriers/markers
+              filtered = true;
+            a.activityType = ActivityType::CONCURRENT_KERNEL;
+            break;
+        }
+
+        if (!filtered && inRange(startTime, endTime, record->begin_ns)) {
```
davidberard98, line 352: if filtered == true then do we just ignore this activity?

davidberard98 (commented):
generally lgtm, left a comment

also, please take a look at the failing test - although it doesn't look related, I can't find it failing in other places.

I can't comment on the hip/rocm-specific stuff, so we can wait for an AMD reviewer if you'd like

jithunnair-amd:
cc @jeffdaily for review

jeffdaily (approved):

jeffdaily:
Approved, pending fixing the failing unit tests.

aaronenyeshi (approved):
LGTM, let's rebase and land this.

jeffdaily:
Now that this is merged, how do we get pytorch to update the kineto submodule?

davidberard98:
in pytorch, `cd third_party/kineto; git checkout KINETO_COMMIT; cd ../../; git add third_party/kineto; git commit ...` - and open a PR for this.

Feel free to open a PR, otherwise LMK and I could also do it.

jeffdaily:
@davidberard98 sorry I didn't explain myself well enough.  I know how to update submodules, but there seems to be an automated process already in place?  See https://github.com/pytorch/pytorch/pull/96240.  Do we just wait or do we submit a PR outside of this automated cycle?

jeffdaily:
https://github.com/pytorch/pytorch/pull/105866

Issue #760, opened 2023-05-18
【Feature Discussion】Make tb_plugin support larger traces


lh-ycx:
Hi guys, I'm recently working on profiling LLM (like GPT-3) workloads using the Pytorch profiler. When I tried to visualize the trace using tensorboard, one major problem was that the trace is too large (typically > 1GB for one step), causing no response in the browser.

I would like to work on this problem to make the tb_plugin support larger traces. Below are a few of my concerns:

- I cannot find a doc for tb_plugin (about the code structure and the plugin's arch), which makes it difficult for me to get started.
- Could you provide some **insight** into this problem? like where is the possible bottleneck (the original tensorboard? the trace analysis server? the frontend browser? or something else?)
- I also noticed that "monitoring daemon for larger scale deployments" is in progress, is this solving the same problem? If yes, can I get involved?

Thanks : )

aaronenyeshi:
Hi @lh-ycx , please feel free to contribute to the tb_plugin directory (the code is outdated, and under maintenance mode only).

- As far as I know, the documentation can be found here: https://github.com/pytorch/kineto/blob/main/tb_plugin/README.md . 
- This is also a problem for Chrome traces, when using the `export_chrome_trace` API. One workaround is to move the `traceEvents` to a root JSON dict, and then open it in Perfetto. Another workaround is to turn `with_stack=False`, which reduces the amount of events significantly.
- My intuition is that there are too many events to show in the frontend browser. One potential solution is to sample events when we're zoomed out. This idea is because, on chrome trace, you can delete a portion of the events, and it will start loading in the frontend.

lmelinda:
I also wonder how much work would it be to incorporate Perfetto UI into the trace tab? For all my large trace files, they load pretty fast in Perfetto UI and does not crash like tensor board profiler tab. If not this could resolve a lot of problems.

PR #776, opened 2023-06-29
optimize memory view using lttb sample

lh-ycx:
Enhancement for Issue #760.

Hey guys, I've optimized the speed of the memory view using the LTTB sampling (Largest-Triangle-Three-Buckets sampling, which is able to downsample time series–like data while retaining the overall shape and variability in the data).

I've tested this using a PyTorch profiler trace of 2G, and the memory view page will not get crashed and the scaling operation is smooth and rather acceptable to me.


lh-ycx:
@chaekit @aaronenyeshi I checked the failed test (test_tensorboard_end2end.py). This test compares the response from the tensorboard frontend with the manually generated and verified `result_check_file.txt`. 

Since the memory curve has been sampled, it cannot match the previous manually generated result. As a result, the test fails.

Shall I regenerate the `result_check_file.txt`? Or, is there any other method to let the test pass?

lh-ycx:
OK, I'll work on it this week.

lh-ycx:
Hi, @aaronenyeshi. Sorry for the late response.

**CI failures**

I've carefully checked the test case and found that the test would always fail in Python 3.9 and 3.10 (please refer to #785). I also find my commit is consistent with the content in `result_check_file.txt`. So there is no need to regenerate the file. Could you please rerun the CI? Thanks a lot.

**Screenshots of the before and after views**

I use a 329M trace to show the before and after views as well as the memory footprint. (A larger trace (up to GBs) can be selected but the unoptimized view would crash.)

Before view:
image

After view:
image

Before memory footprint:
image

After memory footprint:
image

aaronenyeshi (approved):
LGTM, the previous CI issues were unrelated to the changes in this diff.

PR #791, opened 2023-07-25
[kineto] Update to C++17 and make Config::kControllerIntervalMsecs in…

davidberard98:
PyTorch is on C++17 now, so Kineto should be too.

The specific issue was with Config::kControllerIntervalMsecs. This is a static constexpr member - which in pre-C++17 needed a separate namespace definition (i.e. both a declaration `static constexpr chronos::milliseconds kControllerIntervalMsecs` in the class definition and a definition `constexpr chronos::milliseconds Config::kControllerIntervalMsecs`); but after C++17 static constexpr members don't need a separate definition.

Updating to C++17 seems to fix the issue. We could probably remove the `#if __cplusplus < 201703L` handling entirely, but I'm not 100% sure every user of kineto is on C++17

### DIFF
--- a/libkineto/CMakeLists.txt
+++ b/libkineto/CMakeLists.txt
@@ -82,7 +82,7 @@ add_custom_target(libkineto_defs.bzl DEPENDS libkineto_defs.bzl)
 add_dependencies(kineto_base libkineto_defs.bzl)
 
 set_target_properties(kineto_base kineto_api PROPERTIES
-      CXX_STANDARD 14
+      CXX_STANDARD 17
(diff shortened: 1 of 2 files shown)


aaronenyeshi:
> Updating to C++17 seems to fix the issue. We could probably remove the #if __cplusplus < 201703L handling entirely, but I'm not 100% sure every user of kineto is on C++17

I assume that every user of Kineto is probably using PyTorch. So they should have moved to latest PyTorch on C++17 by now.

aaronenyeshi (approved):
LGTM!

Issue #445, opened 2021-10-13
libkineto::api().activityProfiler() returns NULL value


maharishib:
I'm not able to use activity profiler as libkineto::api().activityProfiler() returns NULL value. 
The cause for this is InitializeInjection in init.cpp is not getting called. I tried setting the value of CUDA_INJECTION64_PATH env variable but it didn't have any effect.
Any suggestions on fixing this.
Thanks.

briancoutinho:
is this using pytorch @maharishib or just using the library stand alone? Looks like you need to do an initialization of the activityprofiler initProfilerIfRegistered()

PR #812, opened 2023-10-04
Fix kineto sample programs, add build scripts and documentation.

peterg17:
👋 Hello! I built libkineto + tried to run the sample programs, but after I figured out how to compile, I immediately got a segfault. This seemed to happen for others like in this issue https://github.com/pytorch/kineto/issues/445, and I found that a call to `libstreaming_init()` before `initProfilerIfRegistered()` was necessary to avoid a null pointer in the profiler object.

This patch fixes the issues in the sample programs, and makes it easier to run them by including a build script and instructions. See screenshots below for before/after of the changes:

## Run `kineto_playground.cpp` before code change, segfault
kineto_playground_segfault_before

## Run `kineto_playground.cpp` after code change, runs and outputs chrome trace
kineto_playground_after_fix

### DIFF
--- a/libkineto/sample_programs/README.md
+++ b/libkineto/sample_programs/README.md
@@ -1,0 +1,12 @@
+
+# How to Run Sample Programs
+
+To run `kineto_playground.cpp` in the `sample_programs` folder, you can use the following steps: (Note: scripts below are hard-coded to a specific set of sample programs, you can modify them to work with a different program. TODO: make these scripts more flexible.)
+
+1. `./build-cu.sh`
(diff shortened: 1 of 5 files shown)


aaronenyeshi (approved):
Thank you for improving the sample program! Overall LGTM, running CI now, and will land after it passes.

Please double check you've signed the CLA as the check seems to be failing.

peterg17:
Thanks for the review! I just submitted the CLA again, and I couldn't find a button to re-run that CLA check job. Maybe it will run again automatically? And sounds good :)

PR #819, opened 2023-10-18
PyTorch Profiler Bug Fix to Load Supplied Directory

hamzajzmati:
Summary: Seems to be caused by D47367943, where logdir is always overridden by logdir_spec. This fix will use the first directory in case multiple are provided.

Reviewed By: jcarreiro

Differential Revision: D50370372


aaronenyeshi (approved):
LGTM!

PR #823, opened 2023-10-30
Update TEARDOWN_CUPTI to check value for 1

aaronenyeshi:
Summary:
To allow us to turn off CUPTI Teardown when features such as CUDA Graphs are used, we should check the value is actually 1.

In follow-up diffs, we can turn off TEARDOWN_CUPTI by setting env var:
TEARDOWN_CUPTI=0.

Differential Revision: D50810381

Pulled By: aaronenyeshi


xw285cornell (approved):
LGTM!

aaronenyeshi:
Landed: https://github.com/pytorch/kineto/commit/a30ca3f9509c2cfd28561abbca51328f0bdf9014

PR #824, opened 2023-10-31
Roctracer: Implement reliable data flushing on stop

mwootton:
When stopping and flushing with libroctracer there is no way to guarantee all completion notifications have been received and logged.   Torch.Profiler calls deviceSynchronize() on stopping the tracer but that ensures nothing.  It does add some delay, reducing the race.

I augmented the tracer stop procedure to enqueue an extra async op and delay until the notification for it is received.

This will fix many "flakey" unit tests that look for kernel executions that occur immediately before the end of a recording.

### DIFF
--- a/libkineto/src/RoctracerLogger.cpp
+++ b/libkineto/src/RoctracerLogger.cpp
@@ -31,11 +31,15 @@ class Flush
 public:
   std::atomic<bool> doFlush_ {false};
   std::mutex mutex_;
-  static thread_local uint64_t trip_;
+  std::atomic<uint64_t> maxCorrelationId_;
   uint64_t correlationId_ {0};
+  void reportCorrelation(const uint64_t &cid) {
+    uint64_t prev = maxCorrelationId_;
+    while (prev < cid && maxCorrelationId_.compare_exchange_weak(prev, cid))
+      {}
+  }
 };
 static Flush s_flush;
-thread_local uint64_t Flush::trip_;
 
 RoctracerLogger& RoctracerLogger::singleton() {
   static RoctracerLogger instance;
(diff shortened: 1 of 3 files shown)


aaronenyeshi (approved):
LGTM! The workaround makes sense, and should make stopping more deterministic!

mwootton:
Sorry I left an unused condition variable in there (now removed).  I was originally hoping to wait but ended up having to poll because roctracer_flush_activity_expl has to be called continuously.

I'm glad you could figure out the silly gymnastics. :)

mwootton:
@aaronenyeshi Can you hold off on this for now?  I spoke with Jeff, he would prefer to do this a different way.  A touch more overhead but less invasive.  It would only guarantee the kernels being delivered but would not require the extra operations being injected.

I'm going to log the correlation id from the most recent kernel launch.  On profiler stoppage, wait for that one.

aaronenyeshi:
> @aaronenyeshi Can you hold off on this for now? I spoke with Jeff, he would prefer to do this a different way. A touch more overhead but less invasive. It would only guarantee the kernels being delivered but would not require the extra operations being injected.
> 
> I'm going to log the correlation id from the most recent kernel launch. On profiler stoppage, wait for that one.

@mwootton , okay that works for me, I've cancelled the landing of this PR now.

mwootton:
Okay, new method is working.  Hopefully it survived the cherry-pick. :0
@jeffdaily Looks okay?

libkineto/src/RoctracerLogger.cpp
```diff
@@ -24,6 +26,21 @@ using namespace std::chrono;
...
+class Flush
+{
+public:
+  std::atomic<bool> doFlush_ {false};
+  std::mutex mutex_;
+  std::atomic<uint64_t> maxCorrelationId_;
+  uint64_t correlationId_ {0};
+  void reportCorrelation(const uint64_t &cid) {
+    uint64_t prev = maxCorrelationId_;
+    while (prev < cid && maxCorrelationId_.compare_exchange_weak(prev, cid))
```
jeffdaily, line 38:
```suggestion
    while (prev < cid && !maxCorrelationId_.compare_exchange_weak(prev, cid))
```
Should this need "!"?  compare_exchange_weak returns true if the underlying atomic value was successfully changed.

mwootton: Opps.  Good catch.  Amazingly it (appeared) to work fine as it was, or else I did an accidental edit.  compare_exchange_weak is allowed to return false spuriously (and doesn't update prev), that may be what let it break out of the loop.  Crazy.

aaronenyeshi (approved):

PR #826, opened 2023-11-02
libkineto: Fix various compiler errors with mingw-w64 based toolchain

Biswa96:


Kreijstal:
Thank you, I was having trouble compiling this under mingw

briancoutinho:
thanks for adding the fix 👍

aaronenyeshi (approved):
LGTM

Biswa96:
It would be helpful if the linter CI runs in public GitHub actions. Or at least the CI log should be accessible to everyone.

aaronenyeshi:
> It would be helpful if the linter CI runs in public GitHub actions. Or at least the CI log should be accessible to everyone.

I agree. I checked the lints, and clang-tidy seems to want to remove these two lines in ThreadUtil.cpp.
```
#include <sys/syscall.h>
#include <sys/types.h>
```

PR #829, opened 2023-11-10
Back out "Roctracer: Implement reliable data flushing on stop"

nmacchioni:
Summary:
Revert due to profiling breakage on amd;

`while (s_flush.correlationId_ != 0) ` loops indefinitely, and I don't see `s_flush.correlationId_` being decremented anywhere so that is likely the issue.

Reviewed By: xw285cornell

Differential Revision: D51185531


aaronenyeshi (approved):

aaronenyeshi:
cc @mwootton

PR #830, opened 2023-11-13
Fixed CUPTI on Windows

mantaionut:
Fixed 2 compiling errors on Windows.
Changed callbackLock_ from static to member of the class since on Windows Release in Pytorch CuptiCallbackApi::initCallbackApi gets called before the static initialisation of the mutex so it crashes.

### DIFF
--- a/libkineto/src/CuptiCallbackApi.cpp
+++ b/libkineto/src/CuptiCallbackApi.cpp
@@ -15,7 +15,6 @@
 #include <chrono>
 #include <algorithm>
 #include <mutex>
-#include <shared_mutex>
 
 #ifdef HAS_CUPTI
 #include "cupti_call.h"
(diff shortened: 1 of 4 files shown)


davidberard98:
> CuptiCallbackApi::initCallbackApi gets called before the static initialisation of the mutex so it crashes.

interesting... do you have details on how this is possible?

aaronenyeshi (approved):
LGTM! Thanks for the fix

mantaionut:
> > CuptiCallbackApi::initCallbackApi gets called before the static initialisation of the mutex so it crashes.
> 
> interesting... do you have details on how this is possible?

I  currently have a branch to implement CUPTI in pytorch and it crashes on Windows release 100% for me and in the CI too. The issue is caused by https://github.com/pytorch/pytorch/blob/main/torch/csrc/profiler/kineto_client_interface.cpp#L86 register_libkineto_client is a global object that on constructor calls libkineto_init which calls initCallbackApi if KINETO_USE_DAEMON is set. So in this case we have 2 global objects that on Windows are not created at runtime in the correct order. If you want to test on your side you can use my branch https://github.com/pytorch/pytorch/pull/111568/files without the crash fix.

davidberard98:
@mantaionut ah ok, makes a lot of sense - thanks for the explanation! basically a static initialization order fiasco problem

PR #831, opened 2023-11-15
Make MTIA activities default in Kineto

fenypatel99:
Summary: Unitrace uses default activity types for kineto. For integration with MTIA, we need to make MTIA activity types default in kineto

Differential Revision: D51353262


davidberard98 (approved):

aaronenyeshi (approved):
LGTM, should be safe to land

PR #833, opened 2023-11-17
Rework roctracer shutdown flushing

mwootton:
Reworked roctracer flush on shutdown.  Removing a race condition added while removing the original race condition. :)
Previous implementation had a deadlock in the case where the op buffer had filled and flushed on its own immediately before shutdown.

This new approach is simpler but has to continuously log correlation ids of completed async ops.  This is done on the roctracer supplied callback thread so it is not an overhead/performance issue.  Previous attempt was over-optimized at the cost of not working correctly.  Oops.

### DIFF
--- a/libkineto/src/RoctracerLogger.cpp
+++ b/libkineto/src/RoctracerLogger.cpp
@@ -11,6 +11,8 @@
 #include <cstring>
 #include <chrono>
 #include <time.h>
+#include <mutex>
+#include <unistd.h>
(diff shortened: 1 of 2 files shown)


mwootton:
@jeffdaily This look good now?

jeffdaily (approved):
@aaronenyeshi for merging.

aaronenyeshi (approved):

PR #840, opened 2023-11-29
Fix roctracer flushing issue

mwootton:
Discovered an issue with libroctracer where, immediately after startup, records are not delivered.  Not lost, just not delivered, even on a roctracer_flush_activity() call.

A hipDeviceSynchronize() resolves the issue.  Normally the application, e.g. torch.profiler, would do this synchronize before stopping the profiler.   It looks like this is not always the case though, so we need to inject an additional call.  Without it, stopLogging() may poll forever waiting for data that never comes.

Also, added a timeout to the polling loop.  This will prevent hangs.  Unit tests were the only known examples of the hang.  They should be able to detect the incorrect results if the timeout triggers.  Otherwise "no harm, no foul" applies. :)

### DIFF
--- a/libkineto/src/RoctracerLogger.cpp
+++ b/libkineto/src/RoctracerLogger.cpp
@@ -351,6 +351,7 @@ void RoctracerLogger::stopLogging() {
     return;
   logging_ = false;
 
+  hipDeviceSynchronize();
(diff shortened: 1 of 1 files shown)


mwootton:
@jeffdaily Let's go zombie bug.

PR #848, opened 2023-12-07
Re-word LOG message for Record Counters from Error info to Record counts

aaronenyeshi:
Summary:
This LOG message has been causing a lot of confusion across production jobs. People are seeing the Error info message, and believe the profiler is breaking something. This is actually a harmless message, that shows detailed record counts.

The loggers also mark Error keyword line in red, making it stand out as an Error message.

Differential Revision: D51950718

Pulled By: aaronenyeshi


anupambhatnagar (approved):
👍

briancoutinho:
+1 :)

PR #850, opened 2023-12-13
Roctracer: read kernel names from (NEW) union in the async op record

mwootton:
Roctracer now writes kernelnames into the async op records so we no longer have to glean them from the hipLaunchKernel api calls.  This allows use to populate kernel names from graph playback.
Using this value means parsing the async records at collection time since the char* names are not valid past the callback duration.
Created a structure to hold the parsed data.  Same pattern as for the api records.

### DIFF
--- a/libkineto/src/RoctracerActivityApi.cpp
+++ b/libkineto/src/RoctracerActivityApi.cpp
@@ -39,7 +39,8 @@ typedef enum {
 
 typedef enum {
   HIP_OP_DISPATCH_KIND_UNKNOWN_ = 0,
-  HIP_OP_DISPATCH_KIND_KERNEL_ = 0x11F0
+  HIP_OP_DISPATCH_KIND_KERNEL_ = 0x11F0,
+  HIP_OP_DISPATCH_KIND_TASK_ = 0x11F1
 } hip_op_dispatch_kind_t_;
 
 typedef enum {
@@ -289,6 +290,7 @@ int RoctracerActivityApi::processActivities(
         a.activityType = ActivityType::GPU_MEMSET;
         break;
       case HIP_OP_DISPATCH_KIND_KERNEL_:
+      case HIP_OP_DISPATCH_KIND_TASK_:
       default:
         if (!isLogged(ActivityType::CONCURRENT_KERNEL))
           filtered = true;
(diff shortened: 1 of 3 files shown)


mwootton:
@jeffdaily @pengsun

aaronenyeshi (approved):

aaronenyeshi:
Would this cause any backwards compatibility issues? If I were to update the hash into pytorch/pytorch with this patch, could PyTorch CI have issues with older versions of ROCm?

mwootton:
@aaronenyeshi 
Was hoping to find the issues in CI. :)
Rocm5.3 has support, but, 5.1 does not.  Rocm 5.3 was October 2022.  It silently showed up, I slept through the whole thing.
This PR came from an urgent request.  I could go back and retain the previous operation with ifdefs on the rocm version.  I'd rather we just move forward though.  Together we advance.

aaronenyeshi:
This works for me, i've approved internally. cc @malfet @kit1980 @xw285cornell for their awareness

malfet (approved):

xw285cornell:
This seems to work for the simple workload but for the llm benchmark I have, it's still shows up as "Tasks". I'll need to dig more.

libkineto/src/RoctracerActivityApi.cpp
```diff
@@ -235,127 +235,72 @@ int RoctracerActivityApi::processActivities(
...
+    const char *name = roctracer_op_string(item.domain, item.op, item.kind);
+    a.device = item.device;
+    a.resource = item.queue;
+
+    a.startTime = (item.begin + toffset) / 1000;
+    a.endTime = (item.end + toffset) / 1000;
+    a.id = item.id;
+
+    a.activityType = ActivityType::CONCURRENT_KERNEL;
+    a.activityName = item.kernelName.length() > 0 ? item.kernelName : std::string(name);
```
xw285cornell, line 265: @mwootton in that application, item.kernelName is empty and name is "Task"

mwootton: Yes, that is a different issue with looking up some kernel names.  The will come as a rocm fix.
The fact is says "Task" means I'm looking it up correctly, right? :)

xw285cornell: The original code seems to just get the name from roctracer_op_string and this change will first use item.kernelName, is that right? Why the item.kernelName is empty?

mwootton:
@xw285cornell After discussion it appears some graph kernels are being labelled as "Task" rather than "Kernel" to distinguish them.  I need to account for this when deciding to fetch the kernel names.

mwootton:
Sorry for the do-over.  We are getting out ahead of the established roctracer documentation at this point.

xw285cornell:
sounds great! I'll test it out and let you know

xw285cornell:
Confirm this works. I'll land this internally.

saleelk:
Great! Thanks

PR #856, opened 2024-01-18
ci: Use GITHUB_OUTPUT envvar instead of set-output command

arunsathiya:
`save-state` and `set-output` commands used in GitHub Actions are deprecated and [GitHub recommends using environment files](https://github.blog/changelog/2023-07-24-github-actions-update-on-save-state-and-set-output-commands/).

This PR updates the usage of `set-output` to `$GITHUB_OUTPUT`

Instructions for envvar usage from GitHub docs:

https://docs.github.com/en/actions/using-workflows/workflow-commands-for-github-actions#setting-an-output-parameter

### DIFF
--- a/.github/workflows/tb_plugin_ci.yml
+++ b/.github/workflows/tb_plugin_ci.yml
@@ -24,9 +24,9 @@ jobs:
           echo $GITHUB_BASE_REF
           if [ $GITHUB_BASE_REF == "plugin/vnext" ]
           then
-            echo "::set-output name=matrix::{\"python-version\":[3.8], \"cuda-version\":[\"cpu\"], \"pytorch-version\":[\"nightly\"]}"
+            echo "matrix={\"python-version\":[3.8], \"cuda-version\":[\"cpu\"], \"pytorch-version\":[\"nightly\"]}" >> $GITHUB_OUTPUT
           else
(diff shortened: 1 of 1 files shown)


.github/workflows/tb_plugin_ci.yml
```diff
@@ -24,9 +24,9 @@ jobs:
           echo $GITHUB_BASE_REF
           if [ $GITHUB_BASE_REF == "plugin/vnext" ]
           then
-            echo "::set-output name=matrix::{\"python-version\":[3.8], \"cuda-version\":[\"cpu\"], \"pytorch-version\":[\"nightly\"]}"
+            echo "matrix={\"python-version\":[3.8], \"cuda-version\":[\"cpu\"], \"pytorch-version\":[\"nightly\"]}" >> $GITHUB_OUTPUT
```
aaronenyeshi, line 27: Does $GITHUB_OUTPUT need to be in quotes like "$GITHUB_OUTPUT"?

As shown in https://docs.github.com/en/actions/using-workflows/workflow-commands-for-github-actions#setting-an-output-parameter

aaronenyeshi (changes requested):
See comment. Also ran the CI

.github/workflows/tb_plugin_ci.yml
arunsathiya: I see that the last tests passed, but good to match documentation! So, I have updated in https://github.com/pytorch/kineto/pull/856/commits/c89e3ba49b7187126ed4d0bce1725b8871c5c752.

aaronenyeshi (approved):
LGTM! merging it now

PR #861, opened 2024-01-25
Enable profiler initialization for CPU only use cases

briancoutinho:
Summary:
This PR enables profiler initialization for CPU only use cases. The main goal is to enable on-demand profiling with a daemon when using CPU only mode of PyTorch.
* When CUDA is available the profiler is initialized on first CUDA stream creation (or lazily when profiler is run).
* Since the CUDA stream creation callback does not exist on CPU only PyTorch the profiler is never initied on its own.
* Thus the job does not register with Dynolog when we set "KINETO_USE_DAEMON" env variable to set. 
* This PR tries to fix this.

Part of the resolution for https://github.com/pytorch/pytorch/issues/112389

Differential Revision: D53096124

### DIFF
--- a/libkineto/src/init.cpp
+++ b/libkineto/src/init.cpp
@@ -13,6 +13,7 @@
 // @lint-ignore-every CLANGTIDY facebook-hte-RelativeInclude
 #include "ActivityProfilerProxy.h"
 #include "Config.h"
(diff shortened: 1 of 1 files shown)


aaronenyeshi (approved):
LGTM

PR #868, opened 2024-02-09
Fix CUPTI timestamp on Windows

mantaionut:
I am working on enabling CUPTI on pytorch on Windows https://github.com/pytorch/pytorch/pull/111568 but I had some flaky tests in the CI caused by https://github.com/pytorch/pytorch/blob/main/test/test_foreach.py#L65. After some debugging, I realized that sometimes the tensor_apply_kernel will be `TraceActivity outside of profiling window:`. The issue is caused by the fact that on Windows timestamp from CUPTI is on Windows the monotonic time so we try to sync the system clock with the steady clock.

But after CUDA 11.6 it was added [cuptiActivityRegisterTimestampCallback ](https://forums.developer.nvidia.com/t/cuda-profiler-tools-interface-cupti-for-cuda-toolkit-11-6-is-now-available/200030) for which we register a callback that returns the system time which seems to be the better option.

This branch is used in https://github.com/pytorch/pytorch/pull/111568 which runs the unit tests which pass.
Let me know if there are some additional tests I should do.


libkineto/src/CuptiActivityProfiler.cpp
```diff
@@ -840,6 +840,14 @@ void CuptiActivityProfiler::configure(
       timestamp = system_clock::now();
     }
 #ifdef HAS_CUPTI
+#ifdef _WIN32
+    CUPTI_CALL(
+        cuptiActivityRegisterTimestampCallback([]() -> uint64_t {
```
aaronenyeshi, line 845: I'm wondering if roctracer also needs a similar timestamp callback change. @mwootton , @jeffdaily

aaronenyeshi (approved):
Looks great, will run the CI, and need to double check if we need to do the same for ROCm.

libkineto/src/CuptiActivityProfiler.cpp
aaronenyeshi: I couldn't find the hip equivalent to this CUPTI API: https://docs.nvidia.com/cupti/modules.html#group__CUPTI__ACTIVITY__API_1gc168c4a9b5f4e9bd7553ca26118a307d

mantaionut: Unfortunately, I do not have an amd gpu to check for roctracer. Anything else I could do to help move this forward?

mwootton:
Oh, the reckoning has come.  Nice.

So Kineto has been using the incorrect clock for timestamps since it's inception.  This is because cupti uses the incorrect clock and there was no workaround since it was collecting timestamps internally.  Cupti did introduce a way to override the clock selection (as mentioned in the initial comment) but that would introduce a "cuda minimum version" to Kineto.  If that is okay we should fix this now.

So cupti classically used CLOCK_REALTIME for timestamps.  This clock is a best guess at wall time.  It is periodically corrected via NTP (to remain close to wall time).  Those corrections are applied instantaneously, so the clock can jump backwards.  This is not suitable for profiling because it will invert the ordering of timestamps (i.e. "start" happening after "end").  If you look through Kineto source there are comments explaining the blocks of code designed to detect and throw out events that fall prey to this.  No explanation, just "these come in wrong sometimes".  I suspect that is just the clock issue.

On unix the CLOCK_MONOTONIC is, wait for it, monotonically increasing.  This still has NTP corrections but they are done gradually with rate slewing.  No inverted timestamps.  Roctracer (and anything else I've touched in rocm) use this clock.  The Kineto roctracer implementation has to convert from monotonic to realtime (the kineto internal clock) so I am literally adding the inversions back in. :(  Status quo.

On Windows I'm not sure how to best get a monotonic clock.  You could probably use std::chrono::steady_clock.  I couldn't understand the initial comment, but is sounds like you already have a monotonic (steady) clock on windows.  If cupti is reporting timestamps in that domain (monotonic, steady), you are golden.  If cupti is reporting timestamps in "realtime" or "system" clock then you are doomed exactly like Kineto compiled on Linux is.

If we are willing to accept the Cupti >= CUDA 11.6 limitation, then we should fix Kineto immediately.  Using a non-monotonic clock for profiling is just WRONG.

mwootton:
Oh found this in the cuda api docs:

_This function registers a callback function to obtain timestamp of user's choice instead of using CUPTI provided timestamp. By default CUPTI uses different methods, based on the underlying platform, to retrieve the timestamp Linux and Android use clock_gettime(CLOCK_REALTIME, ..) Windows uses QueryPerformanceCounter() Mac uses mach_absolute_time() QNX uses ClockCycles() Timestamps retrieved using these methods are converted to nanosecond if needed before usage._

Yeah, so here they fess up to using the wrong clock on Linux and Android.  They are using QueryPerformanceCounter() on Windows, which is monotonic.  So Windows will actually be CORRECT if you use QueryPerformanceCounter() to generate your timestamps.

mwootton:
@aaronenyeshi @mantaionut

libkineto/src/CuptiActivityProfiler.cpp
```diff
@@ -840,6 +840,14 @@ void CuptiActivityProfiler::configure(
...
+          auto system = std::chrono::time_point_cast<std::chrono::nanoseconds>(
+              std::chrono::system_clock::now());
```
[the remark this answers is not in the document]
mantaionut: Well we already convert the CUPTI time to system time on Windows https://github.com/pytorch/kineto/blob/main/libkineto/src/CuptiActivityPlatform.cpp#L21. The time needs to be in between the system clock start and end https://github.com/pytorch/kineto/blob/a1c3fea49769bbf31343554cdb569d5e88ac7242/libkineto/src/CuptiActivityProfiler.cpp#L408 so it will happen to be outside that window. So in this case it makes sense to either use system time on Windows too, or maybe we could either convert to monotonic on every platform or maybe change on windows to have captureWindowStartTime_  and captureWindowEndTime_ use monotonic time too.

libkineto/src/CuptiActivityProfiler.cpp
aaronenyeshi: We have some internal CI , let me try importing in there.

aaronenyeshi:
Sorry for the late reply, and really great discussions, thanks @mwootton. I feel we need to discuss this further, and I will create a Github Issue to continue this. 

To unblock @mantaionut, I think the windows side diff looks okay, I will port this internally and test to see if AMD is okay too.

aaronenyeshi:
> If we are willing to accept the Cupti >= CUDA 11.6 limitation, then we should fix Kineto immediately. Using a non-monotonic clock for profiling is just WRONG.

CUDA >=11.6 limitation should be okay as pytorch/pytorch is supporting 11.8 and 12.1 for Windows and newer. https://pytorch.org/get-started/locally/

Issue #843, opened 2023-12-05
[Bug] Operator device duration is zero in tensorboard


shinytang6:

shinytang6:
I think the main reason is that the external id of cpu_op are no longer associated with the external id of runtime/kernel op, so tensorboard plugin cannot find the correlation.

PR #873, opened 2024-02-20
Correct external id for runtime apis and kernels

fwenguang:
Fix https://github.com/pytorch/kineto/issues/843


aaronenyeshi (approved):
LGTM!

PR #874, opened 2024-02-21
Support PrivateUse1 key 

fwenguang:
Summary:
1. Add some ActivityTypes for PrivateUse1.
2. Add interface pushCorrelationId and similar interfaces to IActivityProfilerSession.
3. Move output_base.h to include dir so that PrivateUse1 user can call member functions of ActivityLogger.

### DIFF
--- a/libkineto/include/ActivityType.h
+++ b/libkineto/include/ActivityType.h
@@ -40,20 +40,11 @@ enum class ActivityType {
     HPU_OP, // HPU host side runtime event
     XPU_RUNTIME, // host side xpu runtime events
     COLLECTIVE_COMM, // collective communication
-
-    // PrivateUse1 Activity types
-    PRIVATEUSE1_USER_ANNOTATION,
-    PRIVATEUSE1_MEMCPY,
-    PRIVATEUSE1_MEMSET,
-    PRIVATEUSE1_CONCURRENT_KERNEL, // privateUse1 on-device kernels
     PRIVATEUSE1_RUNTIME, // host side privateUse1 runtime events
     PRIVATEUSE1_DRIVER, // host side privateUse1 driver events
-    PRIVATEUSE1_SYNC, // synchronization events between runtime and kernels
 
     ENUM_COUNT, // This is to add buffer and not used for any profiling logic. Add your new type before it.
     OPTIONAL_ACTIVITY_TYPE_START = GLOW_RUNTIME,
-    PRIVATEUSE1_ACTIVITY_TYPE_START = PRIVATEUSE1_USER_ANNOTATION,
-    PRIVATEUSE1_ACTIVITY_TYPE_END = PRIVATEUSE1_SYNC,
 };
 
 const char* toString(ActivityType t);
--- a/libkineto/include/ActivityType.h
+++ b/libkineto/include/ActivityType.h
@@ -10,6 +10,7 @@
 
 #include <array>
 #include <string>
+#include <set>
 
 namespace libkineto {
 
@@ -40,12 +41,25 @@ enum class ActivityType {
     XPU_RUNTIME, // host side xpu runtime events
     COLLECTIVE_COMM, // collective communication
 
+    // PrivateUse1 Activity types
+    PRIVATEUSE1_USER_ANNOTATION,
+    PRIVATEUSE1_MEMCPY,
+    PRIVATEUSE1_MEMSET,
+    PRIVATEUSE1_CONCURRENT_KERNEL, // privateUse1 on-device kernels
+    PRIVATEUSE1_EXTERNAL_CORRELATION,
+    PRIVATEUSE1_RUNTIME, // host side privateUse1 runtime events
+    PRIVATEUSE1_DRIVER, // host side privateUse1 driver events
+    PRIVATEUSE1_SYNC, // synchronization events between runtime and kernels
+
     ENUM_COUNT, // This is to add buffer and not used for any profiling logic. Add your new type before it.
     OPTIONAL_ACTIVITY_TYPE_START = GLOW_RUNTIME,
+    PRIVATEUSE1_ACTIVITY_TYPE_START = PRIVATEUSE1_USER_ANNOTATION,
+    PRIVATEUSE1_ACTIVITY_TYPE_END = PRIVATEUSE1_SYNC,
 };
 
 const char* toString(ActivityType t);
 ActivityType toActivityType(const std::string& str);
+bool hasPrivateUse1Type(const std::set<ActivityType>& activityTypes);
 
 // Return an array of all activity types except COUNT
 constexpr int activityTypeCount = (int)ActivityType::ENUM_COUNT;
--- a/libkineto/include/ActivityType.h
+++ b/libkineto/include/ActivityType.h
@@ -40,6 +40,9 @@ enum class ActivityType {
     HPU_OP, // HPU host side runtime event
     XPU_RUNTIME, // host side xpu runtime events
     COLLECTIVE_COMM, // collective communication
+
+    // PRIVATEUSE1 Activity types are used for custom backends.
+    // The corresponding device type is `DeviceType::PrivateUse1` in PyTorch.
     PRIVATEUSE1_RUNTIME, // host side privateUse1 runtime events
     PRIVATEUSE1_DRIVER, // host side privateUse1 driver events
 
--- a/libkineto/include/ActivityType.h
+++ b/libkineto/include/ActivityType.h
@@ -46,7 +46,6 @@ enum class ActivityType {
     PRIVATEUSE1_MEMCPY,
     PRIVATEUSE1_MEMSET,
     PRIVATEUSE1_CONCURRENT_KERNEL, // privateUse1 on-device kernels
-    PRIVATEUSE1_EXTERNAL_CORRELATION,
     PRIVATEUSE1_RUNTIME, // host side privateUse1 runtime events
     PRIVATEUSE1_DRIVER, // host side privateUse1 driver events
     PRIVATEUSE1_SYNC, // synchronization events between runtime and kernels
--- a/libkineto/include/IActivityProfiler.h
+++ b/libkineto/include/IActivityProfiler.h
@@ -66,6 +66,10 @@ struct ResourceInfo {
   int64_t deviceId;       // id of device which owns this resource (specified in DeviceInfo.id)
   const std::string name; // resource name
 };
+
+using getLinkedActivityCallback =
+  std::function<const ITraceActivity*(int32_t)>;
+
 /* IActivityProfilerSession:
  *   an opaque object that can be used by a high level profiler to
  *   start/stop and return trace events.
@@ -81,8 +85,6 @@ class IActivityProfilerSession {
   // stop the trace collection synchronously
   virtual void stop() = 0;
 
-  virtual void reset() = 0;
-
   TraceStatus status() {
     return status_;
   }
@@ -94,7 +96,7 @@ class IActivityProfilerSession {
   virtual void processTrace(ActivityLogger& logger) = 0;
 
   virtual void processTrace(ActivityLogger& logger,
-    std::function<const ITraceActivity*(int32_t)> getLinkedActivity,
+    getLinkedActivityCallback getLinkedActivity,
     int64_t startTime, int64_t endTime) {
     processTrace(logger);
   }
--- a/libkineto/include/IActivityProfiler.h
+++ b/libkineto/include/IActivityProfiler.h
@@ -81,6 +81,8 @@ class IActivityProfilerSession {
   // stop the trace collection synchronously
   virtual void stop() = 0;
 
+  virtual void reset() = 0;
+
   TraceStatus status() {
     return status_;
   }
@@ -90,6 +92,12 @@ class IActivityProfilerSession {
 
   // processes trace activities using logger
   virtual void processTrace(ActivityLogger& logger) = 0;
+
+  virtual void processTrace(ActivityLogger& logger,
+    std::function<const ITraceActivity*(int32_t)> getLinkedActivity,
+    int64_t startTime, int64_t endTime) {
+    processTrace(logger);
+  }
 
   // returns device info used in this trace, could be nullptr
   virtual std::unique_ptr<DeviceInfo> getDeviceInfo() = 0;
@@ -102,6 +110,12 @@ class IActivityProfilerSession {
 
   // XXX define trace formats
   // virtual save(string name, TraceFormat format)
+
+  virtual void pushCorrelationId(uint64_t id) {}
+  virtual void popCorrelationId() {}
+
+  virtual void pushUserCorrelationId(uint64_t id) {}
+  virtual void popUserCorrelationId() {}
 
  protected:
   TraceStatus status_ = TraceStatus::READY;
--- a/libkineto/include/IActivityProfiler.h
+++ b/libkineto/include/IActivityProfiler.h
@@ -23,7 +23,6 @@
 
 namespace libkineto {
 
-using namespace KINETO_NAMESPACE;
 struct CpuTraceBuffer;
 
 #ifdef _MSC_VER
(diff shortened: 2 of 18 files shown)


libkineto/include/ActivityType.h
```diff
@@ -40,12 +41,25 @@ enum class ActivityType {
...
 
+    // PrivateUse1 Activity types
+    PRIVATEUSE1_USER_ANNOTATION,
+    PRIVATEUSE1_MEMCPY,
+    PRIVATEUSE1_MEMSET,
+    PRIVATEUSE1_CONCURRENT_KERNEL, // privateUse1 on-device kernels
+    PRIVATEUSE1_EXTERNAL_CORRELATION,
+    PRIVATEUSE1_RUNTIME, // host side privateUse1 runtime events
+    PRIVATEUSE1_DRIVER, // host side privateUse1 driver events
+    PRIVATEUSE1_SYNC, // synchronization events between runtime and kernels
```
aaronenyeshi, line 52: Could we re-use the above ActivityTypes for most of these, and add PRIVATEUSE1_RUNTIME and PRIVATEUSE1_DRIVER only like MTIA above?

aaronenyeshi (changes requested):

libkineto/include/ActivityType.h
fwenguang: We need to get the DeviceType::PrivateUse1 from Activities. So, i think these five ActivityTypes https://github.com/pytorch/pytorch/pull/120556/files#diff-11a30835eab7801f23620d212a8b7347c87530432971d8777a92f937d31a0bb9R354-R359 are still needed.

Of course, like EXTERNAL_CORRELATION,  can indeed be reused. I will delete PRIVATEUSE1_EXTERNAL_CORRELATION.

fwenguang: I noticed that XPU uses another way to get DeviceType. And [pr](https://github.com/pytorch/pytorch/pull/120185/files#diff-11a30835eab7801f23620d212a8b7347c87530432971d8777a92f937d31a0bb9R327) had been merged. If so, We can indeed reuse most of the these, and only need to add PRIVATEUSE1_RUNTIME and PRIVATEUSE1_DRIVER.

fwenguang:
Hi, @aaronenyeshi . Could you help to review this pr? Is there any other suggestions?

libkineto/src/CuptiActivityProfiler.cpp
```diff
@@ -288,7 +289,10 @@ void CuptiActivityProfiler::processTraceInternal(ActivityLogger& logger) {
 
   for (const auto& session : sessions_) {
     LOG(INFO) << "Processing child profiler trace";
-    session->processTrace(logger);
+    session->processTrace(logger,
+        std::bind(&CuptiActivityProfiler::cpuActivity, this, std::placeholders::_1),
+        captureWindowStartTime_,
+        captureWindowEndTime_);
```
aaronenyeshi, line 295: Why are we passing these 3 new arguments? could you please add some comments or description in the commit, I don't see it used, other than a new virtual function in IActivityProfiler.

libkineto/src/ActivityProfilerProxy.cpp
```diff
@@ -92,8 +91,7 @@ bool ActivityProfilerProxy::isActive() {
 }
 
 void ActivityProfilerProxy::pushCorrelationId(uint64_t id) {
-  CuptiActivityApi::pushCorrelationID(id,
-    CuptiActivityApi::CorrelationFlowType::Default);
+  controller_->pushCorrelationId(id);
```
aaronenyeshi, line 94: simple question: why do we want to pass it down into the controller and the activity profilers? I guess you may have a follow-up diff?

aaronenyeshi (commented):
overall looks okay, I have a few questions. Also, will import so we can fix any internal build systems

libkineto/test/ConfigTest.cpp
```diff
@@ -122,6 +122,12 @@ TEST(ParseTest, ActivityTypes) {
   EXPECT_TRUE(cfg2.parse("ACTIVITY_TYPES = xpu_Runtime"));
   EXPECT_EQ(cfg2.selectedActivityTypes(),
     std::set<ActivityType>({ActivityType::XPU_RUNTIME}));
+
+  EXPECT_TRUE(cfg2.parse("ACTIVITY_TYPES=privateuse1_Runtime,privateuse1_kernel"));
+  EXPECT_EQ(cfg2.selectedActivityTypes(),
+    std::set<ActivityType>({ActivityType::PRIVATEUSE1_RUNTIME,
+                            ActivityType::PRIVATEUSE1_CONCURRENT_KERNEL}));
```
aaronenyeshi, line 129: I think we meant to use PRIVATEUSE1_DRIVER here

libkineto/include/ActivityType.h
aaronenyeshi: cc @xunsongh, would be great to get cross-review since XPU and PrivateUse1 may share very similar implementations.

xunsongh: I have no objections to this section, and I appreciate your efforts in reusing the code while attempting to maintain consistency with the XPU modifications. Thank you.

libkineto/src/ActivityProfilerProxy.cpp
xunsongh: I share the same question. From your changes, it seems you're aiming to extend the `IActivityProfilerSession` with methods to allow sessions running on `privateuse1` to also access External ID information. However, the current implementation suggests that sessions are intended to serve solely as child profilers. If `privateuse1` employs a custom main profiler, it should be directly registered with the `registerProfiler()` method rather than being added to the session hierarchy using the `configChildProfiler()` method. If this is not the case, then the session should adhere to Kineto's design for profilers, where all manipulations of External IDs should fall under the purview of the main profiler, not child sessions.

This is just a perspective from me outside of Kineto's development, and I hope it doesn't impact your decision-making process if there are any inaccuracies in my understanding. Thank you.

libkineto/src/CuptiActivityProfiler.cpp
fwenguang: I have implemented my kineto plugin in an externel repo, where i used this new virtual function. 

According to https://github.com/pytorch/kineto/pull/662#issuecomment-1298935447,  I implemented my ActivityProfiler and ActivityProfilerSession.  My ActivityProfiler catching events for kernels, runtime apis, correlation_external_ids and so on. Like cupti, We also only save events catched in the range from captureWindowStartTime_ to captureWindowEndTime_.  So, when `processTrace`, we need to know this two timepoint.  And we want to link our runtime apis and kernels  to  their related cpu op. So,  we passing `CuptiActivityProfiler::cpuActivity` function.

Sadly for that our repo is not OSS yet, It's ongoing. So, I can't share a link for my code.

libkineto/src/ActivityProfilerProxy.cpp
fwenguang: > simple question: why do we want to pass it down into the controller and the activity profilers? I guess you may have a follow-up diff?

Some answers in https://github.com/pytorch/kineto/pull/874#discussion_r1508428353. In my plugin, we want to catch events  of EXTERNAL_CORRELATION, so we need to call pushCorrelationId.

libkineto/test/ConfigTest.cpp
fwenguang: Yeah, done.

libkineto/src/ActivityProfilerProxy.cpp
aaronenyeshi: Actually, I think the new implementation is supposed to have every profiler as a Child Profiler, so your case is okay. In the future we should try to separate out CUPTI and AMD from the main profiler, and into child profiler sessions instead. Thank you for following the recommendations. cc @briancoutinho , @gdankel

fwenguang:
Hi @aaronenyeshi , is this PR ready to be merged?

libkineto/include/ActivityType.h
```diff
@@ -39,13 +40,16 @@ enum class ActivityType {
     HPU_OP, // HPU host side runtime event
     XPU_RUNTIME, // host side xpu runtime events
     COLLECTIVE_COMM, // collective communication
+    PRIVATEUSE1_RUNTIME, // host side privateUse1 runtime events
```
briancoutinho, line 43: Can you add a comment that PRIVATE USE is for custom backends that are not yet released or widely known

libkineto/src/ActivityType.cpp
```diff
@@ -60,6 +62,17 @@ ActivityType toActivityType(const std::string& str) {
   throw std::invalid_argument(fmt::format("Invalid activity type: {}", str));
 }
 
+bool hasPrivateUse1Type(const std::set<ActivityType>& activityTypes) {
+  for (const auto type : activityTypes) {
```
briancoutinho, line 66: can you just use std::set count() method ?

libkineto/src/ActivityProfilerProxy.cpp
briancoutinho: +1 on this thanks for implementing this as a child module @xunsongh  this could hopefully fix the shortcomings of child profiler architecture today.

libkineto/include/IActivityProfiler.h
```diff
@@ -81,6 +81,8 @@ class IActivityProfilerSession {
   // stop the trace collection synchronously
   virtual void stop() = 0;
 
+  virtual void reset() = 0;
```
briancoutinho, line 84: can you add a comment what the reset() does? Just a quick question, why have a reset() why not create a new session that starts with fresh state?

libkineto/include/IActivityProfiler.h
```diff
@@ -91,6 +93,12 @@ class IActivityProfilerSession {
   // processes trace activities using logger
   virtual void processTrace(ActivityLogger& logger) = 0;
 
+  virtual void processTrace(ActivityLogger& logger,
```
briancoutinho, line 96: is your intent here to have a default implementation that ignores the getLinkedActivity() callback?

libkineto/include/IActivityProfiler.h
```diff
@@ -91,6 +93,12 @@ class IActivityProfilerSession {
...
+    std::function<const ITraceActivity*(int32_t)> getLinkedActivity,
```
briancoutinho, line 97: Please add a using typedef :) 
using getLinkedActivityCallback = std::function<const ITraceActivity*(int32_t)>;

libkineto/src/CuptiActivityProfiler.cpp
briancoutinho: yea @fwenguang  just add a comment that cpuActivity() function here is getting the linked cpuActivity :)
appreciate you implementing this!

libkineto/src/CuptiActivityProfiler.cpp
```diff
@@ -944,6 +948,9 @@ void CuptiActivityProfiler::stopTraceInternal(
 
 void CuptiActivityProfiler::resetInternal() {
   resetTraceData();
+  for (auto& session : sessions_) {
+    session->reset();
```
briancoutinho, line 952: If you see resetTraceData() in the line above the sessions_ object is cleared ! So after line 951 sessions_ is empty.
This is why i was asking if you really need a reset() API ?

libkineto/src/CuptiActivityProfiler.cpp
```diff
@@ -1154,6 +1164,38 @@ CuptiActivityProfiler::getLoggerMetadata() {
   return loggerMD;
 }
 
+void CuptiActivityProfiler::pushCorrelationId(uint64_t id) {
```
briancoutinho, line 1167: thank you for abstracting this :)

libkineto/src/CuptiActivityProfiler.cpp
```diff
@@ -1080,7 +1087,8 @@ void CuptiActivityProfiler::finalizeTrace(const Config& config, ActivityLogger&
   if (!process_name.empty()) {
     logger.handleDeviceInfo(
         {pid, process_name, "CPU"}, captureWindowStartTime_);
-    if (!cpuOnly_) {
+    if (!cpuOnly_ &&
+        !libkineto::hasPrivateUse1Type(config.selectedActivityTypes())) {
```
briancoutinho, line 1091: @fwenguang I understand what you are trying to do here. When NVIDIA or AMD GPU or Private use1 is not present you don't want this to run. How about we update this to not be specific to private use 1?  
Like 
hasAcceleratorActivities() ? 
and this can check for GPU events, HIP (AMD) events , MTIA events and PrivateUse1 events?

briancoutinho (commented):
Thank you so much for improving the code base here and abstracting the plugin API :) really appreciate it.
Besides some code suggestions I had a few high level feedback on
1. reset() API, i don't think its needed since sessions are always short lived and clean up automatically when resetInternal -> resetTraceData is caleld
1. Comment on line 1003, we need to think a bit on how we calle cpuOnly to more generic thing like hasAccelerator() or something like that

libkineto/src/CuptiActivityProfiler.cpp
briancoutinho: cc @aaronenyeshi wdyt?

libkineto/include/ActivityType.h
kevint324: https://pytorch.org/tutorials/advanced/privateuseone.html

"Privateuse1"  is the recommended way for any new device backend to plugin itself into pytorch despite of whether public released or widely known.

The naming "private use" is indeed a little bit confusing but I guess it's just historical reason.

fwenguang: updated

libkineto/src/ActivityType.cpp
fwenguang: Okey, Done.

libkineto/include/IActivityProfiler.h
fwenguang: I have removed this addition. It is really not needed.

libkineto/include/IActivityProfiler.h
fwenguang: yea, i replace the function in caller with this new one. Add a defalut implementation in order to not break others child ProfilerSession.

libkineto/include/IActivityProfiler.h
fwenguang: Updated.

libkineto/src/CuptiActivityProfiler.cpp
fwenguang: Thanks for your correction. I didn't noticed that sessions are cleared in resetTraceData() before. Already remove this addition.

briancoutinho:
Thanks changes look great, I think @aaronenyeshi needs to import to fb and we can check on results there

aaronenyeshi:
Thanks for the changes, I will import the latest version. There are several internal builds/extensions that I need to fix before landing.

aaronenyeshi (approved):

fwenguang:
Hi @aaronenyeshi , is there anything i need to do about these 2 failed internal checks?

fwenguang:
Hi @aaronenyeshi , Sorry to bother you again, this pr has not been updated for a long time. please help to handle this pr. Thank you.

aaronenyeshi:
> Hi @aaronenyeshi , Sorry to bother you again, this pr has not been updated for a long time. please help to handle this pr. Thank you.

Sorry about this, I will take another look today. There are a bunch of failures due to internal code. Will give you an update soon.

libkineto/src/CuptiActivityProfiler.cpp
aaronenyeshi: @briancoutinho I think this is okay for Florian's use case, but we should try to clean this up later.

aaronenyeshi:
I've made slight updates to make internal code / abstractions work. Will push out now

aaronenyeshi (approved):

aaronenyeshi:
I cannot push it out (since I'm not the author of the diff), but I will try to land it internally once it passes all signals.

PR #882, opened 2024-02-29
reduce verbosity of some dynolog warnings

davidberard98:
Summary:
User reports are that these will appear frequently when used in OSS with
KINETO_USE_DAEON=1

Differential Revision: D54347978


aaronenyeshi (approved):
This looks okay to me, how often so these error messages occur? also cc @briancoutinho

PR #884, opened 2024-03-02
Fix stream ID for cuda sync events at GPU scope

briancoutinho:
Summary: Stream ID was being casted to large 2^32 -1 but it should be -1

Differential Revision: D54445884


aaronenyeshi (approved):
LGTM!

PR #896, opened 2024-04-03
Fix output_json decimal outputs to be 3 digit zero padded

sraikund16:
Summary: Currently the output for kineto in the nanosecond granularity is incorrect because the decimals are not zero padded. Add formatter in string to make sure values are zero padded in all decimals.

Differential Revision: D55668543


aaronenyeshi (approved):
Looks great to me!

PR #900, opened 2024-04-08
Initialize BackwardNode with start_time, end_time and tid in torch_tb_profiler

wraymo:
Summary:

Currently when building backward modules in torch_tb_profiler, we initialize `start_time` and `end_time` to `None`, and `tid` to `0` and try to fill these three fields in a bottom–up manner. However, during this process if at least one of the leaf nodes is a `BackwardNode` and the initial values of `start_time` and `end_time` are `None`, we will encounter errors in `self.children.sort(key=lambda x: (x.start_time, -x.end_time))` when calling `fill_stats`. The parent's `tid` can also be set to 0. To fix this bug, this PR assigns a `BackwardNode` with initial values of `start_time`, `end_time`, `tid`. These values can be updated later when traversing the tree backwards.

### DIFF
--- a/tb_plugin/torch_tb_profiler/profiler/op_tree.py
+++ b/tb_plugin/torch_tb_profiler/profiler/op_tree.py
@@ -271,8 +271,8 @@ class OpTreeBuilder:
             return
 
         if isinstance(node, ModuleNode):
-            backward_node = BackwardNode(name=node.name + '.backward', start_time=None, end_time=None,
-                                         type='backward', tid=0)
+            backward_node = BackwardNode(name=node.name + '.backward', start_time=node.start_time,
+                                         end_time=node.end_time, type='backward', tid=node.tid)
             if parent is None:
                 result.append(backward_node)
             else:


aaronenyeshi (approved):
Makes sense, thank you!

xuzhao9 (approved):

xuzhao9:
LGTM, thanks for fixing this!

wizzniu:
> Summary:
> 
> Currently when building backward modules in torch_tb_profiler, we initialize `start_time` and `end_time` to `None`, and `tid` to `0` and try to fill these three fields in a bottom–up manner. However, during this process if at least one of the leaf nodes is a `BackwardNode` and the initial values of `start_time` and `end_time` are `None`, we will encounter errors in `self.children.sort(key=lambda x: (x.start_time, -x.end_time))` when calling `fill_stats`. The parent's `tid` can also be set to 0. To fix this bug, this PR assigns a `BackwardNode` with initial values of `start_time`, `end_time`, `tid`. These values can be updated later when traversing the tree backwards.

Hi, @wraymo  I noticed that in your solution, a BackwardNode is initialized with the corresponding ModuleNode's `start_time`, `end_time` and `tid`. I think it could lead to a incorrect stats result of its parent node, though this solution can fix the error log. In fact, if there exists a such leaf `BackwardNode`, it proves the corresponding `ModuleNode`'s children and children's children don't associate with any backward node. And in this case, this leaf `BackwardNode` I think is meaningless and it shouldn't be involved in its parent's fill_stats. So I think only removing this leaf `BackwardNode` from its parent's children list is okay. By the way, the BackwardNode 's initial value of start_time and end_time shouldn't be set using its corresponding ModuleNode, and it would take the forward nodes' stats into BackwardNode while traversing the tree backwards, which is unexpected for BackwardNode's definition. Of course, above is just my opinion and we can discuss further. cc @aaronenyeshi @xuzhao9

Issue #894, opened 2024-03-28
Roctracer crashes when number of samples too high


aaronenyeshi:
As a workaround, we are currently using a thread_local mutex and thread_local unordered_map in RoctracerLogger.cpp's `api_callback()` function. See: https://github.com/pytorch/kineto/blob/main/libkineto/src/RoctracerLogger.cpp#L99-L101

This is strange. We expect that Roctracer runtime events will **only enter and exit `api_callback()`, one event at a time per thread, and in-order of start and end**. However, if we remove the mutex, GPU tracing will crash during high QPS (examples/sec).

## Crashing stack trace (replaced the internal call stack with ...):
```
*** Aborted at 1711481163 (Unix time, try 'date -d @1711481163') ***
*** Signal 11 (SIGSEGV) (0x659b9001bb778) received by PID 1816440 (pthread TID 0x7fad2bbd0640) (linux TID 1994978) (maybe from PID 1816440, UID 416185) (code: -6), stack trace: ***
    @ 000000004eecfb1d folly::symbolizer::(anonymous namespace)::signalHandler(int, siginfo_t*, void*)
                       ./fbcode/folly/experimental/symbolizer/SignalHandler.cpp:449
    @ 000000000004453f (unknown)
    @ 00000000000c567a std::_Rb_tree_insert_and_rebalance(bool, std::_Rb_tree_node_base*, std::_Rb_tree_node_base*, std::_Rb_tree_node_base&)
    @ 0000000046ed6bd5 std::_Rb_tree_iterator<std::pair<unsigned long const, unsigned long> > std::_Rb_tree<unsigned long, std::pair<unsigned long const, unsigned long>, std::_Select1st<std::pair<unsigned long const, unsigned long> >, std::less<unsigned long>, std::allocator<std::pair<unsigned long const, unsigned long> > >::_M_emplace_hint_unique<std::piecewise_construct_t const&, std::tuple<unsigned long const&>, std::tuple<> >(std::_Rb_tree_const_iterator<std::pair<unsigned long const, unsigned long> >, std::piecewise_construct_t const&, std::tuple<unsigned long const&>&&, std::tuple<>&&)
                       fbcode/third-party-buck/platform010/build/libgcc/include/c++/trunk/bits/stl_tree.h:2398
    @ 000000005adee4da RoctracerLogger::api_callback(unsigned int, unsigned int, void const*, void*)
[...]

PR #902, opened 2024-04-11
Fix roctracer crash when number of samples too high

aaronenyeshi:
Summary:
There is a map access at the bottom of the `api_callback()` function that is storing the correlation id and external correlation id relation. This map is shared between threads, and is causing a SIGSEGV crash during RBTree insert_and_rebalance.

To fix this, we add a mutex for the map. This is likely why the previous thread_local mutex was fixing crashes, due to delayed timing, so we can remove the thread_local mutex now.

Resolves: https://github.com/pytorch/kineto/issues/894

Differential Revision: D56017970

Pulled By: aaronenyeshi

### DIFF
--- a/libkineto/src/RoctracerLogger.cpp
+++ b/libkineto/src/RoctracerLogger.cpp
@@ -97,22 +97,16 @@ void RoctracerLogger::api_callback(uint32_t domain, uint32_t cid, const void* ca
     // Pack callbacks into row structures
 
     thread_local std::unordered_map<activity_correlation_id_t, timespec> timestamps;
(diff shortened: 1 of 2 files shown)


xw285cornell (approved):
lgtm - so the thread local map isn’t the reason we seg fault right?

houseroad (approved):
Thanks for fixing the issue. Better to add some unittest to cover the logic :-)

aaronenyeshi:
> lgtm - so the thread local map isn’t the reason we seg fault right?

Yes that's right, this shared map `dis->externalCorrelations_` seemed to be the cause of the seg fault.

PR #906, opened 2024-04-15
update process group ID to process group name

shengbao-zheng:
Summary:
deprecate process group id field with process group name
now ncclkernel displays ProcessGroupName, desc and collectiveName

Differential Revision: D55612967


aaronenyeshi (approved):
LGTM, see internal review. Make sure strings are quoted.

PR #907, opened 2024-04-15
Remove all instances of TMP_LIBKINETO_NANOSECOND from Kineto

sraikund16:
Summary: Now that we have ns granularity working in main we can remove the temporary guards that were needed to make sure that Kineto was in sync with pytorch. Remove all the guards and follow up with a diff where we remove the compiler flags

Differential Revision: D56147637


aaronenyeshi (approved):
Great work!

PR #909, opened 2024-04-18
remove hasPrivateUse1Type function

fwenguang:
Fix issue in https://github.com/pytorch/pytorch/pull/120556#issuecomment-2061557740.


fwenguang:
Hi @aaronenyeshi , to fix https://github.com/pytorch/pytorch/pull/120556#issuecomment-2061557740, I have already deleted the hasPrivateUse1Type function. As for privateuse1, writing GPU info into its JSON file won't cause any issues, it just won't be utilized. In the future, we can try to find a new way to clearly write the information for each device.

aaronenyeshi (approved):
LGTM!

PR #910, opened 2024-04-18
use relative time to ensure the 'ts' field can be accurately read

fwenguang:
Fix https://github.com/pytorch/kineto/pull/890#discussion_r1530009921.


fwenguang:
Hi @aaronenyeshi , i just use the program loading time as the baseline. What is your suggestions about this?

aaronenyeshi (approved):
This is great! I think static loading time as base_time is fine, since all timestamps should be valid.

aaronenyeshi:
Need to land some changes to unit tests before landing this. See: https://github.com/pytorch/kineto/pull/911.

aaronenyeshi:
Couple of internal test giving the error:
```
kineto/libkineto/src/output_json.cpp:58:16: error: declaration requires a global constructor [-Werror,-Wglobal-constructors]
static int64_t base_time =
               ^
1 error generated.
```

aaronenyeshi:
There is an internal test failure:
```
INFO:2024-04-19 08:38:59 6655:6655 output_json.cpp:510] Chrome Trace written to /tmp/libkineto_testedEuBM.json
INFO:2024-04-19 08:38:59 6655:6655 output_json.cpp:564] Renamed the trace file to /tmp/libkineto_testedEuBM.json
terminate called after throwing an instance of 'folly::json::parse_error'
  what():  json parse error on line 8 near `-660, "dur": 0.0': expected '}'
*** Aborted at 1713541139 (Unix time, try 'date -d @1713541139') ***
*** Signal 6 (SIGABRT) (0x7559000019ff) received by PID 6655 (pthread TID 0x7f87b978e000) (linux TID 6655) (maybe from PID 6655, UID 30041) (code: -6), stack trace: ***
==6655==WARNING: ASan doesn't fully support makecontext/swapcontext functions and may produce false positives in some cases!
    @ 0000000000272c67 (unknown)
    @ 00000000002713e1 (unknown)
    @ 00000000000445bf (unknown)
                       /home/engshare/third-party2/glibc/2.34/src/glibc-2.34/signal/../sysdeps/unix/sysv/linux/libc_sigaction.c:8
                       -> /home/engshare/third-party2/glibc/2.34/src/glibc-2.34/signal/../sysdeps/unix/sysv/linux/x86_64/libc_sigaction.c
    @ 000000000009c9f3 __GI___pthread_kill
                       /home/engshare/third-party2/glibc/2.34/src/glibc-2.34/nptl/pthread_kill.c:46
```

I've changed all instances of `start_time_ns` in `CuptiActivityProfilerTest.cpp` to use `system_clock::now()`. Could this be some difference between system_clock and high_resolution_clock?
```
int64_t start_time_ns = libkineto::timeSinceEpoch(std::chrono::system_clock::now());
```

libkineto/src/output_json.h
```diff
@@ -91,4 +92,17 @@ class ChromeTraceLogger : public libkineto::ActivityLogger {
...
 };
 
+class ChromeTraceBaseTime {
+public:
+  ChromeTraceBaseTime() = default;
+  static void init() {
+    get();
+  }
+  static int64_t& get() {
+    static int64_t base_time = libkineto::timeSinceEpoch(
```
aaronenyeshi, line 102: This is triggering a CLANGTIDY warning:
```
Inlined static variable warning.
Local variable base_time with global storage is defined in inline function get.  Since this function is inlined, multiple copies of this variable may be used at runtime (due to dynamic linking) even though it has global storage.
Found local variables with global storage in inline functions which may
cause multiple copies to appear at runtime when used in dev mode (due
to dynamic linking and using -fvisibility-inlines-hidden). 
```

aaronenyeshi: Fixed this by making the class a static singleton

aaronenyeshi:
worked around the internal test failure by checking that `transToRelativeTime` always returns a 0 or positive int.

Issue #838, opened 2023-11-27
Don't hardcode /opt/rocm in CMakeLists.txt


suranap:
Most HPC systems use [Environment Modules](https://github.com/cea-hpc/modules) to load libraries. Therefore, the ROCm libraries are loaded explicitly by version: `module load rocm/5.6.0`. On the systems I've used ROCM_PATH is defined as `/opt/rocm-5.6.0`. I'd suggest changing these lines to `${ROCM_SOURCE _DIR}/lib`. 

https://github.com/pytorch/kineto/blob/490f305345b2c06169b02e98418645f3def47b95/libkineto/CMakeLists.txt#L193-L195

sraikund16:
Fixed in  #912

PR #912, opened 2024-04-18
Remove Hardcoding of opt/rocm in CMakeLists.txt

sraikund16:
Summary: In https://github.com/pytorch/kineto/issues/838, suranap points out that we should not be using hardcoded paths for the rocm source

Differential Revision: D56319191


aaronenyeshi (approved):

aaronenyeshi:
cc @mwootton @jeffdaily

PR #914, opened 2024-04-21
Fix GPU User Annotations overlapping with associated GPU ops

aaronenyeshi:
Summary: Since we transitioned the profiler events to store ns timestamps, there are overlapping gpu_user_annotation and gpu kernel events. The overlap was previously fixed by making the gpu_user_annotations start slightly earlier than associated kernels, but it seems that 1ns is not enough for Chrome Trace Viewer. Switch it to 1000ns, which was 1us equivalent to before.

Differential Revision: D56401601

Pulled By: aaronenyeshi


fwenguang:
After landing this patch: https://github.com/pytorch/kineto/pull/910,will this issue be fixed? @aaronenyeshi

briancoutinho (approved):
lgtm :)

PR #917, opened 2024-04-24
Add TSC Clock Callback to CUPTI

sraikund16:
Summary:
Right now we use the default clock for CUPTI which is not monotonic nor particularly fast. To fix this, lets define a getAppoximateTime call for Kineto and use that to retrieve timestamp counts. Afterwards we can have profiler set the converter from timestamp count to timestamp. We add this converter to the timestamp() and duration() functions in Kineto.

One important detail is that this is not to be used for Windows or CUDA versions before 11.6. For this reason, necessary ifdefs have been placed. Profiler follow up diff will enable the features in this diff.

Reviewed By: aaronenyeshi

Differential Revision: D56525885


aaronenyeshi:
Looks already landed in https://github.com/pytorch/kineto/commit/e4a2f1750ca28acc0edb83265cc6859326eccecd

PR #918, opened 2024-04-26
Make ipcfabric have a single definition

briancoutinho:
Summary:
There is an ODR one definition rule violation that was causing a crash on sigrid
https://fb.workplace.com/groups/560979627394613/posts/2909061125919773/?comment_id=2909752389183980&reply_comment_id=2910102119149007
Sigrid includes both kineto and ipcfabric via dynolog, and on kineto the class is fused in internal. Also adding Logger.h caused a mess.

Tries to resolve this issue
Note ENABLE_IPC_FABRIC define is only used in open source kineto (and PyTorch) but not in internal fbcode

Differential Revision: D56577485

### DIFF
--- a/libkineto/src/IpcFabricConfigClient.cpp
+++ b/libkineto/src/IpcFabricConfigClient.cpp
@@ -9,7 +9,6 @@
 #ifdef __linux__
 
 #include "IpcFabricConfigClient.h"
-#include "Logger.h"
 
 #include <random>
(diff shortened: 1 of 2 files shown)


aaronenyeshi:
Looks already landed in https://github.com/pytorch/kineto/commit/117ab8c80a93b6536c6ff6bcd446b220940427a3

PR #919, opened 2024-04-26
Change relative time of 'ts' field from init time to the start of this year

aaronenyeshi:
Summary: Use the start of this year as the base_time which all the timestamps will be relative to. This is better than using init time of the tracer because traces across ranks are starting at different starting points, and its very hard to correlate them especially when working on PP, FSDP, etc.

Differential Revision: D56642236

Pulled By: aaronenyeshi

### DIFF
--- a/libkineto/src/output_json.h
+++ b/libkineto/src/output_json.h
@@ -8,9 +8,11 @@
 
 #pragma once
 
+#include <chrono>
 #include <fstream>
 #include <map>
 #include <ostream>
+#include <ratio>
 #include <thread>
 #include <unordered_map>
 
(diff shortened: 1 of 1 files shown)


aaronenyeshi:
cc @fwenguang , there is an issue where certain tools depend on the actual time, so we need to make sure all the traces across ranks are aligned to a relative timestamp. So I'm setting base_time to jan 1 of current year.

aaronenyeshi:
Even with under 1 year of significant digits, I'm getting overlapping. The timestamps are around 17 digits in str.
image

fwenguang:
17 digits will make double overflow, is overlapping caused by it? 

How about recording the base_time into json file, will this solve the tools issue?

aaronenyeshi:
> 17 digits will make double overflow, is overlapping caused by it?
> 
> How about recording the base_time into json file, will this solve the tools issue?

That's a good point, but I'm not sure how widespread the impact to tooling is. A lot of tools may be making assumptions. I've done some calculations, and given the double has 53 bit mantissa, we can keep all significant digits if we round base_time to the nearest 3 month interval. Tested that this works.

Rounding base_time to yearly intervals:
image
Rounding base_time to 3 month intervals:
image

aaronenyeshi:
Landed in https://github.com/pytorch/kineto/commit/8bf50de29b08a3f51d9f1486268e43396faa0915

PR #921, opened 2024-04-30
Make Kineto Stage Log Level Lower

sraikund16:
Summary:
Users have complained that STAGE level logs are print by default. If a user is running many profiles it can certainly clutter STDOUT. Lower the STAGE level such that it is below the default.

Brought up by this issue https://github.com/pytorch/pytorch/issues/83383

Differential Revision: D56646009


aaronenyeshi (approved):
LGTM

aaronenyeshi:
Landed in [D56646009](https://www.internalfb.com/diff/D56646009)

PR #926, opened 2024-05-07
Add sortIndex to fix ordering of roctracer GPU id tracks

aaronenyeshi:
Summary:
Although hipGetDeviceProperties shows 8 devices enumerating from 0 to 7, when using roctracer_record_t, the record->device_id enumerates from 2 to 9. This is because roctracer considers id 0 and 1 as CPU sockets.

So we can expect up to 4 CPU sockets, and 8 GPUs, up to a total count of 12. Which would mean the enumeration will be 0 - 11. This will help put the GPU tracks in order and at the bottom of the chrome trace.

Original Bug report in ROCm/roctracer: https://github.com/ROCm/roctracer/issues/98. We can just show the device id that is given by roctracer.

Differential Revision: D57067065

Pulled By: aaronenyeshi

## Test Plan:
Before:
image

After:
image

### DIFF
--- a/libkineto/include/IActivityProfiler.h
+++ b/libkineto/include/IActivityProfiler.h
@@ -40,12 +40,19 @@ enum class TraceStatus {
 };
 
 /* DeviceInfo:
- *   Can be used to specify process name, PID and device label
+ *   Can be used to specify process name, sort order, PID and device label.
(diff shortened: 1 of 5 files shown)


houseroad (commented):
commented internally

PR #950, opened 2024-06-13
Add RoctracerActivityProfilerTest Unit Tests

aaronenyeshi:
Summary:
Adding 5 unit tests for the CuptiActivityProfiler and its interactions with RoctracerActivityApi. Mock the RoctracerLogger and the RoctracerActivityApi and pass them into the CuptiActivityProfiler orchestrator, then verify the trace files produced.

- **SyncTrace:** Mocks 5 CPU ops, 3 CPU Kernel Runtime Launches, 2 CPU Memcpy Runtime events, 3 GPU Kernels, 2 GPU Memcpy events. Checks that the trace output matches what was passed in, and names are correct after parsing.
- **GpuNCCLCollectiveTest:** Similar to CUPTI, check that NCCL metadata is properly passed into the CPU and GPU ops.
- **GpuUserAnnotationTest:** Check that GPU user annotations added via CorrelationDomain1 works properly, and the annotations are in the final trace.
- **SubActivityProfilers:** Check subactivityprofiler children will continue to work, such as glow runtime mock activity profiler.
- **JsonGPUIDSortTest:** Check that JSON file contains expected number of process_labels and process_sort_index , so that GPU rows in Chrome traces will always be sorted after CPU rows.

Differential Revision: D58554825

Pulled By: aaronenyeshi

### DIFF
--- a/libkineto/src/RoctracerActivityApi.h
+++ b/libkineto/src/RoctracerActivityApi.h
@@ -53,7 +53,7 @@ class RoctracerActivityApi {
   void clearActivities();
   void teardownContext() {}
 
-  int processActivities(
+  virtual int processActivities(
     std::function<void(const roctracerBase*)> handler,
     std::function<void(uint64_t, uint64_t, RoctracerLogger::CorrelationDomain)> correlationHandler);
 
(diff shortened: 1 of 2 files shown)


aaronenyeshi:
This may fail on NVIDIA A10G environment, I need to add some guarding.

houseroad (approved):
LGTM, thanks!

Issue #885, opened 2024-03-04
How to add customized metadata with on demand profiling ? 


staugust:
When profiling with `torch.profiler.profile` ,  generated json file has a section called `distributedInfo` shown as below
```json
{
  "distributedInfo": {"backend": "nccl", "rank": 0, "world_size": 2}
}
```
But there's no such section in generated file when on-demand profiling is triggered.  As a result, Holistic Trace Analysis cannot be used to analysis those files. 
Is this by design or there's something to do to make those file generated by `kineto` have `distributedInfo` as well? Hoping some one can help. Thanks.

aaronenyeshi:
This is a bug, if you would like, please submit a fix for this issue. cc @briancoutinho

staugust:
> This is a bug, if you would like, please submit a fix for this issue. cc @briancoutinho

Thanks for kindly response,  I will give it a try in few days.

sraikund16:
@staugust do you still plan on working on this?

staugust:
@sraikund16   I'm afraid I don't have enough time to  work on this in next one or two weeks.  It would be great if you have time to solve it, otherwise,  I would work on this later.

sraikund16:
@staugust I am going to take up this task, thanks

sraikund16:
reopening because needs follow up

sraikund16:
Closing for now as this was added for NCCL traces in https://github.com/facebookresearch/HolisticTraceAnalysis/pull/157

If there is a need for gloo/mpi backends we can reopen and do a follow up

PR #956, opened 2024-06-27
Add Dist Info for On-Demand NCCL Traces

sraikund16:
Summary:
Currently, auto-trace uses pytorches c10d.distributed library to set up the "distributedInfo" field in the chrome tracing. This makes sense as all the process groups are initialized and maintained in python. Because of this, however, on-demand is unable to retrieve this information directly as it has no access to the python side of a graph. However, we can glean the process group information based on the information that is added in to the dispatcher.

In this diff, we keep a global to enumerate all the nccl information via the args in the collectives as we print them out to the trace. At the end of the trace we output the distributed information. In order to prevent double counting for auto-trace, we skip if there was previously distributedInfo outputted to the trace.

Fixes https://github.com/pytorch/kineto/issues/885

Differential Revision: D58736045

### DIFF
--- a/libkineto/src/output_json.cpp
+++ b/libkineto/src/output_json.cpp
@@ -12,7 +12,6 @@
 #include <fstream>
 #include <time.h>
 #include <map>
-
 #include "Config.h"
 #include "DeviceProperties.h"
 #include "TraceSpan.h"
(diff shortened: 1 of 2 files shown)


aaronenyeshi (approved):
LGTM!

PR #965, opened 2024-07-26
fix static variable client access vialotion in DamonConfigLoader 

staugust:
As discussed in pytorch issue [#129626](https://github.com/pytorch/pytorch/issues/129626) , `updateThread` of `Config` will keep on accessing `client` after main thread quits.  But when main thread quits,  `client` will be destructed.  Functions in `updateThread` may use a dangling pointer of `client`, which will cause invalid memory access, and finally, `Segment fault` occurs, a core file will be generated, which  doesn't behave as expected.


aaronenyeshi (approved):
LGTM!

PR #967, opened 2024-07-26
Setters and getters for trace IDs

sanrise:
Summary:
This commit moves the trace ID initialization logic to be inline with how ActivityProfilers interact with LoggerObservers by setting a default local trace ID when a profiler config does not contain a trace ID.

The USTLoggerCollector, which is our internal LoggerObserver will record the trace ID for a given environment (which is a PID today) using the `setTraceID` method that all LoggerObservers must staisfy. Additionally any internal calls to read this trace ID, for example from our ManifoldChromeTrace logger, may use a special `getTraceID` method that ships with `USTLoggerCollector`.

Group trace IDs are handled accordingly.

Differential Revision: D60267172


sraikund16 (approved):
LGTM

aaronenyeshi (approved):
I forget if this diff was refactored somewhere, accepting to unblock.

PR #968, opened 2024-07-26
Add Logging for Empty Traces

sraikund16:
Summary: Recently, we have had users seen empty traces when the system is idle leading to confusion as to whether it was caused by a bug in kineto formatting or not. This diff adds further logging to ensure that the user is aware that kineto finds no valid trace events and is expecting an empty trace.

Differential Revision: D60311331

### DIFF
--- a/libkineto/src/CuptiActivityProfiler.cpp
+++ b/libkineto/src/CuptiActivityProfiler.cpp
@@ -275,6 +275,8 @@ void CuptiActivityProfiler::processTraceInternal(ActivityLogger& logger) {
   VLOG(0) << "Profile time range: " << captureWindowStartTime_ << " - "
           << captureWindowEndTime_;
   logger.handleTraceStart(metadata_);
+  setCpuActivityPresent(false);
+  setGpuActivityPresent(false);
   for (auto& cpu_trace : traceBuffers_->cpu) {
     string trace_name = cpu_trace->span.name;
     VLOG(0) << "Processing CPU buffer for " << trace_name << " ("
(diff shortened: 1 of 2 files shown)


aaronenyeshi (changes requested):
Should we also add the functions into roctracer's handleactivity functions?

sraikund16:
> Should we also add the functions into roctracer's handleactivity functions?

@aaronenyeshi Sure, can we do that as a followup diff instead?

aaronenyeshi:
> > Should we also add the functions into roctracer's handleactivity functions?
> 
> @aaronenyeshi Sure, can we do that as a followup diff instead?

yes that sounds good

libkineto/src/CuptiActivityProfiler.cpp
```diff
@@ -326,6 +331,9 @@ void CuptiActivityProfiler::processTraceInternal(ActivityLogger& logger) {
     LOGGER_OBSERVER_ADD_EVENT_COUNT(count);
   }
 #endif // HAS_ROCTRACER
+  if (!traceNonEmpty()) {
+    LOG(WARNING) << "No Valid Trace Events found. Outputting empty trace.";
```
aaronenyeshi, line 335: Could we clarify this means no CPU and GPU events?

aaronenyeshi (approved):
LGTM, small nit

sraikund16:
> > > Should we also add the functions into roctracer's handleactivity functions?
> > 
> > 
> > @aaronenyeshi Sure, can we do that as a followup diff instead?
> 
> yes that sounds good

Looking at it now, I think this already covers the roctracer impl. handleGpuActivity with ItraceActivity as an arg is shared between CUPTI and Rocm, and the Rocm handleRuntimeActivity is also covered in this diff. We will see for sure when the next diff tests on all architectures though

PR #971, opened 2024-07-30
Fix Kineto Stress Test

sraikund16:
Summary:
The Kineto Stress test was failing for two reasons:

1) In the test itself, we did not set the converter for the TSC timestamp, making all values out of range

2) In Kineto itself there was a bug regarding child profilers. When these profilers are set their spans are set to all 0. However, when we set the record start in Kineto we use the start span of the CPU trace to get the start of the trace itself. In the future we should probably decouple these things but for now I added an edge case to ensure that the profile is not malformed.

Reviewed By: aaronenyeshi

Differential Revision: D60415474


aaronenyeshi (approved):
LGTM!

PR #972, opened 2024-08-01
Add API for Dynamic Activity Toggling [1/n]

sraikund16:
Summary: During PT2 there are many GPU/CPU events that are unneccessary to profile in between a given step. To remedy this, we can add an API that takes in a list of activities and an arg to toggle said activies or not. For this diff we are just adding the Kineto side to turn on/off the "CUDA" events which includes AMD/Roctracer. A follow up will be added for the generic profiler side.  Subsequent diffs will be added for CPU toggling and e2e testing.

Differential Revision: D60542040


aaronenyeshi (approved):
LGTM!

PR #978, opened 2024-08-16
Correct the condition when setting External id

fwenguang:
The `External id` will be set to op.linkedActivity()->correlationId() if op.linkedActivity() is not nullptr. So in `if` condition we should check if op.linkedActivity()->correlationId() is 0.


fwenguang:
Hi  @aaronenyeshi , please help to review this pr, thank you.

aaronenyeshi (approved):
LGTM!

PR #986, opened 2024-09-12
Extend CPU User Annotations to End of Profile

sraikund16:
Summary: If a CPU User Annotation doesn't end by the time the profile ends, the annotation is marked as a 0-length event. This can be annoying to look at because it seems like profiler just never got the annotation event when it did. Lets set the end time to the end of profiling.

Differential Revision: D62604717


aaronenyeshi (approved):
LGTM, and makes sense!

PR #992, opened 2024-09-23
Add 1ns buffer to Roctracer Events

sraikund16:
Summary: As reported in https://github.com/ROCm/roctracer/issues/105, there is an issue where event starts and ends can "tie". This can cause a visual issue in the traces. Lets add a tiny buffer so the events are separate. At the single nanosecond level, the timings are inaccurate anyways so it doesn't really hurt to add this buffer in the meanwhile. Remove/wrap in ifdef once it is issue is resolved

Differential Revision: D63296093

### DIFF
--- a/libkineto/src/RoctracerActivity.h
+++ b/libkineto/src/RoctracerActivity.h
@@ -101,6 +101,15 @@ struct GpuActivity : public RoctracerActivity<roctracerAsyncRow> {
   void log(ActivityLogger& logger) const override;
   const std::string metadataJson() const override;
 
+  // Add small buffer to fix visual error created by https://github.com/ROCm/roctracer/issues/105
+  // Once this is resolved we can use ifdef to handle having this buffer or not based on version
+  int64_t timestamp() const override {
+    return activity_.begin + 1;
+  }
+  int64_t duration() const override {
+    return activity_.end - (activity_.begin + 1);
+  }
+
  private:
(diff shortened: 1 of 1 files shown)


libkineto/src/RoctracerActivity.h
```diff
@@ -101,6 +101,15 @@ struct GpuActivity : public RoctracerActivity<roctracerAsyncRow> {
   void log(ActivityLogger& logger) const override;
   const std::string metadataJson() const override;
 
+  // Add small buffer to fix visual error created by https://github.com/ROCm/roctracer/issues/105
+  // Once this is resolved we can use ifdef to handle having this buffer or not based on version
+  int64_t timestamp() const override {
+    return activity_.begin + 1;
+  }
+  int64_t duration() const override {
+    return activity_.end - (activity_.begin + 1);
```
aaronenyeshi, line 110: should it be -2 , so that it finishes right before? Or is it okay if end time == parent end time?

aaronenyeshi (approved):
LGTM!

libkineto/src/RoctracerActivity.h
sraikund16: So we are only shaving off a nanosecond off the beginning since this is on the gpu track only so there won't be any parent/child minus the annotations which isn't a big deal since we set those ourselves.

PR #993, opened 2024-09-25
Fix Bandwidth Calculations in JSON

sraikund16:
Summary:
We are using TSC timestamps as durations instead of the conversion to seconds. We should never call the start/end bare, only use the duration() and timstamp() functions

Differential Revision: D63422014


davidberard98 (approved):

sanrise (approved):
Lgtm, thanks!

aaronenyeshi (approved):

PR #995, opened 2024-10-03
Remove Stream from HIP Launches

sraikund16:
Summary: We would like for AMD traces to match CUDA matches as closely as possible. Right now we are adding stream to the CPU launches whereas in CUDA we do not. This will make zoomer try to process these events when they shouldn't. To make matters worse, the stream id that is embedded in the cpu event don't even match the GPU event. For this reason, lets get rid of them all together.

Differential Revision: D63798661


aaronenyeshi (approved):
sounds good to me

PR #1001, opened 2024-10-12
Fix the issue of incorrect External id for non-PyTorch activities

fwenguang:
Sorry about the incomplete patch addressing the issue of setting the external ID.
I think only the PyTorch activities should set the external id to op.correlationId(), while the external ids for the runtimes and kernels should be set to op.linkedActivity()->correlationId().
However, there may be cases where the linkedActivity() for the runtime or kernel is nullptr (for example, when directly launching a kernel using a pybind mapping interface). In such cases, the external id should be empty, not set to op.correlationId().

### DIFF
--- a/libkineto/src/output_json.cpp
+++ b/libkineto/src/output_json.cpp
@@ -351,11 +351,22 @@ void ChromeTraceLogger::handleActivity(const libkineto::ITraceActivity& op) {
   int external_id = 0;
   if (op.linkedActivity()) {
     external_id = op.linkedActivity()->correlationId();
-  } else if (op.type() == libkineto::ActivityType::CPU_OP ||
-      op.type() == libkineto::ActivityType::CPU_INSTANT_EVENT ||
-      op.type() == libkineto::ActivityType::USER_ANNOTATION ||
-      op.type() == libkineto::ActivityType::PYTHON_FUNCTION) {
-    external_id = op.correlationId();
+  } else {
+    // Some runtime events and kernels may not have a linked activity,
+    // should not set an "External id" for them. Otherwise, these events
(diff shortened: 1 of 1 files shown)


libkineto/src/output_json.cpp
```diff
@@ -348,12 +348,18 @@ void ChromeTraceLogger::handleActivity(const libkineto::ITraceActivity& op) {
     duration += 2; // Still need it to end at the original point rounded up.
   }
 
+  int external_id = 0;
+  if (op.linkedActivity()) {
+    external_id = op.linkedActivity()->correlationId();
+  } else if (op.type() == libkineto::ActivityType::CPU_OP ||
```
fwenguang, line 354: These types align with https://github.com/pytorch/pytorch/blob/main/torch/csrc/profiler/collection.cpp#L1023-L1026

fwenguang:
Hi, @aaronenyeshi , could you please take a look at whether this patch is okay?

libkineto/src/output_json.cpp
sraikund16: I think logically this makes sense although I was wondering if instead of having this whitelist of events that set external_id = op.correlationId(), can we just have a blacklist of events where we want to skip (ie. CUDA_RUNTIME/CUDA_KERNEL events). I ask this because there could be a type that we are missing and I would rather have extra information (even if wrong) rather than have a trace missing information.

sraikund16:
Overall looks reasonable

libkineto/src/output_json.cpp
sraikund16: Awesome, thanks!

sraikund16 (approved):

PR #1008, opened 2024-10-31
Support child profilers adding their deviceProperties to logger

fwenguang:
This patch is intended to support our out-of-tree Kineto plugin in adding its deviceProperties.

### DIFF
--- a/libkineto/src/CuptiActivityProfiler.cpp
+++ b/libkineto/src/CuptiActivityProfiler.cpp
@@ -8,6 +8,7 @@
 
 #include "CuptiActivityProfiler.h"
 #include <fmt/format.h>
+#include <fmt/ranges.h>
 #include <time.h>
 #include <atomic>
 #include <cstdint>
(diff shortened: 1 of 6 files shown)


fwenguang:
Hi @sraikund16 , could you help review this?

sraikund16:
This is a cool idea, thanks for submitting this. I think this will affect our internal testing so I'll try to dedicate some time to make sure things work this week.

libkineto/src/output_membuf.h
```diff
@@ -66,6 +66,13 @@ class MemoryTraceLogger : public ActivityLogger {
     metadata_ = metadata;
   }
 
+  void handleTraceStart(
```
sraikund16, line 69: Same as comment above

libkineto/src/CuptiActivityProfiler.cpp
```diff
@@ -306,7 +308,17 @@ void CuptiActivityProfiler::processTraceInternal(ActivityLogger& logger) {
   for (auto& pair : versionMetadata_) {
     addMetadata(pair.first, pair.second);
   }
-  logger.handleTraceStart(metadata_);
+  std::vector<std::string> device_properties;
+  if (auto props = devicePropertiesJson(); !props.empty()) {
```
sraikund16, line 312: nit: can we define these vars outside of the if statement? I think our internal linter will complain about lines like these

aaronenyeshi (approved):

PR #1010, opened 2024-11-05
Add Bandwidth Calculations to RoCM Traces

sraikund16:
Summary:
Takes the hip launch for memset and memcpy metadata and forwards it to kernel events. Afterwards use duration to calcaulate bandwidth.

Also, make sure global vars are thread safe and anonymous

Differential Revision: D65452222


aaronenyeshi (approved):
LGTM, please check linter msgs

sraikund16:
> LGTM, please check linter msgs

its failing because of some lint on a file that this PR doesn't touch. I will probably bypass and use a forward fix

PR #1014, opened 2024-11-14
Record comms input and output tensor information

sanrise:
Summary: Just copy the output from NCCL metadata about tensor information. Allows for easier analysis of kernel memory access patterns in downstream tools.

Differential Revision: D65785010


sraikund16 (approved):
accepting to unblock, lets make sure to add a trace though

PR #1015, opened 2024-11-20
Disable NvPerfMetric for CUDA 12.6.0

sraikund16:
Summary: 12.6.0 changes all the headers and APIs used in this file so we need to disable it for now

Differential Revision: D66265298


libkineto/src/CuptiNvPerfMetric.cpp
```diff
@@ -46,7 +46,7 @@ namespace nvperf {
 // After CUDA RT 11.04, the structure has changed.
 // TODO update the structure NVPA_RawMetricsConfig to support 11.04
 #if defined(USE_CUPTI_RANGE_PROFILER) && defined(CUDART_VERSION) && \
-    CUDART_VERSION > 10000
+    CUDART_VERSION < 12060
```
briancoutinho, line 49: oh why not > and  < here?

briancoutinho (approved):
lgtm :) minor comment

PR #1017, opened 2024-11-25
Undo revert of the PR #1014

sanrise:
Summary:
Reverts a rollback D66458621

Revert the Kineto rollback, this would have partially solved the issue since this part controls transmission of the metadata to the corresponding kernel. but record_param_comms in pytorch is the real issue and was still recording this metadata and would still make an invalid trace JSON when working with GPUs>30 (our truncation case).

Differential Revision: D66475394


sraikund16 (approved):
sgtm lets just wait until the escape quote fix goes in

PR #1023, opened 2024-12-20
Add Extra Overhead Cases

sraikund16:
Summary: CUPTI has added extra cases for overhead events. We should add them to kineto so these events are not just set as unknown

Differential Revision: D67539465


davidberard98 (approved):

sanrise (approved):
LGTM. Will improve the overall trace quality.

aaronenyeshi (approved):

PR #1027, opened 2025-01-08
Omit Non Human Readable Names in Kineto

sraikund16:
Summary: When running with_stack there are instances where the JSON output is configured improperly because certain values are non-human readable. We should at least patch this by omitting such values until we find a long-term solution.

Reviewed By: aaronenyeshi, mkyybx

Differential Revision: D67923913


aaronenyeshi (approved):
LGTM!

PR #1031, opened 2025-01-29
Add Metadata for Annotation

sraikund16:
Summary: We recently extended annotations for on-demand to the end of a profile if the annotation does not finish by then. This is useful to know where the tracing ended. However, it is hard to tell when looking to see if the annotation ended as the trace ended or if the annotation ended because of the tracing ended. We add this extra metadata to deliniate this

Differential Revision: D68864115


sanrise (approved):
LGTM

PR #1035, opened 2025-02-07
Enable init for daemon cases, remove event profiler code

### LINKED ISSUE #1032: Provide a way to initialize connection to dynolog without triggering cupti API
Context in https://github.com/facebookincubator/dynolog/issues/286. 

In pytorch, we have only 2 call sites of libkineto_init (https://github.com/search?q=repo%3Apytorch%2Fpytorch%20libkineto_init&type=code). - One is the to when we need to register ourselves to dynolog ([KINETO_USE_DAEMON=1](https://github.com/pytorch/pytorch/blob/6371c25b91fb66756b5a2df67741d4e7a72a7261/torch/csrc/profiler/kineto_client_interface.cpp#L78)). 
- One is for the programmable torch.profiler interface (https://github.com/pytorch/pytorch/blob/6371c25b91fb66756b5a2df67741d4e7a72a7261/torch/csrc/profiler/kineto_shim.cpp#L247).

First one is triggered from `global_kineto_init`, but `libkineto_init` actually triggers a bunch of cupti API, which defeats the purpose of lazy cupti attach. Therefore we are potentially affecting the perf of the program until cuptiFinalize is called. 

I think the right design would be decoupling registration to dynolog from all the cupti initialization. At `global_kineto_init`, we only register to dynolog but don't do cupti calls. And then when we get signal from dynolog, we trigger ondemand profile and then attach cupti context, basically following the pattern of the second use case in pytorch above. 

This requires
- Some work to separate registration to dynolog from all the cupti initialization in libkineto
- Provide API in libkineto so that `global_kineto_init` in pytorch only trigger the registration to dynolog part
- Change pytorch code `global_kineto_init` accordingly.

briancoutinho:
Summary:
Fix #1032 
Cleans up the initialization for kineto
* The current method to init kineto for CUDA builds is to add a callback on CUDA context. But this leads to CUPTI being enabled right from the start.
* For the case where profiling daemon is enabled (dynolog), this PR initialized the profiler and config loader always, for both CPU and CUDA builds. This should be safe to do as kineto_init now happens when torch is imported.
* Still leaving the CUPTI callback approach above for non dynolog/daemon use cases, this leave behavior inside Meta unchanged.
* Let's start cleaning up Even profiler. From here it is offiically turned off. Will start nuking the files soon.

Differential Revision: D69285243

## Test

Clone change in pytorch repo and build it.

Run dynolog in one shell
```
  ../dynolog/usr/local/bin/dynolog --enable_ipc_monitor
 ```
 Run test python program
```
export KINETO_USE_DAEMON=1
python3 third_party/kineto/libkineto/third_party/dynolog/scripts/pytorch/linear_model_example.py
```
 
 Without this change we daemon loader registry but no connection with dynolog
 ```
 INFO:2025-02-07 00:42:44 2893200:2893200 init.cpp:136] Registering daemon config loader, cpuOnly =  0
cuda:0
99 170.32142639160156
10099 8.817169189453125
 ```

With the change we see dynolog getting connected to 
```
INFO:2025-02-07 00:48:52 2955842:2955842 init.cpp:140] Registering daemon config loader, cpuOnly =  0
INFO:2025-02-07 00:48:52 2955842:2955842 CuptiActivityProfiler.cpp:264] CUDA versions. CUPTI: 18; Runtime: 12000; Driver: 12000
INFO:2025-02-07 00:48:52 2955842:2955842 DaemonConfigLoader.cpp:66] Setting communication fabric enabled = 1
INFO:2025-02-07 00:48:52 2955842:2955842 IpcFabricConfigClient.cpp:88] Setting up IPC Fabric at endpoint: dynoconfigclientfa3396c8-b168-4802-bd20-264cfbe514d2 status = initialized
INFO:2025-02-07 00:48:52 2955842:2955842 DaemonConfigLoader.cpp:66] Setting communication fabric enabled = 1
INFO:2025-02-07 00:48:52 2955842:2955842 DaemonConfigLoader.cpp:66] Setting communication fabric enabled = 1
```

We also find prints in dynolog shell 
```
I20250207 00:48:57.009657 2955379 LibkinetoConfigManager.cpp:171] Registered process (2955842, 3411461, 2540747) for job 0.
I20250207 00:48:57.009698 2955379 IPCMonitor.cpp:84] getLibkinetoOnDemandRequest() : job id 0 pids = 2955842
```

Last, we can now trigger and collect trace, use a third shell
```
$> ../dynolog/usr/local/bin/dyno gputrace --log-file /tmp/test_trace.json
...
response length = 147
response = {"activityProfilersBusy":0,"activityProfilersTriggered":[2955842],"eventProfilersBusy":0,"eventProfilersTriggered":[],"processesMatched":[2955842]}
Matched 1 processes
```

### DIFF
--- a/libkineto/src/init.cpp
+++ b/libkineto/src/init.cpp
@@ -36,9 +36,10 @@ namespace KINETO_NAMESPACE {
 #if __linux__ || defined(HAS_CUPTI)
 static bool initialized = false;
 
-static void initProfilersCPU() {
+static void initProfilers() {
   if (!initialized) {
     libkineto::api().initProfilerIfRegistered();
+    libkineto::api().configLoader().initBaseConfig();
     initialized = true;
     VLOG(0) << "libkineto profilers activated";
   }
@@ -47,11 +48,6 @@ static void initProfilersCPU() {
 #endif // __linux__ || defined(HAS_CUPTI)
 
 #ifdef HAS_CUPTI
-static std::mutex& initEventMutex() {
-  static std::mutex initMutex_;
-  return initMutex_;
-}
-
 bool enableEventProfiler() {
   if (getenv("KINETO_ENABLE_EVENT_PROFILER") != nullptr) {
     return true;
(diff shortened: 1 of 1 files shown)


sanrise (approved):
approving on OSS to unblock. added a minor question on the internal diff [D69285243](https://www.internalfb.com/diff/D69285243).

PR #1053, opened 2025-03-03
Fix bugs in kineto sample programs.

Conless:
Thank you for maintaining Kineto and for all the work that goes into making these tools available! I’ve been using it recently and find some bugs in the sample programs. This PR aims to fix them. Specifically, 
- Fix incorrect function call: In libkineto/sample_programs/kineto_playground.cu (line 79), the example incorrectly calls `basicMemcpyFromDevice` instead of `basicMemcpyToDevice`, which may be a typo.
https://github.com/pytorch/kineto/blob/26661c0c7620267f09ac54a2900e55f08cb0799e/libkineto/sample_programs/kineto_playground.cu#L77-L82
- Add `CHECK_CUDA(cudaDeviceSynchronize())` to the main function to make similar potential runtime errors more clearly, preventing silent failures.
- Add a call to `libkineto_init` in the main function and include the necessary libraries in the build script, to ensure the example runs out of the box.

### DIFF
--- a/libkineto/sample_programs/build.sh
+++ b/libkineto/sample_programs/build.sh
@@ -18,5 +18,7 @@ g++ \
   -lpthread \
   -lcuda \
   -lcudart \
+  -lcupti \
+  -lnvperf_host \
   /usr/local/lib/libkineto.a \
   kplay_cu.o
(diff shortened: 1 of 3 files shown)


sraikund16:
Thanks for fixes!

PR #1074, opened 2025-04-11
Support Option for Chrome Trace Unit to be in Nanoseconds

mcalman:
PR #923 changed the default chrome trace unit to be `ms`. This was done to make readability easier, however, it has the following adverse side effects:

1. Activities that are too close together are incorrectly treated by Chrome Trace-Viewer as nested and moved to separate lines. A discussion on the cause of this can be found [here](https://github.com/catapult-project/catapult/issues/927). Setting  `"displayTimeUnit": "ns"` resolves this issue

2. Activities with very short durations round to durations of `0.000 ms` when using `ms`

Simple illustration of issue with `"displayTimeUnit": "ms"` ([ms.pt.trace.json](https://github.com/user-attachments/files/19711600/ms.pt.trace.json)):
image

Issues resolved by switching to `"displayTimeUnit": "ns"` ([ns.pt.trace.json](https://github.com/user-attachments/files/19711601/ns.pt.trace.json)):
image

### DIFF
--- a/libkineto/include/GenericTraceActivity.h
+++ b/libkineto/include/GenericTraceActivity.h
@@ -141,7 +141,7 @@ class GenericTraceActivity : public ITraceActivity {
   struct Flow {
     Flow() : id(0), type(0), start(0) {}
     // Ids must be unique within each type
-    uint32_t id : 27;
+    uint32_t id;
     // Type will be used to connect flows between profilers, as
     // well as look up flow information (name etc)
     uint32_t type : 4;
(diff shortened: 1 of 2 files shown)


sraikund16:
Unfortunately, our internal users have voiced that they prefer the default to be ms. Can you instead introduce a flag or env var to set this?

mcalman:
@sraikund16 I have updated the PR keep the default as `ms` and only switch to `ns` if the build flag `-DDISPLAY_TRACE_IN_NS` is set.

PR #1082, opened 2025-05-05
Roctracer: collect cpu timestamps in approximateTime to avoid ts inversions

mwootton:
Roctracer reports gpu timestamp values is CLOCK_MONOTONIC.  This is an accurate and monotonic clock suitable for profiling.  This causes issue when trying to interface with the mixture of inaccurate and non-monotonic clocks used in kineto and pytorch.

Currently there are timestamp inversions between hip calls and autograd ranges.  This is causing headaches with analysis (as callers are exiting before their callees). The only reliable solution to avoid this is to collect cpu timestamps for hip calls in the same clock as pytorch/autograd, i.e. approximateTime().

GPU timestamps (from the accurate clock) will still need to be converted to the compromised clock.  Ideally we could do this incrementally as the kernel completions come in... in an effort to control the clock divergence.  Sadly the time converter is not available at collection time so it is not possible to compute an offset at collection time.  So all gpu timestamps will use the same time offset (post-collection) and perhaps drift terribly over long traces.

In this PR:
Collect hip call timestamps in approximateClock
Convert gpu timestamps similarly to before


mwootton:
Merged to work with #1080 and #1083

sraikund16 (approved):

PR #1083, opened 2025-05-06
Use TSC by default for GPU Only

sraikund16:
Summary: Changed implementation to use make own converter and use in the GPU only case instead of defaulting to a system clock.

Differential Revision: D74278271


sraikund16:
@mwootton

PR #1088, opened 2025-05-21
Add toggleCollectionDynamic for XPUPTI

frost-intel:
This PR adds support for toggleCollectionDynamic for Xpupti.

Since Xpupti enables/disables activities based on the session rather than the profiler instance, I cast the `IActivityProfilerSession` to an `XpuptiActivitySession`. **This assumes that all sessions are XpuptiActivitySessions**.  In upstream Pytorch, the design rule for now is that there is no simultaneous support for XPU and any other device (CUDA/Roc), so this assumption should be safe?


zejun-chen:
> This PR adds support for toggleCollectionDynamic for Xpupti.
> 
> Since Xpupti enables/disables activities based on the session rather than the profiler instance, I cast the `IActivityProfilerSession` to an `XpuptiActivitySession`. **This assumes that all sessions are XpuptiActivitySessions**. In upstream Pytorch, the design rule for now is that there is no simultaneous support for XPU and any other device (CUDA/Roc), so this assumption should be safe?

The assumption makes sense because when HAS_XPUPTI=true, only xpu device is active i think, so you can cast the base class to xpu class.

frost-intel:
> LGTM but could you find if there are related UTs for this feature. We need to enable the associated UTs.

It looks like there isn't any test coverage for this feature in kineto, but I'll edit [test_profiler.py](https://github.com/pytorch/pytorch/blob/3f34d260401d07485fb9be672284ffd4ee8a1634/test/profiler/test_profiler.py#L2038) in pytorch to cover the XPU case when adding the toggle functionality.

Update: Tests for XPU toggling have been added in pytorch/pytorch#155135

frost-intel:
@sraikund16 Could you please review this PR?

libkineto/src/CuptiActivityProfiler.cpp
```diff
@@ -1209,6 +1212,12 @@ void CuptiActivityProfiler::toggleCollectionDynamic(const bool enable) {
     cupti_.disableActivities(derivedConfig_->profileActivityTypes());
   }
 #endif
+#ifdef HAS_XPUPTI
+  for (auto& session : sessions_) {
```
sraikund16, line 1216: Can you leave a comment here explaining that you are making the assumption that all sessions will be XPUPTI since there are no downstream profilers for it? I think a TODO to determine session type is a good idea

sraikund16 (approved):
Overall, LGTM!

PR #1098, opened 2025-06-02
Fix fmt::format() in logging

sraikund16:
Summary: Change << fmt::format() to fmt::print for *most* logging cases. We can't do it for the LOG(...) cases because it can return (void) 0 if the severity is not high enough. Luckly for us, these cases are rare as they only occur on unexpected errors. The big instances of this speedup occur in Logger.cpp and output_json.cpp and we are able to fix all instances of it.

Differential Revision: D75828524

### DIFF
--- a/libkineto/src/Config.cpp
+++ b/libkineto/src/Config.cpp
@@ -12,6 +12,7 @@
 
 #include <fmt/chrono.h>
 #include <fmt/format.h>
+#include <fmt/ostream.h>
 #include <fmt/ranges.h>
 #include <time.h>
 #include <chrono>
@@ -577,37 +578,46 @@ void Config::setReportPeriod(milliseconds msecs) {
 }
 
 void Config::printActivityProfilerConfig(std::ostream& s) const {
-  s << "  Log file: " << activitiesLogFile() << std::endl;
+  fmt::print(s, "  Log file: {}\n", activitiesLogFile());
   if (hasProfileStartIteration()) {
-    s << "  Trace start Iteration: " << profileStartIteration() << std::endl;
-    s << "  Trace warmup Iterations: " << activitiesWarmupIterations()
-      << std::endl;
-    s << "  Trace profile Iterations: " << activitiesRunIterations()
-      << std::endl;
+    fmt::print(
+        s,
+        "  Trace start Iteration: {}\n"
+        "  Trace warmup Iterations: {}\n"
+        "  Trace profile Iterations: {}\n",
+        profileStartIteration(),
+        activitiesWarmupIterations(),
+        activitiesRunIterations());
     if (profileStartIterationRoundUp() > 0) {
-      s << "  Trace start iteration roundup : "
-        << profileStartIterationRoundUp() << std::endl;
+      fmt::print(
+          "  Trace start iteration roundup : {}\n",
+          profileStartIterationRoundUp());
     }
   } else if (hasProfileStartTime()) {
     std::time_t t_c = system_clock::to_time_t(requestTimestamp());
     std::tm tm{};
     localtime_r(&t_c, &tm);
-    s << "  Trace start time: " << fmt::format("{:%Y-%m-%d %H:%M:%S}", tm);
-    s << "  Trace duration: " << activitiesDuration().count() << "ms"
-      << std::endl;
-    s << "  Warmup duration: " << activitiesWarmupDuration().count() << "s"
-      << std::endl;
+    fmt::print(
+        s,
+        "  Trace start time: {:%Y-%m-%d %H:%M:%S}\n"
+        "  Trace duration: {}ms\n"
+        "  Warmup duration: {}s\n",
+        tm,
+        activitiesDuration().count(),
+        activitiesWarmupDuration().count());
   }
 
-  s << "  Max GPU buffer size: " << activitiesMaxGpuBufferSize() / 1024 / 1024
-    << "MB" << std::endl;
+  fmt::print(
+      s,
+      "  Max GPU buffer size: {:.0f}MB\n",
+      activitiesMaxGpuBufferSize() / 1024.0 / 1024.0);
 
-  std::vector<const char*> activities;
+  std::vector<std::string> activities;
+  activities.reserve(selectedActivityTypes_.size());
   for (const auto& activity : selectedActivityTypes_) {
     activities.push_back(toString(activity));
   }
-  s << "  Enabled activities: " << fmt::format("{}", fmt::join(activities, ","))
-    << std::endl;
+  fmt::print(s, "  Enabled activities: {}\n", fmt::join(activities, ","));
 
   AbstractConfig::printActivityProfilerConfig(s);
 }
--- a/libkineto/src/CuptiNvPerfMetric.h
+++ b/libkineto/src/CuptiNvPerfMetric.h
@@ -32,6 +32,8 @@ struct CuptiProfilerResult {
 /* Utilities for CUPTI and NVIDIA PerfWorks Metric API
  */
 
+// ok to use fmt::format as error will not occur often. Can't use fmt::print
+// easily since LOG(...) can return void, causes compiler error
 #define NVPW_CALL(call)                                                \
   [&]() -> bool {                                                      \
     NVPA_Status _status_ = call;                                       \
--- a/libkineto/src/CuptiRangeProfilerConfig.cpp
+++ b/libkineto/src/CuptiRangeProfilerConfig.cpp
@@ -12,6 +12,7 @@
 #include <stdlib.h>
 
 #include <fmt/format.h>
+#include <fmt/ostream.h>
 #include <fmt/ranges.h>
 #include <ostream>
 
@@ -59,12 +60,14 @@ void CuptiRangeProfilerConfig::setDefaults() {
 void CuptiRangeProfilerConfig::printActivityProfilerConfig(
     std::ostream& s) const {
   if (activitiesCuptiMetrics_.size() > 0) {
-    s << "Cupti Profiler metrics : "
-      << fmt::format("{}", fmt::join(activitiesCuptiMetrics_, ", "))
-      << std::endl;
-    s << "Cupti Profiler measure per kernel : " << cuptiProfilerPerKernel_
-      << std::endl;
-    s << "Cupti Profiler max ranges : " << cuptiProfilerMaxRanges_ << std::endl;
+    fmt::print(
+        s,
+        "Cupti Profiler metrics : {}\n",
+        "Cupti Profiler measure per kernel : {}\n",
+        "Cupti Profiler max ranges : {}\n",
+        fmt::join(activitiesCuptiMetrics_, ", "),
+        cuptiProfilerPerKernel_,
+        cuptiProfilerMaxRanges_);
   }
 }
 
--- a/libkineto/src/Logger.cpp
+++ b/libkineto/src/Logger.cpp
@@ -21,6 +21,7 @@
 
 #include <fmt/chrono.h>
 #include <fmt/format.h>
+#include <fmt/ostream.h>
 
 #include "ThreadUtil.h"
 
@@ -41,10 +42,15 @@ Logger::Logger(int severity, int line, const char* filePath, int errnum)
       std::chrono::system_clock::to_time_t(std::chrono::system_clock::now());
   std::tm tm_result;
   localtime_r(&tt, &tm_result);
-  const char* file = strrchr(filePath, '/');
-  buf_ << fmt::format("{:%Y-%m-%d %H:%M:%S}", tm_result) << " "
-       << processId(false) << ":" << systemThreadId(false) << " "
-       << (file ? file + 1 : filePath) << ":" << line << "] ";
+  const char* file = std::strrchr(filePath, '/');
+  fmt::print(
+      buf_,
+      "{:%Y-%m-%d %H:%M:%S} {}:{} {}:{}] ",
+      tm_result,
+      processId(false),
+      systemThreadId(false),
+      (file ? file + 1 : filePath),
+      line);
 }
 
 Logger::~Logger() {
--- a/libkineto/src/output_json.cpp
+++ b/libkineto/src/output_json.cpp
@@ -9,6 +9,7 @@
 #include "output_json.h"
 
 #include <fmt/format.h>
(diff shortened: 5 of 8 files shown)


libkineto/src/Config.cpp
```diff
@@ -592,11 +593,10 @@ void Config::printActivityProfilerConfig(std::ostream& s) const {
...
     localtime_r(&t_c, &tm);
-    s << "  Trace start time: " << fmt::format("{:%Y-%m-%d %H:%M:%S}", tm);
-    s << "  Trace duration: " << activitiesDuration().count() << "ms"
-      << std::endl;
-    s << "  Warmup duration: " << activitiesWarmupDuration().count() << "s"
-      << std::endl;
+    fmt::print(s, "  Trace start time: {:%Y-%m-%d %H:%M:%S}\n", tm);
+    fmt::print(s, "  Trace duration: {}ms\n", activitiesDuration().count());
+    fmt::print(
+        s, "  Warmup duration: {}s\n", activitiesWarmupDuration().count());
```
Skylion007: Reduce the overhead and better buffer but converting it one fmtlib call. fmtlib gives the largest speedup when formatting large strings. This is also safe because it does runtime and/or compile time type checks depending on the CPP version
```suggestion
    fmt::print(
        s,
        "  Trace start time: {:%Y-%m-%d %H:%M:%S}\n"
        "  Trace duration: {}ms\n"
        "  Warmup duration: {}s\n",
        tm,
        activitiesDuration().count(),
        activitiesWarmupDuration().count()
    );

File: libkineto/src/Config.cpp [hunk 2]
```diff
@@ -592,11 +593,10 @@ void Config::printActivityProfilerConfig(std::ostream& s) const {
...
   }
 
   s << "  Max GPU buffer size: " << activitiesMaxGpuBufferSize() / 1024 / 1024
```
Skylion007: This could also be merged together

libkineto/src/CuptiRangeProfilerConfig.cpp
```diff
@@ -59,12 +60,12 @@ void CuptiRangeProfilerConfig::setDefaults() {
...
-      << std::endl;
-    s << "Cupti Profiler measure per kernel : " << cuptiProfilerPerKernel_
-      << std::endl;
-    s << "Cupti Profiler max ranges : " << cuptiProfilerMaxRanges_ << std::endl;
+    fmt::print(
+        "Cupti Profiler metrics : {}\n",
+        fmt::join(activitiesCuptiMetrics_, ", "));
+    fmt::print(
+        "Cupti Profiler measure per kernel : {}\n", cuptiProfilerPerKernel_);
+    fmt::print("Cupti Profiler max ranges : {}\n", cuptiProfilerMaxRanges_);
```
Skylion007:
```suggestion
    fmt::print(
        s,
        "Cupti Profiler metrics : {}\n"
        "Cupti Profiler measure per kernel : {}\n
        "Cupti Profiler max ranges : {}\n"
        fmt::join(activitiesCuptiMetrics_, ", "),
        cuptiProfilerPerKernel_,
        cuptiProfilerMaxRanges_
    );
```
**Oof, you also forgot to include the ostream as an arg, so this will just go to std::out...** 

This fixes it and merges it into one call.

libkineto/src/output_csv.cpp
```diff
@@ -56,14 +57,17 @@ void EventCSVLogger::handleSample(
...
-      *out_ << device << ",";
-      *out_ << s.name;
+      fmt::print(
+          *out_,
+          "{:%Y-%m-%d %H:%M:%S},{},{},{},",
+          tm_result,
+          sample.deltaMsec,
+          device,
+          s.name);
       for (const auto& p : s.percentileValues) {
```
Skylion007: So close to being a good fmt::join, will be easy when we get C++20 range transforms

libkineto/src/output_json.cpp
```diff
@@ -135,24 +137,25 @@ void ChromeTraceLogger::handleTraceStart(
...
-  traceOf_ << fmt::format(
+  fmt::print(
+      traceOf_,
       R"JSON(
-  "displayTimeUnit": "{}",
-  "baseTimeNanoseconds": {},)JSON",
+    "displayTimeUnit": "{}",
+    "baseTimeNanoseconds": {},)JSON",
       display_unit,
       ChromeTraceBaseTime::singleton().get());
```
Skylion007: traceEvents json below can be merged into the print

libkineto/src/Config.cpp
```diff
@@ -606,8 +606,7 @@ void Config::printActivityProfilerConfig(std::ostream& s) const {
   for (const auto& activity : selectedActivityTypes_) {
     activities.push_back(toString(activity));
   }
-  s << "  Enabled activities: " << fmt::format("{}", fmt::join(activities, ","))
-    << std::endl;
```
Skylion007, line 610: FYI, clang-tidy perf checks tells you to avoid std::endl anyway because it unnecessarily flushes the ostream. We've already banned them upstream in PyTorch.

libkineto/src/Config.cpp
```diff
@@ -592,11 +593,10 @@ void Config::printActivityProfilerConfig(std::ostream& s) const {
     std::time_t t_c = system_clock::to_time_t(requestTimestamp());
     std::tm tm{};
     localtime_r(&t_c, &tm);
-    s << "  Trace start time: " << fmt::format("{:%Y-%m-%d %H:%M:%S}", tm);
```
Skylion007, line 595: The `hasProfileStartIteration()` if statement above could also be rewritten.

Skylion007:
You could convert the other LOG errors with a utility function that just skipped the formatting if it returned void, but yeah this is a great start. :)

libkineto/src/CuptiNvPerfMetric.h
```diff
@@ -32,6 +32,8 @@ struct CuptiProfilerResult {
 /* Utilities for CUPTI and NVIDIA PerfWorks Metric API
  */
 
+// ok to use fmt::format as error will not occur often. Can't use fmt::print
+// easily since LOG(...) can return void, causes compiler error
```
Skylion007, line 36: Could use a utility template function to check if the first arg is a void ostream and skip, but yeah, totally get it. https://github.com/MengRao/fmtlog is a fun library that solves this and a lot of problems, but yeah probably overkill here.

Something like:
```
template <typename... Args>
inline void safe_print(FILE* f, Args&&... args) {
    if (f) {
        fmt::print(f, std::forward<Args>(args)...);
    }
}
```
Probably with a std::nullptr overload or something is a better way to actually sort out the compile error. This is already a massive improvement though and a great start

libkineto/src/output_csv.cpp
```diff
@@ -56,14 +57,17 @@ void EventCSVLogger::handleSample(
...
-        *out_ << "," << p.second;
+        fmt::print(*out_, "{},", p.second);
```
Skylion007: Interesting, what is the type of p.second? This will fix the compiler error, (you can also define a custom fmt::formatter tye for the type of p.second of course, and probably should if you want to squeeze out even more performance). (Ah, it's a SampleValue): https://github.com/pytorch/kineto/blob/3f4beb08ad7e49be28eafa1216233df73e5be06f/libkineto/src/SampleListener.h#L114
```suggestion
        fmt::print(*out_, "{},", fmt::streamed(p.second));
```

Skylion007: Honestly, better to leave this line as it was before  since this might also create a bunch of temporary strings for a simple format.
or add this specialization into SampleListener.h
```cpp
#include <fmt/format.h>

namespace KINETO_NAMESPACE {

// Specialize fmt::formatter for SampleValue so you can write:
//   fmt::print("{}", someSampleValue);
template <>
struct fmt::formatter<SampleValue> {
  // Parsing is trivial: we don’t support any format‐specifiers beyond “{}”
  constexpr auto parse(fmt::format_parse_context& ctx) {
    return ctx.begin();
  }

  template <typename FormatContext>
  auto format(const SampleValue& v, FormatContext& ctx) {
    if (v.isInt()) {
      return fmt::format_to(ctx.out(), "{}", v.getInt());
    } else { // DOUBLE
      return fmt::format_to(ctx.out(), "{}", v.getDouble());
    }
  }
};

} // namespace KINETO_NAMESPACE
```
Might be worth it since this seems like it might be a hot path. Also would allow you rewrite this as a fmt::join if you wanted.

libkineto/src/Logger.cpp
```diff
@@ -42,9 +43,14 @@ Logger::Logger(int severity, int line, const char* filePath, int errnum)
   std::tm tm_result;
   localtime_r(&tt, &tm_result);
   const char* file = strrchr(filePath, '/');
```
Skylion007:
```suggestion
  const char* file = std::strrchr(filePath, '/');
```
STD + template FTW (make sure to add proper include)

libkineto/src/CuptiRangeProfilerConfig.cpp
sraikund16: shoot, thanks for the catch! I used AI to generate some of these instead of copy/pasting and didn't check the output carefully enough 😅

libkineto/src/output_csv.cpp
sraikund16: yeah I think this is only used for the EventProfilerController which is not really used anymore. Lets leave it for now and if necessary come back to this file

libkineto/src/CuptiNvPerfMetric.h
sraikund16: Sounds good to me! I was thinking about doing something similar to what we do in collection in profiler but the overloading logic seemed like more work than it was worth.

libkineto/src/Config.cpp
```diff
@@ -579,11 +580,14 @@ void Config::setReportPeriod(milliseconds msecs) {
...
+    fmt::print(
+        s,
+        "  Trace start Iteration: {}\n"
+        "  Trace warmup Iterations: {}\n"
+        "  Trace profile Iterations: {}\n",
+        profileStartIteration(),
+        activitiesWarmupIterations(),
+        activitiesRunIterations());
     if (profileStartIterationRoundUp() > 0) {
       s << "  Trace start iteration roundup : "
```
Skylion007: the endl here also flushes the `<<` which is unnecessary and slows down logging

libkineto/src/Config.cpp
```diff
@@ -592,22 +596,26 @@ void Config::printActivityProfilerConfig(std::ostream& s) const {
...
   }
 
-  s << "  Max GPU buffer size: " << activitiesMaxGpuBufferSize() / 1024 / 1024
-    << "MB" << std::endl;
+  fmt::print(
+      s,
+      "  Max GPU buffer size: {:.0f}MB\n",
+      activitiesMaxGpuBufferSize() / 1024.0 / 1024.0);
 
   std::vector<const char*> activities;
```
Skylion007:
```suggestion
  std::vector<const char*> activities;
  activities.reserve(selectedActivityTypes_.size());

File: libkineto/src/Config.cpp [hunk 13]
```diff
@@ -579,11 +580,14 @@ void Config::setReportPeriod(milliseconds msecs) {
 void Config::printActivityProfilerConfig(std::ostream& s) const {
   s << "  Log file: " << activitiesLogFile() << std::endl;
```
Skylion007:
```suggestion
  fmt::print(s, "  Log file: {}\n", activitiesLogFile());

File: libkineto/src/Config.cpp [hunk 14]
```diff
@@ -577,37 +578,45 @@ void Config::setReportPeriod(milliseconds msecs) {
 }
 
 void Config::printActivityProfilerConfig(std::ostream& s) const {
-  s << "  Log file: " << activitiesLogFile() << std::endl;
+  s << "  Log file: " << activitiesLogFile();
```
Skylion007: newline was dropped though:
```suggestion
  s << "  Log file: " << activitiesLogFile() << '\n';
```
or
```suggestion
  fmt::print(s, "  Log file: {}\n", activitiesLogFile());

File: libkineto/src/Config.cpp [hunk 15]
```diff
@@ -577,37 +578,45 @@ void Config::setReportPeriod(milliseconds msecs) {
...
   }
 
-  s << "  Max GPU buffer size: " << activitiesMaxGpuBufferSize() / 1024 / 1024
-    << "MB" << std::endl;
+  fmt::print(
+      s,
+      "  Max GPU buffer size: {:.0f}MB\n",
+      activitiesMaxGpuBufferSize() / 1024.0 / 1024.0);
 
   std::vector<const char*> activities;
```
Skylion007:
```suggestion
  std::vector<std::string> activities
```
Does toString already return a string? No reason to cast, just let it get captured by rValue

Skylion007 (approved):
One minor nit on the std::vector. Otherwise looks good

libkineto/src/output_json.cpp
```diff
@@ -117,9 +118,10 @@ void ChromeTraceLogger::metadataToJSON(
       distInfo_.distInfo_present_ = true;
     }
     sanitizeStrForJSON(sanitizedValue);
-    traceOf_ << fmt::format(
+    fmt::print(
+        traceOf_,
         R"JSON(
-  "{}": {},)JSON",
+      "{}": {},)JSON",
```
Skylion007, line 124: slight nit on alignment in this file, not quite aligned with above (but it wasn't aligned before so).

sraikund16: Our internal IDE for some reason aligns it this way, not sure why

libkineto/src/Config.cpp
```diff
@@ -577,37 +578,46 @@ void Config::setReportPeriod(milliseconds msecs) {
...
+        tm,
+        activitiesDuration().count(),
+        activitiesWarmupDuration().count());
   }
 
-  s << "  Max GPU buffer size: " << activitiesMaxGpuBufferSize() / 1024 / 1024
-    << "MB" << std::endl;
+  fmt::print(
+      s,
+      "  Max GPU buffer size: {:.0f}MB\n",
```
Skylion007, line 612: Was this a float or a double before? seems like a formatting change too?

sraikund16: Yeah I figured it made more sense to make it a float since users can customize the buffer size

PR #1100, opened 2025-06-04
Optimize string2hex method in output.json

Skylion007:
Use libfmt APIs directly to convert value to hexstring.

### DIFF
--- a/libkineto/src/output_json.cpp
+++ b/libkineto/src/output_json.cpp
@@ -10,6 +10,7 @@
 
 #include <fmt/format.h>
 #include <fmt/ostream.h>
+#include <iterator>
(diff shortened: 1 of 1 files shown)


libkineto/src/output_json.cpp
```diff
@@ -9,9 +9,9 @@
 #include "output_json.h"
 
 #include <fmt/format.h>
+#include <iterator>
```
Skylion007: We don't even `#include <string>` in this file though...

davidberard98:
@Skylion007 can you rebase over the merge conflict?

Skylion007:
@davidberard98 Done

PR #1102, opened 2025-06-12
Clear Buffer When Toggling

sraikund16:
Summary: Get rid of extraneous events when toggling. We have seen that in D76431700 there can sometimes be a build up of "junk" events in CUPTI sent over because we are turning off activities without flushing. We should flush whenever we toggle to make sure these events get dropped

Differential Revision: D76453586

### DIFF
--- a/libkineto/src/CuptiActivityProfiler.cpp
+++ b/libkineto/src/CuptiActivityProfiler.cpp
@@ -1198,18 +1198,22 @@ void CuptiActivityProfiler::ensureCollectTraceDone() {
 void CuptiActivityProfiler::toggleCollectionDynamic(const bool enable) {
 #ifdef HAS_CUPTI
   if (enable) {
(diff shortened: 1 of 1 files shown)


libkineto/src/CuptiActivityProfiler.cpp
```diff
@@ -1198,18 +1198,22 @@ void CuptiActivityProfiler::ensureCollectTraceDone() {
 void CuptiActivityProfiler::toggleCollectionDynamic(const bool enable) {
 #ifdef HAS_CUPTI
   if (enable) {
+    cupti_.clearActivities();
```
xw285cornell, line 1201: I feel what we need is a flush rather than clear

sraikund16: clearActivities does call flush in the function

PR #1109, opened 2025-06-29
Cleanup CMake targets

cyyever:
There are some changes:
1. Update minimum CMake version to 3.22 to use CUDA::cupti target to simplify code.
2. Use CUDA::nvperf_host if the target is available (for CMake >=3.27), otherwise we create it.
3. Fix Windows build errors.
4. Fix XPU sources detection.
5. Skip CUPTI tests on non-CUDA environments.
6. Simplify KINETO_LIBRARY_TYPE handling.
7. Don't install gtest targets.

Note: Windows shared builds don't work because no C++ symbol exported, this failure doesn't affect PyTorch builds because it always uses static builds.. This serious failure should be fixed in a later PR.


sraikund16:
Can you add a description?

cyyever:
@sraikund16 It is ready.

sraikund16:
Looks like it is failing to find cupti headers

cyyever:
@sraikund16 Fixed

sraikund16:
looks like same error :(

cyyever:
> looks like same error :(

It may because I wrote `LIBKINETO_NOCUPTI` to cache as string, now fixed.

sraikund16:
@cyyever still failing it looks like. Do you mind running the tests locally and then letting me know when it ready?

cyyever:
@sraikund16 It's interesting that I always pass locally..

cyyever:
@sraikund16 Is it possible that cupti is not installed in the docker image?

cyyever:
@sraikund16 I have reverted cupti detection to previous behaviour.

PR #1115, opened 2025-07-18
[CUPTI] Do not clear out per-thread buffer flag

Aidyn-A:
According to the [documentation](https://docs.nvidia.com/cupti/api/group__CUPTI__ACTIVITY__API.html#_CPPv4N23CUpti_ActivityAttribute46CUPTI_ACTIVITY_ATTR_PER_THREAD_ACTIVITY_BUFFERE) it is not safe to modify `CUPTI_ACTIVITY_ATTR_PER_THREAD_ACTIVITY_BUFFER` attribute: "Changing this attribute in the middle of the profiling session will result in undefined behavior."

### DIFF
--- a/libkineto/src/CuptiActivityApi.cpp
+++ b/libkineto/src/CuptiActivityApi.cpp
@@ -202,6 +202,15 @@ std::unique_ptr<CuptiActivityBufferMap> CuptiActivityApi::activityBuffers() {
     flushOverhead =
         duration_cast<microseconds>(system_clock::now() - t1).count();
   }
+
+#if (CUDART_VERSION >= 12030)
+  // Clear out per-thread buffer flag in case it was set
+  uint8_t value = 0;
+  size_t sizeof_value = sizeof(value);
+
+  CUPTI_CALL(cuptiActivitySetAttribute(
+      CUPTI_ACTIVITY_ATTR_PER_THREAD_ACTIVITY_BUFFER, &sizeof_value, &value));
+#endif // (CUDART_VERSION >= 12030)
 #endif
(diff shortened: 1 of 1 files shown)


sraikund16:
@Aidyn-A the flag is being turned off after all of the profiling activity is disabled, so I believe it is safe in this case.

Aidyn-A:
> @Aidyn-A the flag is being turned off after all of the profiling activity is disabled, so I believe it is safe in this case.

Thanks for the quick response. My experiments show that it is still not safe call it after all activities are disabled. If clearing it out is a must, it is safe to do it after the flush.

sraikund16:
Ok sounds good lets add it to `CuptiActivityApi::flushActivities` then?

libkineto/src/CuptiActivityApi.cpp
```diff
@@ -202,6 +202,15 @@ std::unique_ptr<CuptiActivityBufferMap> CuptiActivityApi::activityBuffers() {
...
         duration_cast<microseconds>(system_clock::now() - t1).count();
   }
+
+#if (CUDART_VERSION >= 12030)
+  // Clear out per-thread buffer flag in case it was set
+  uint8_t value = 0;
+  size_t sizeof_value = sizeof(value);
+
+  CUPTI_CALL(cuptiActivitySetAttribute(
+      CUPTI_ACTIVITY_ATTR_PER_THREAD_ACTIVITY_BUFFER, &sizeof_value, &value));
```
Aidyn-A, line 212: This location seems the most appropriate place to reset the flag.

sraikund16: sgtm! Thanks!

PR #1119, opened 2025-08-04
Should check if processGroupDesc is empty when adding processGroupDesc info

D-D-H:


sraikund16:
Can you explain why this change is necessary

D-D-H:
Hi @sraikund16 ,

We occasionally encountered the following invalid JSON format when we use PyTorch Profiler internally.

```
{
  "ph": "X", "cat": "kernel", "name": "ncclDevKernel_AllReduce_Sum_f16_RING_LL(ncclDevKernelArgsStorage<4096ul>)", "pid": 0, "tid": 33,
  "ts": 2889716883753.876, "dur": 309.282,
  "args": {
    "External id": 270910,      "queued": 0, "device": 0, "context": 1,      "stream": 33, "correlation": 9990707,      "registers per thread": 96,      "shared memory": 89344,      "blocks per SM": 0.027778,      "warps per SM": 0.472222,      "grid": [2, 1, 1],      "block": [544, 1, 1],      "est. achieved occupancy %": 0, "Collective name": "allreduce", "In msg nelems": 2097152, "Out msg nelems": 2097152, "Group size": 2, "dtype": "Half", "In split size": "[]", "Out split size": "[]", "Process Group Name": "12", "Process Group Description": , "Process Group Ranks": "[0, 1]"
  }
},
```

Although Perfetto can parse it, we sometimes need to parse it ourselves using Python json module.

sraikund16:
@D-D-H ahh thanks for clearing this up. Can we instead check that both processGroupDesc and processGroupName are both not empty?

D-D-H:
> @D-D-H ahh thanks for clearing this up. Can we instead check that both processGroupDesc and processGroupName are both not empty?

Okay.

PR #1128, opened 2025-08-25
Add rocprofiler-sdk support 2

mwootton:
Supersedes https://github.com/pytorch/kineto/pull/1050

Convert to using rocprofiler-sdk instead of roctracer for collecting hip api calls and AMD gpu activity.

Reuses most existing roctracer infrastructure with a name for name replacement. Simultaneous support for both roctracer and rocprofiler-sdk was deemed impractical. This would require a whole new set of #ifdefs, a major refactor of the roctracer code, and additional build support. Even then, only one could be active at a time (and you wouldn't want both active).

In homage to the abandoned refactor, RocLogger.cpp/h were created to contain the rocprofbase classes and the api filter.

Roctracer has no established end date.
Rocprofiler-sdk is in rocm_3.1 forward.

This will create a dependency where (newest kineto on old rocm) and (old kineto on newest rocm) could fail to build with AMD gpu support. That window is already over 1 year wide.


pruthvistony:
@mwootton ,
Since this PR breaks internal builds based on older ROCm 6.2 release, can you update this PR and conditionalize it based on ROCm version.

ie. 
use ROCTracer if ROCM_VERSION <6.4
and use ROCProf if ROCM_VERSION >=6.4

conditionally compile rocTracer file or rocProf files and keep both files in the repo.

cc @jeffdaily

mwootton:
Added support to conditionally compile the older roctracer implementation on rocm versions < 6.4.

mwootton:
@sraikund16 Could you take a look at this an confirm it functions properly on your internal builds?

PR #1130, opened 2025-08-26
Enable Windows shared libary builds by exporting all symbols

cyyever:


cyyever:
@sraikund16 FYI

sraikund16:
@cyyever I am not sure this will fix the issue. Have you tried testing it locally?

The problem is a linking issue, not an exporting one. It looks like when we are building torch_cpu for windows, the dependency for nvperf_host.lib does not propagate from kineto if using `INTERFACE IMPORTED`. We might need to use the direct path as we did in the old implementation since we don't want to add these dependencies to the parent Cmake file.

sraikund16:
I think this might be the appropriate fix: https://github.com/pytorch/kineto/pull/1133

cyyever:
@sraikund16 This PR is not for fixing the PyTorch issue, it is for the Windows shared builds in general. I have tested it locally on Windows, the tests can link to kineto dll and pass.

aaronenyeshi (approved):

PR #1131, opened 2025-08-26
Fix HAS_CUPTI_RANGE_PROFILER check

cyyever:
Enable `USE_CUPTI_RANGE_PROFILER`.


libkineto/src/CuptiRangeProfilerApi.h
```diff
@@ -13,14 +13,13 @@
 #include <cuda_runtime_api.h>
 // Using CUDA 11 and above due to usage of API:
 // cuptiProfilerGetCounterAvailability.
-// Starting from CUDA 12.06 the Profiler API is superseded by Range Profiler API
+// Starting from CUDA 12.6 the Profiler API is superseded by Range Profiler API
 // This needs significant rework. See
 // https://docs.nvidia.com/cupti/main/main.html#evolution-of-the-profiling-apis
 #if defined(USE_CUPTI_RANGE_PROFILER) && defined(CUDART_VERSION) && \
-    CUDART_VERSION >= 10000 && CUDA_VERSION >= 11000 && CUDA_VERSION <= 12060
+    CUDART_VERSION >= 10000 && CUDA_VERSION >= 12060
```
sraikund16, line 20: Not sure this is correct. I recall the range profiler breaking on later versions

cyyever: @sraikund16 Can reproduce? Nevertheless, the guard is wrong according to the above comment. If the profile still breaks, we should fix it rather than incorrectly disabling it.

sraikund16: @cyyever the api has been entirely replaced it would require an entire rework to enable it for those versions, which is currently low priority

sraikund16: @cyyever based on the NV link provided, it looks like we can use the current APIs until 13.0 although it is not recommended. However, by 13.0 we will need to use the new APIs as the old ones are deprecated. Either way it would require a rework.

cyyever: If the current APIs exist during the whole CUDA 13 cycle, then we have lots of time for refactoring.

PR #1135, opened 2025-08-29
Revert CMakeList files

sraikund16:
Summary:
Right now we are blocked from updating the Kineto Hash because of build issues on windows. Not only are windows builds broken, but also the CMakeLists were causing spurious errors in MacOS builds: https://github.com/pytorch/pytorch/issues/161510

Lets revert to the old version so there are no changes when updating the hash and then reintroduce the changes one-by one later to better understand what is breaking.

Differential Revision: D81332424


cyyever:
@sraikund16 It looks like the recent bump of dynolog is the cause. Reverting CMakeList is not a good idea and wouldn't help.

Issue #1121, opened 2025-08-07
[Feature] Enabling dynamic profiler plugins


briancoutinho:
# Enabling dynamic profiler plugins (Feature Proposal)
Authors:  @yisitu, @zli669, @briancoutinho (NVIDIA)

## TLDR

* We would like to contribute a feature to Kineto to enable the plugging in of a new profiler for NVIDIA GPUs for streaming performance metrics and low-overhead tracing over a long duration.
* We propose a Dynamic plugin capability that enables Kineto to load the new profiler module. The Dynamic interface is an extension of the statically compiled [ActivityProfilerPlugin API](https://github.com/pytorch/kineto/blob/main/libkineto/include/IActivityProfiler.h#L147C7-L147C24). Dynamic loading makes Kineto more extensible and decouples plugin development from Kineto/PyTorch mainline.

### Why a dynamic plugin?

* The plugin feature allows the development and extension of new profiler capabilities before they are ready to be published. Otherwise, developing in PyTorch and Kineto will require a full compile time source integration and slow iterative development cycles.  
* Extending the [ActivityProfiler plugin interface](https://github.com/pytorch/kineto/blob/main/libkineto/include/IActivityProfiler.h#L147C7-L147C24) to enable dynamic plugins has the added benefit of making kineto more extensible. Closed source plugins from other vendors or components can be easily integrated into kineto without complex compiler source hacks. It also permits independent development and prototyping of plugins.

## Overview of Changes

We will be posting a PR shortly; the changes primarily include:

1. Dynamic plugin C interface and implementation.  
   1. Implement a fully C style interface for Activity Profiler plugins and add the shim necessary.  
   2. Basic versioning and forward/backward compatibility support.  
2. Adding a runtime flag for CUPTI to be disabled so that it does not conflict with the new profiler plugin.

## Prototype Results

We have a working prototype with kineto. You can download the trace attached below

* CUDA kernel events and correlation arrows work as usual.  
* Also see the “GPU PM Counter” rows. Potentially any valid set of PM counters can be sampled.

<img width="738" height="336" alt="Image" src="https://github.com/user-attachments/assets/2e419b9b-9a69-4fa1-90f3-7d08a6ffe2bb" />

Here is a sample trace json that can be viewed in Perfetto or Chrome Trace.
[resnet_training.json.gz](https://github.com/user-attachments/files/21673476/resnet_training.json.gz)

Please let us know any questions or concerns.

[...]

PR #1148, opened 2025-09-18
Add dynamic plugin along with counter events, cupti disable support

briancoutinho:
Credits to @zli669 for implementing this change, I am just setting this up here for contribution to kineto.

# Overview
Adds the capability to dynamically load plugins for Kineto (#1121). The core idea is to have plugin modules that can be made available as a shared object file (.so). Kineto loads all .so object files in a specified path:
```
export KINETO_PLUGIN_LIB_DIR_PATH=/foo/bar/
```
These plugins can then register themselves with kineto, start/stop profiling and return trace events to kineto.
Please look at `DynamicPluginTest.cpp` for example usage of this API.

The PR also enables counter events that can show up in the trace.

# Details
## Companion changes
First, let's address some simpler changes:
1. **Support for Performance Counter Events in Chrome Trace Format**: Adds a new activity type `GPU_PM_COUNTER`, enabling a stream of performance events to be logged by any plugin in Kineto.
2. **Environment Variable `KINETO_DISABLE_CUPTI`**: Disables using CUPTI on NVIDIA GPUs to support non-CUPTI based profiling modules. This helps avoid conflicts with CUPTI, as NVIDIA support is currently closely coupled with it.

## Dynamic Plugin changes

The Dynamic plugin interface is a generalization of the [IActivityProfiler](https://github.com/pytorch/kineto/blob/cccb043c598d8624f6a0268231ecb71625807569/libkineto/include/IActivityProfiler.h#L147) interface in Kineto. The support is implemented in three parts, recommended for review in this order:

### 1) Plugin C Interface: `libkineto/include/KinetoDynamicPluginInterface.h`
The plugin shared object must implement `KinetoPlugin_register()`, providing function pointers for trace plugin functions: start(), stop(), create(), destroy(), processEvents(). Kineto calls these function pointers.

To ensure ABI compatibility and avoid C++ compiler mismatches, a pure C interface is used. This interface includes C versions of key structures and enums like `ActivityType`, `ProfileEvent`, `Flows`, etc.

Lastly, To transfer trace data from the plugin to Kineto, an opaque C object called `KinetoTraceBuilder` is used. The key idea is the trace builder handles generating events and transferring them to kineto's Activity profiler. This ensures **there is NO dynamic memory being passed around between plugin shared object and kineto**

### 2) Shim to Internal Plugin Interface: `libkineto/src/dynamic_plugin/PluginProfiler.h`
The `PluginProfiler` class implements the IActivityProfiler interface used internally in Kineto. It controls the profiling session and generates session trace data using the function pointers from the shared object.

#### 2.2) Trace Event Builder: `libkineto/src/dynamic_plugin/PluginTraceBuilder.h`
Implements the TraceEvent builder used by the shared object plugin.

### 3) Plugin Loader: `libkineto/src/dynamic_plugin/PluginLoader.h`
Finally, the plugin loader handles the discovery of shared object plugins and dynamically loads them into the address space using standard techniques.

# Testing

## Unit Test

Added a `DynamicPlugTest` which basically tests out the PluginProfiler class and the creation/start/stop functions. It also tests the trace event builder. The unit test does not do any dynamic loading though.
Build
```
$> mkdir build && cd build
$> cmake .. &&  make
```
Test
```
$> ctest -R 'DynamicPlugin.*'
Test project /localhome/local-bcoutinho/kineto/kineto_brian/libkineto/build
    Start 15: DynamicPluginTest.PluginProfilerLifecycle
1/2 Test #15: DynamicPluginTest.PluginProfilerLifecycle ...   Passed    0.00 sec
    Start 16: DynamicPluginTest.EventBuilderProcessing
2/2 Test #16: DynamicPluginTest.EventBuilderProcessing ....   Passed    0.00 sec

100% tests passed, 0 tests failed out of 2
```

## End end test with Always On Profiling Plugin

Build PyTorch and import this branch to third_party/kineto. Base version tested on ``
I used a simple test program
Normal operation uses CUPTI
```
python3 ~/torch_samples/test_profiling_simple.py
```

Run using plugin: 
```
$> $ KINETO_LOG_LEVEL=0 KINETO_PLUGIN_LIB_DIR_PATH=$AON_KINETO_PLUGIN_LIB_DIR_PATH KINETO_DISABLE_CUPTI=1 python3 ~/torch_samples/test_profiling_simple.py
Using device: cuda
INFO:2025-10-03 21:48:26 384669:384669 init.cpp:164] Setting initCupti = 0 from environment KINETO_DISABLE_CUPTI=1

INFO:2025-10-03 21:48:26 384669:384669 PluginLoader.h:127] Found symbol KinetoPlugin_register() from /home/bcoutinho/aon_kineto_plugin_build/c36614826/libAonKinetoPlugin.so
*********************************************
***** [AON] KinetoPlugin_register(). *******
*********************************************
INFO:2025-10-03 21:48:26 384669:384669 CuptiActivityProfiler.cpp:244] CUDA versions. CUPTI: 130001; Runtime: 13000; Driver: 13000
*********************************************
******** [AON] [PLUGIN] query(). *********
*********************************************
Adding supported activities  Log file: /tmp/libkineto_activities_384669.json
  Trace start time: 2025-10-03 21:48:33
  Trace duration: 500ms
  Warmup duration: 5s
  Max GPU buffer size: 128MB
  Enabled activities: cpu_op,user_annotation,gpu_user_annotation,gpu_memcpy,gpu_memset,kernel,external_correlation,cuda_runtime,cuda_driver,cpu_instant_event,python_function,overhead,xpu_runtime,privateuse1_runtime,privateuse1_driver
INFO:2025-10-03 21:48:26 384669:384669 CuptiActivityProfiler.cpp:1017] [Profiler = AON Profiler] Evaluating whether to run child profiler.
*********************************************
******** [AON] [PLUGIN] create(). ********
*********************************************
INFO:2025-10-03 21:48:29 384669:384669 CuptiActivityProfiler.cpp:1025] [Profiler = AON Profiler] Running child profiler AON Profiler for 500 ms
...
INFO:2025-10-03 21:48:29 384669:384669 CuptiActivityProfiler.cpp:1242] Starting child profiler session
*********************************************
******** [AON] [PLUGIN] start(). *********
*********************************************
INFO:2025-10-03 21:48:29 384669:384669 CuptiActivityProfiler.cpp:1279] Stopping child profiler session
*********************************************
******** [AON] [PLUGIN] stop(). **********
*********************************************
STAGE:2025-10-03 21:48:29 384669:384669 ActivityProfilerController.cpp:396] Completed Stage: Collection
INFO:2025-10-03 21:48:29 384669:384669 CuptiActivityProfiler.cpp:293] Processing 1 CPU buffers
INFO:2025-10-03 21:48:29 384669:384669 CuptiActivityProfiler.cpp:394] Processing child profiler trace
*********************************************
***** [AON] [PLUGIN] processEvents(). *******
*********************************************
...
```
I can open the trace and find events added by plugin
```
  {
    "ph": "X", "cat": "kernel", "name": "_Z13gemmk1_kernelIifLi256ELi5ELb0ELb0ELb0ELb0E30cublasGemvTensorStridedBatchedIKfES2_S0_IfEfLi0EEv18cublasGemmk1ParamsIT0_T7_T8_T9_T10_N8biasTypeINS8_10value_typeES9_E4typeEE<<<(1, 1, 1), (256, 1, 1)>>>", "pid": 0, "tid": 7,
    "ts": 0.000, "dur": 0.000,
    "args": {
      "isImmediateLaunch": true, "hesCorrelationId": 2155942657, "apiCorrelationId": 8459009
    }
  },
```

### DIFF
--- a/libkineto/include/KinetoDynamicPluginInterface.h
+++ b/libkineto/include/KinetoDynamicPluginInterface.h
@@ -1,0 +1,448 @@
+#ifndef KINETO_DYNAMIC_PLUGIN_FORMAT_H
+#define KINETO_DYNAMIC_PLUGIN_FORMAT_H
+
+#include <stdint.h>
+
+#define KINETO_PLUGIN_LIB_DIR_PATH_ENV_VARIABLE "KINETO_PLUGIN_LIB_DIR_PATH"
+
+// NOTE: This file uses unpadded struct size for version control
+// To support backward compatibility, do NOT remove, reorder, or resize the
+// existing fields For details, visit
+// https://github.com/annarev/community/commit/09e231324c3b2b1f653c3f385fd2120edd460815
+#define KINETO_PLUGIN_UNPADDED_STRUCT_SIZE(struct_type, last_field_name) \
+  (size_t)(offsetof(struct_type, last_field_name) +                      \
+           sizeof(((struct_type*)1000)->last_field_name))
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+// Event types that can be supported via plugin. Please sync
+// with ActivityType.h if you add new event types.
+enum KinetoPlugin_ProfileEventType {
+  KINETO_PLUGIN_PROFILE_EVENT_TYPE_INVALID = 0,
+  KINETO_PLUGIN_PROFILE_EVENT_TYPE_CPU_OP, // cpu side ops
+  KINETO_PLUGIN_PROFILE_EVENT_TYPE_USER_ANNOTATION,
+  KINETO_PLUGIN_PROFILE_EVENT_TYPE_GPU_USER_ANNOTATION,
+  KINETO_PLUGIN_PROFILE_EVENT_TYPE_GPU_MEMCPY,
+  KINETO_PLUGIN_PROFILE_EVENT_TYPE_GPU_MEMSET,
+  KINETO_PLUGIN_PROFILE_EVENT_TYPE_CONCURRENT_KERNEL, // on-device kernels
+  KINETO_PLUGIN_PROFILE_EVENT_TYPE_EXTERNAL_CORRELATION,
+  KINETO_PLUGIN_PROFILE_EVENT_TYPE_CUDA_RUNTIME, // host side cuda runtime
+                                                 // events
+  KINETO_PLUGIN_PROFILE_EVENT_TYPE_CUDA_DRIVER, // host side cuda driver events
+  KINETO_PLUGIN_PROFILE_EVENT_TYPE_CPU_INSTANT_EVENT, // host side point-like
+                                                      // events
+  KINETO_PLUGIN_PROFILE_EVENT_TYPE_PYTHON_FUNCTION,
+  KINETO_PLUGIN_PROFILE_EVENT_TYPE_OVERHEAD, // CUPTI induced overhead events
+  KINETO_PLUGIN_PROFILE_EVENT_TYPE_MTIA_RUNTIME, // host side MTIA runtime
+                                                 // events
+  KINETO_PLUGIN_PROFILE_EVENT_TYPE_MTIA_CCP_EVENTS, // MTIA ondevice CCP events
+  KINETO_PLUGIN_PROFILE_EVENT_TYPE_MTIA_INSIGHT, // MTIA Insight Events
+                                                 // sampled from its overhead
+                                                 // API.
+  KINETO_PLUGIN_PROFILE_EVENT_TYPE_CUDA_SYNC, // synchronization events between
+                                              // runtime and kernels
+  // Optional Activity types
+  KINETO_PLUGIN_PROFILE_EVENT_TYPE_GLOW_RUNTIME, // host side glow runtime
+                                                 // events
+  KINETO_PLUGIN_PROFILE_EVENT_TYPE_CUDA_PROFILER_RANGE, // CUPTI Profiler range
+                                                        // for performance
+                                                        // metrics
+  KINETO_PLUGIN_PROFILE_EVENT_TYPE_HPU_OP, // HPU host side runtime event
+  KINETO_PLUGIN_PROFILE_EVENT_TYPE_XPU_RUNTIME, // host side xpu runtime events
+  KINETO_PLUGIN_PROFILE_EVENT_TYPE_COLLECTIVE_COMM, // collective communication
+  KINETO_PLUGIN_PROFILE_EVENT_TYPE_GPU_PM_COUNTER, // GPU performance monitoring
+                                                   // counter
+
+  // PRIVATEUSE1 Activity types are used for custom backends.
+  // The corresponding device type is `DeviceType::PrivateUse1` in PyTorch.
+  KINETO_PLUGIN_PROFILE_EVENT_TYPE_PRIVATEUSE1_RUNTIME, // host side privateUse1
+                                                        // runtime events
+  KINETO_PLUGIN_PROFILE_EVENT_TYPE_PRIVATEUSE1_DRIVER, // host side privateUse1
+                                                       // driver events
+
+  KINETO_PLUGIN_PROFILE_EVENT_NUM_TYPES
+};
+
+struct KinetoPlugin_ProfileEvent {
+  // Always set to KINETO_PLUGIN_PROFILE_EVENT_UNPADDED_STRUCT_SIZE
+  size_t unpaddedStructSize;
+  // Event type
+  KinetoPlugin_ProfileEventType eventType;
+  // Start timestamp in UTC nanoseconds
+  int64_t startTimeUtcNs;
+  // End timestamp in UTC nanoseconds
+  int64_t endTimeUtcNs;
+  // Event ID
+  int64_t eventId;
+  // If CPU, equivalent to process ID
+  union {
+    int32_t deviceId;
+    int32_t processId;
+  };
+  // If CPU, equivalent to thread ID
+  union {
+    int32_t resourceId;
+    int32_t threadId;
+  };
+};
+#define KINETO_PLUGIN_PROFILE_EVENT_UNPADDED_STRUCT_SIZE \
+  KINETO_PLUGIN_UNPADDED_STRUCT_SIZE(struct KinetoPlugin_ProfileEvent, threadId)
+
+enum KinetoPlugin_ProfileEventFlowType {
+  KINETO_PLUGIN_PROFILE_EVENT_FLOW_TYPE_INVALID = 0,
+  KINETO_PLUGIN_PROFILE_EVENT_FLOW_TYPE_FWD_BWD,
+  KINETO_PLUGIN_PROFILE_EVENT_FLOW_TYPE_ASYNC_CPU_GPU,
+  KINETO_PLUGIN_PROFILE_EVENT_NUM_FLOW_TYPES
+};
+
+struct KinetoPlugin_ProfileEventFlow {
+  // Always set to KINETO_PLUGIN_PROFILE_EVENT_FLOW_UNPADDED_STRUCT_SIZE
+  size_t unpaddedStructSize;
+  // Flow type
+  KinetoPlugin_ProfileEventFlowType flowType;
+  // Use the same non-zero ID to connect two events
+  uint32_t flowId;
+  // Set to true if flow is a starting point
+  bool isStartPoint;
+};
+#define KINETO_PLUGIN_PROFILE_EVENT_FLOW_UNPADDED_STRUCT_SIZE \
+  KINETO_PLUGIN_UNPADDED_STRUCT_SIZE(                         \
+      struct KinetoPlugin_ProfileEventFlow, isStartPoint)
+
+struct KinetoPlugin_ProfileDeviceInfo {
+  // Always set to KINETO_PLUGIN_PROFILE_DEVICE_INFO_UNPADDED_STRUCT_SIZE
+  size_t unpaddedStructSize;
+  // Device ID
+  int64_t deviceId;
+  // Device sort index
+  int64_t sortIndex;
+  // Device name
+  const char* pName;
+  // Device label
+  const char* pLabel;
+};
+#define KINETO_PLUGIN_PROFILE_DEVICE_INFO_UNPADDED_STRUCT_SIZE \
+  KINETO_PLUGIN_UNPADDED_STRUCT_SIZE(                          \
+      struct KinetoPlugin_ProfileDeviceInfo, pLabel)
+struct KinetoPlugin_ProfileResourceInfo {
+  // Always set to KINETO_PLUGIN_PROFILE_RESOURCE_INFO_UNPADDED_STRUCT_SIZE
+  size_t unpaddedStructSize;
+  // If CPU, equivalent to process ID
+  union {
+    int32_t deviceId;
+    int32_t processId;
+  };
+  // If CPU, equivalent to thread ID
+  union {
+    int32_t resourceId;
+    int32_t threadId;
+  };
+  // Lower value is displayed more towards top
+  int32_t displayOrder;
+  // Textual display in the viewer (ephemeral pointer)
+  const char* pName;
+};
+#define KINETO_PLUGIN_PROFILE_RESOURCE_INFO_UNPADDED_STRUCT_SIZE \
+  KINETO_PLUGIN_UNPADDED_STRUCT_SIZE(                            \
+      struct KinetoPlugin_ProfileResourceInfo, pName)
+
+typedef struct KinetoPlugin_TraceBuilderHandle KinetoPlugin_TraceBuilderHandle;
+
+// Trace Builder is an opaque object provided by Kineto that helps the plugin
+// log event data without having to serialize down the wire over a C interface
+// Trace Builder also takes care of memory management / ownership
+struct KinetoPlugin_TraceBuilder {
+  // Always set to KINETO_PLUGIN_TRACE_BUILDER_UNPADDED_STRUCT_SIZE
+  size_t unpaddedStructSize;
+
+  KinetoPlugin_TraceBuilderHandle* pTraceBuilderHandle;
+
+
+  // Usage: For each event, call addEvent(), then setLastEventName().
+  // Optionally call addLastEventMetadata() and setLastEventFlow().
+  // Use addDeviceInfo() and addResourceInfo() for device/resource metadata.
+  //
+  // Example:
+  //   KinetoPlugin_ProfileEvent event = {
+  //     .unpaddedStructSize = KINETO_PLUGIN_PROFILE_EVENT_UNPADDED_STRUCT_SIZE,
+  //     .eventType = KINETO_PLUGIN_PROFILE_EVENT_TYPE_CUDA_RUNTIME,
+  //     .startTimeUtcNs = 1000000000, .endTimeUtcNs = 1000005000,
+  //     .eventId = 1, .deviceId = 0, .resourceId = 123
+  //   };
+  //   traceBuilder->addEvent(traceBuilder->pTraceBuilderHandle, &event);
+  //   traceBuilder->setLastEventName(traceBuilder->pTraceBuilderHandle,
+  //   "cudaLaunchKernel");
+
+  // Several of these APIs do not have failure cases, but we still return an
+  // integer for future extensibility.
+
+  // returns 0 on success, -1 on failure, generally this is not expected to fail.
+  int (*addEvent)(
+      KinetoPlugin_TraceBuilderHandle* pTraceBuilderHandle,
+      const struct KinetoPlugin_ProfileEvent* pProfileEvent);
+
+  // returns 0 on success, -1 on failure, this can happen if the last event is not set.
+  int (*setLastEventName)(
+      KinetoPlugin_TraceBuilderHandle* pTraceBuilderHandle,
+      const char* pName);
+
+  // returns 0 on success, -1 on failure, this can happen if the last event is not set.
+  int (*setLastEventFlow)(
+      KinetoPlugin_TraceBuilderHandle* pTraceBuilderHandle,
+      const struct KinetoPlugin_ProfileEventFlow* pProfileEventFlow);
+  // If metadata value is a string itself (e.g., "metadata value"), MUST add
+  // additional quote around it (e.g., "\"metadata value\"") If metadata value
+  // is a list (e.g., 1, 2, 3), MUST add square bracket around it (e.g., "[1, 2,
+  // 3]")
+  // returns 0 on success, -1 on failure, this can happen if the last event is not set.
+  int (*addLastEventMetadata)(
+      KinetoPlugin_TraceBuilderHandle* pTraceBuilderHandle,
+      const char* pKey,
+      const char* pValue);
+
+  // returns 0 on success, -1 on failure, generally this is not expected to fail.
+  int (*addDeviceInfo)(
+      KinetoPlugin_TraceBuilderHandle* pTraceBuilderHandle,
+      const struct KinetoPlugin_ProfileDeviceInfo* pProfileDeviceInfo);
+
+  // returns 0 on success, -1 on failure, generally this is not expected to fail.
+  int (*addResourceInfo)(
+      KinetoPlugin_TraceBuilderHandle* pTraceBuilderHandle,
+      const struct KinetoPlugin_ProfileResourceInfo* pProfileResourceInfo);
+};
+#define KINETO_PLUGIN_TRACE_BUILDER_UNPADDED_STRUCT_SIZE \
+  KINETO_PLUGIN_UNPADDED_STRUCT_SIZE(                    \
+      struct KinetoPlugin_TraceBuilder, addResourceInfo)
+
+typedef struct KinetoPlugin_ProfilerHandle KinetoPlugin_ProfilerHandle;
+
+struct KinetoPlugin_ProfilerCreate_Params {
+  // Always set to KINETO_PLUGIN_PROFILER_CREATE_PARAMS_UNPADDED_STRUCT_SIZE
+  size_t unpaddedStructSize;
+
+  // [out] An instance created by plugin
+  KinetoPlugin_ProfilerHandle* pProfilerHandle;
+
+  // [in] Enabled activity types.
+  KinetoPlugin_ProfileEventType* pEnabledActivityTypes;
+
+  // [in] Max length of pEnabledActivityTypes
+  size_t enabledActivityTypesMaxLen;
+};
+#define KINETO_PLUGIN_PROFILER_CREATE_PARAMS_UNPADDED_STRUCT_SIZE \
+  KINETO_PLUGIN_UNPADDED_STRUCT_SIZE(                             \
+      struct KinetoPlugin_ProfilerCreate_Params, enabledActivityTypesMaxLen)
+
+struct KinetoPlugin_ProfilerDestroy_Params {
+  // Always set to KINETO_PLUGIN_PROFILER_DESTROY_PARAMS_UNPADDED_STRUCT_SIZE
+  size_t unpaddedStructSize;
+
+  // [in] An instance created via profilerCreate()
+  KinetoPlugin_ProfilerHandle* pProfilerHandle;
+};
+#define KINETO_PLUGIN_PROFILER_DESTROY_PARAMS_UNPADDED_STRUCT_SIZE \
+  KINETO_PLUGIN_UNPADDED_STRUCT_SIZE(                              \
+      struct KinetoPlugin_ProfilerDestroy_Params, pProfilerHandle)
+
+struct KinetoPlugin_ProfilerQuery_Params {
+  // Always set to KINETO_PLUGIN_PROFILER_QUERY_PARAMS_UNPADDED_STRUCT_SIZE
+  size_t unpaddedStructSize;
+
+  // [in] An instance created via profilerCreate()
+  KinetoPlugin_ProfilerHandle* pProfilerHandle;
+
+  // [in/out] Memory space to receive profiler name
+  char* pProfilerName;
+
+  // [in] Max length of pProfilerName excluding null terminator
+  size_t profilerNameMaxLen;
+
+  // [in/out] Supported activity types.
+  KinetoPlugin_ProfileEventType* pSupportedActivityTypes;
+
+  // [in] Max length of pSupportedActivityTypes
+  size_t supportedActivityTypesMaxLen;
+};
+#define KINETO_PLUGIN_PROFILER_QUERY_PARAMS_UNPADDED_STRUCT_SIZE \
+  KINETO_PLUGIN_UNPADDED_STRUCT_SIZE(                            \
+      struct KinetoPlugin_ProfilerQuery_Params, supportedActivityTypesMaxLen)
+
+struct KinetoPlugin_ProfilerStart_Params {
+  // Always set to KINETO_PLUGIN_PROFILER_START_PARAMS_UNPADDED_STRUCT_SIZE
+  size_t unpaddedStructSize;
+
+  // [in] An instance created via profilerCreate()
+  KinetoPlugin_ProfilerHandle* pProfilerHandle;
+};
+#define KINETO_PLUGIN_PROFILER_START_PARAMS_UNPADDED_STRUCT_SIZE \
+  KINETO_PLUGIN_UNPADDED_STRUCT_SIZE(                            \
+      struct KinetoPlugin_ProfilerStart_Params, pProfilerHandle)
+
+struct KinetoPlugin_ProfilerStop_Params {
+  // Always set to KINETO_PLUGIN_PROFILER_STOP_PARAMS_UNPADDED_STRUCT_SIZE
+  size_t unpaddedStructSize;
+
+  // [in] An instance created via profilerCreate()
+  KinetoPlugin_ProfilerHandle* pProfilerHandle;
+};
+#define KINETO_PLUGIN_PROFILER_STOP_PARAMS_UNPADDED_STRUCT_SIZE \
+  KINETO_PLUGIN_UNPADDED_STRUCT_SIZE(                           \
+      struct KinetoPlugin_ProfilerStop_Params, pProfilerHandle)
+
+struct KinetoPlugin_ProfilerProcessEvents_Params {
+  // Always set to
+  // KINETO_PLUGIN_PROFILER_PROCESS_EVENTS_PARAMS_UNPADDED_STRUCT_SIZE
+  size_t unpaddedStructSize;
+
+  // [in] An instance created via profilerCreate()
+  KinetoPlugin_ProfilerHandle* pProfilerHandle;
+
+  // [in] APIs to build a trace
+  const struct KinetoPlugin_TraceBuilder* pTraceBuilder;
+};
+#define KINETO_PLUGIN_PROFILER_PROCESS_EVENTS_PARAMS_UNPADDED_STRUCT_SIZE \
+  KINETO_PLUGIN_UNPADDED_STRUCT_SIZE(                                     \
+      struct KinetoPlugin_ProfilerProcessEvents_Params, pTraceBuilder)
+
+struct KinetoPlugin_ProfilerPushCorrelationId_Params {
+  // Always set to
+  // KINETO_PLUGIN_PROFILER_PUSH_CORRELATION_ID_PARAMS_UNPADDED_STRUCT_SIZE
+  size_t unpaddedStructSize;
+
+  // [in] Correlation ID
+  uint64_t correlationId;
+
+  // [in] An instance created via profilerCreate()
+  KinetoPlugin_ProfilerHandle* pProfilerHandle;
+};
+#define KINETO_PLUGIN_PROFILER_PUSH_CORRELATION_ID_PARAMS_UNPADDED_STRUCT_SIZE \
+  KINETO_PLUGIN_UNPADDED_STRUCT_SIZE(                                          \
+      struct KinetoPlugin_ProfilerPushCorrelationId_Params, pProfilerHandle)
+
+struct KinetoPlugin_ProfilerPopCorrelationId_Params {
+  // Always set to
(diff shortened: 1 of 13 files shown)


libkineto/include/DynamicPluginInterface.h
```diff
@@ -0,0 +1,277 @@
...
+  KINETO_PLUGIN_PROFILE_EVENT_TYPE_CPU_INSTANT_EVENT, // host side point-like
+                                                      // events
+  KINETO_PLUGIN_PROFILE_EVENT_TYPE_PYTHON_FUNCTION,
+  KINETO_PLUGIN_PROFILE_EVENT_TYPE_OVERHEAD,  // CUPTI induced overhead events
+                                              // sampled from its overhead API.
+  KINETO_PLUGIN_PROFILE_EVENT_TYPE_CUDA_SYNC, // synchronization events between
+                                              // runtime and kernels
+  KINETO_PLUGIN_PROFILE_EVENT_TYPE_GPU_PM_COUNTER, // GPU PM counters
+  KINETO_PLUGIN_PROFILE_EVENT_NUM_TYPES
+};
```
moksiuc, line 66: This part is the most problematic in my opinion. As it is fixed set it would be impossible to create new plugin without contributing to pytorch code and changing this file by adding new enums supported by new plugin.
Then there would exist enum in pytorch code that is used nowhere in the code, as it would be used only by the plugin.

How about more dynamic approach ?
Having integer here instead of fixed enums. The lower range 0..N would map to fixed enums and higher values would be assigned dynamically to plugins:

0 - A-1 : fixed enumx
A - B-1 : 'enums' for the 1st plugin
B - C-1 : 'enums' for the 2nd plugin
...

During initialization the plugin would get the base address in enums address space: A for 1st plugin, B for 2nd plugin.
In return it would provide the list of strings, containing supported activities, that would be converted in provided order to its enum space.

If you prefer to keep fixed list please add here KINETO_PLUGIN_PROFILE_EVENT_TYPE_XPU_RUNTIME.

libkineto/include/KinetoDynamicPluginInterface.h
```diff
@@ -0,0 +1,277 @@
...
+enum KinetoPlugin_ProfileEventType {
+  KINETO_PLUGIN_PROFILE_EVENT_TYPE_INVALID = 0,
+  KINETO_PLUGIN_PROFILE_EVENT_TYPE_CPU_OP, // cpu side ops
+  KINETO_PLUGIN_PROFILE_EVENT_TYPE_USER_ANNOTATION,
+  KINETO_PLUGIN_PROFILE_EVENT_TYPE_GPU_USER_ANNOTATION,
+  KINETO_PLUGIN_PROFILE_EVENT_TYPE_GPU_MEMCPY,
+  KINETO_PLUGIN_PROFILE_EVENT_TYPE_GPU_MEMSET,
+  KINETO_PLUGIN_PROFILE_EVENT_TYPE_CONCURRENT_KERNEL, // on-device kernels
+  KINETO_PLUGIN_PROFILE_EVENT_TYPE_EXTERNAL_CORRELATION,
+  KINETO_PLUGIN_PROFILE_EVENT_TYPE_CUDA_RUNTIME, // host side cuda runtime
```
tsocha, line 31: These events should be more generic in my opinion.

Why CUDA is hardcoded?
What about Intel's XPU, AMD's ROCM, Google's TPU, some custom FPGA, etc?

libkineto/test/DynamicPluginTest.cpp
```diff
@@ -0,0 +1,285 @@
...
+struct MockPluginHandle {
+  bool created = false;
+  bool active = false;
+  std::string name = "MockPlugin";
+};
+
+// Mock implementation of the plugin interface
+class MockPlugin {
+public:
+  static MockPluginHandle *handles[10]; // Support up to 10 handles for testing
```
tsocha: Since this test is written in C++, maybe it's worth to use smart pointers here.

libkineto/src/init.cpp
```diff
@@ -143,10 +144,19 @@ void libkineto_init(bool cpuOnly, bool logOnError) {
 #endif
 
 #ifdef HAS_CUPTI
+  bool initCupti = true;
+
+  const char* kDisableCuptiEnvVar = "KINETO_DISABLE_CUPTI";
+  const char* disableCuptiEnvVal = getenv(kDisableCuptiEnvVar);
+  if (disableCuptiEnvVal) {
+    initCupti = atoi(disableCuptiEnvVal) ? false : true;
+    LOG(INFO) << "Setting initCupti = " << initCupti << " from environment " << kDisableCuptiEnvVar << "=" << disableCuptiEnvVal;
```
tsocha: Don't we have any function to handle ENV variables?
Could you use/add it as an util?

libkineto/src/dynamic_plugin/PluginProfiler.h
```diff
@@ -0,0 +1,263 @@
...
+        KINETO_PLUGIN_PROFILER_QUERY_PARAMS_UNPADDED_STRUCT_SIZE};
+    queryParams.pProfilerHandle = nullptr;
+    queryParams.pProfilerName = &profilerName[0];
+    queryParams.profilerNameMaxLen = 31;
+    errorCode = profiler_.profilerQuery(&queryParams);
+    if (errorCode != 0) {
+      name_.assign("N/A");
+    } else {
+      name_.assign(queryParams.pProfilerName);
+    }
```
australopitek: The same code is already in PluginProfilerSession constructor. Consider extracting it to common function.

libkineto/include/KinetoDynamicPluginInterface.h
briancoutinho: @tsocha I completely agree with you. Note that this enum is reflecting the C++ enums in ActivityType.h.
https://github.com/pytorch/kineto/blob/main/libkineto/include/ActivityType.h#L19-L28 

Due to historical reasons the activities are named after CUDA but ideally they should be named generically rather than for each GPUs's runtime. I don't think we can fix it in this PR but I would do something like
```
enum class .. {

GPU_RUNTIME

CUDA_RUNTIME=GPU_RUNTIME
XPU_RUNTIME=GPU_RUNTIME
MTIA_RUNTIME=GPU_RUNTIME
```
cc @sraikund16  that makes sense long term?

libkineto/include/DynamicPluginInterface.h
briancoutinho: Actually, we should try to make the ActivityTypes more generic so we have lesser number of event types in the list and every new platform should not be adding its own version runtime and *PU events. That way new plugins should rarely ever need adding enums.

For now I will update with adding all activity types in the header file.

libkineto/src/init.cpp
briancoutinho: Yea I can use that. I think do that.

libkineto/src/dynamic_plugin/PluginProfiler.h
briancoutinho: Ah you right. We should probably compute this once here and pass it on as a parameter to the session classes. Thx for catching this!

libkineto/test/DynamicPluginTest.cpp
briancoutinho: Done

libkineto/include/DynamicPluginInterface.h
briancoutinho: I have added all the activity types for completeness here now.

libkineto/src/dynamic_plugin/PluginProfiler.h
```diff
@@ -0,0 +1,246 @@
...
+        ActivityType::CUDA_PROFILER_RANGE};
+    return supported;
+  }
+
+  std::unique_ptr<IActivityProfilerSession>
+  configure(const std::set<ActivityType> &activity_types,
+            const Config &config) override {
+    // [TODO] Check activity types to determine if creating session or nullptr
+    return std::make_unique<PluginProfilerSession>(profiler_, name_);
+  }
```
australopitek: Do you plan to pass activity_types to the plugin profiler? Currently it doesn't know what activities should be collected. Also, there's no validation, so in my tests all events reported by the plugin in `.profilerProcessEvents` were pushed to the trace json, even events not requested by the user.
activity_types could be passed in `.profilerCreate` function.

briancoutinho:
cc @sraikund16  @davidberard98  this is ready for review. 
I have some internal tests pending but that should not impact code on kineto.

libkineto/src/dynamic_plugin/PluginProfiler.h
briancoutinho: Yea that was a TODO, we pass in set of enabled activities and the configure() function decides if plugin should be enabled or not. If not it returns nullptr. Let me update that in a revision to this PR.

There aren't validations when events are generated if they match activities pushed to json though. This is true of the static plugin API as well, each plugin should stick to what events it will be generating.

Adding validation can be done in the base class as a different PR.

libkineto/src/dynamic_plugin/PluginProfiler.h
```diff
@@ -0,0 +1,246 @@
...
+  void processTrace(ActivityLogger &logger) override {
+    if (pProfilerHandle_ == nullptr) {
+      return;
+    }
+
+    auto traceSpan =
+        TraceSpan(lastStartTimestampUtcNs_, lastStopTimestampUtcNs_, name_);
+    PluginTraceBuilder pluginTraceBuilder{traceSpan};
+    const KinetoPlugin_TraceBuilder traceBuilder =
+        pluginTraceBuilder.toCTraceBuilder();
```
australopitek, line 182: What about .iteration parameter of TraceSpan? I think it's not increased anywhere in this plugin, so in traces it's always equal to -1.

briancoutinho: @australopitek I have not see the iteration element actually being used. PyTorch will typically add a "ProfilerStep"N user annotation on every iteration, and that is what I use in the traces.

libkineto/src/dynamic_plugin/PluginProfiler.h
briancoutinho: Update the `query()` interface  so a plugin can now provide list of activities it supports.
Configure handler will then check if there is overlap in supported activties.

briancoutinho:
@sraikund16  / @aaronenyeshi  / @sanrise  Gentle nudge, please help with review.

libkineto/include/KinetoDynamicPluginInterface.h
```diff
@@ -0,0 +1,307 @@
...
+
+  // [in/out] Supported activity types.
+  KinetoPlugin_ProfileEventType *pSupportedActivityTypes;
+
+  // [in] Max length of pSupportedActivityTypes
+  size_t supportedActivityTypesMaxLen;
+};
+#define KINETO_PLUGIN_PROFILER_QUERY_PARAMS_UNPADDED_STRUCT_SIZE               \
+  KINETO_PLUGIN_UNPADDED_STRUCT_SIZE(struct KinetoPlugin_ProfilerQuery_Params, \
+                                     profilerNameMaxLen)
```
australopitek: `profilerNameMaxLen` should be changed to `supportedActivityTypesMaxLen`.

briancoutinho:
@sraikund16  Please help with review:)

briancoutinho:
<a href="https://gitme.me/image?url=https%3A%2F%2Fmedia4.giphy.com%2Fmedia%2Fv1.Y2lkPTI5NzJhMmIxejRzdTVxOWQzM3VkdndhdHRjN3dpMnY0eDE0OXFzc2pnMGZqdHkwNSZlcD12MV9naWZzX3NlYXJjaCZjdD1n%2FaZ4sQUpybai5y%2Fgiphy.gif&token=please" data-gitmeme-token="please">please</a>

libkineto/src/dynamic_plugin/PluginProfiler.h
```diff
@@ -0,0 +1,278 @@
...
+#include "KinetoDynamicPluginInterface.h"
+#include "PluginTraceBuilder.h"
+#include "PluginUtils.h"
+
+namespace libkineto {
+
+// This file handles pure C plugin profiler interface and converts to internal
+// profiler interface
+
+class PluginProfilerSession : public IActivityProfilerSession {
```
australopitek, line 19: Below methods from IActivityProfilerSession are not implemented in PluginProfilerSession:
```
  virtual void pushCorrelationId(uint64_t /*id*/) {}
  virtual void popCorrelationId() {}
  virtual void pushUserCorrelationId(uint64_t /*id*/) {}
  virtual void popUserCorrelationId() {}
```
Is this intended and not needed in external plugin, or just a mistake?

libkineto/src/dynamic_plugin/PluginUtils.h
```diff
@@ -0,0 +1,60 @@
...
+    return ActivityType::PYTHON_FUNCTION;
+  case KINETO_PLUGIN_PROFILE_EVENT_TYPE_OVERHEAD:
+    return ActivityType::OVERHEAD;
+  case KINETO_PLUGIN_PROFILE_EVENT_TYPE_CUDA_SYNC:
+    return ActivityType::CUDA_SYNC;
+  case KINETO_PLUGIN_PROFILE_EVENT_TYPE_GPU_PM_COUNTER:
+    return ActivityType::GPU_PM_COUNTER;
+  default:
+    // use kernel type as a default
+    return ActivityType::CONCURRENT_KERNEL;
```
australopitek: Some of KinetoPlugin_ProfileEventType are not handled in this function. I think it wasn't updated when you extended enum in https://github.com/pytorch/kineto/pull/1148/commits/06415498b6f4cc77f50814334e38492d007afc94

libkineto/src/dynamic_plugin/PluginProfiler.h
australopitek, line 19: Do you plan to expose toggleCollectionDynamic functionality, so plugin profilers can react to it?

libkineto/include/KinetoDynamicPluginInterface.h
briancoutinho: Thanks for catching this, updating it!

libkineto/src/dynamic_plugin/PluginProfiler.h
briancoutinho: I have not considered adding this to the interface yet because I did not use it in the plugn But for completeness sake let's add this. I can keep it unimplemented on my side.

briancoutinho: Not currently since this is not yet part of the IActivityProfilerSession interface.
However, the structures here can be extended as long as we keep the same order
https://github.com/pytorch/kineto/blob/main/libkineto/include/IActivityProfiler.h

libkineto/src/dynamic_plugin/PluginUtils.h
briancoutinho: Done

briancoutinho:
Hi @malfet curious if we could get some feedback on this PR. guessing folks are busy but is there anyone from the maintainers from Meta who can help import and get this to the next stage. 
My CLA signup just went through today, let me know if there is anything else needed from my side.

libkineto/src/dynamic_plugin/PluginProfiler.h
australopitek: I still think it's not perfect solution. We are currently moving Xpupti profiler to external plugin. `XPU_RUNTIME`, similarly to `CUDA_RUNTIME`, is present in `kCpuTypes` https://github.com/pytorch/pytorch/blob/main/torch/csrc/profiler/kineto_shim.cpp#L20
So if the user calls profiler with `activities=[ProfilerActivity.CPU]`, xpu plugin will be fully loaded and will collect all traces, including kernel activities, memcpy and memset, the same as if user called `activities=[ProfilerActivity.CPU, ProfilerActivity.XPU]`

I would still suggest to pass requested `activity_types` to plugin profiler, e.g. in KinetoPlugin_ProfilerCreate_Params.

libkineto/src/dynamic_plugin/PluginLoader.h
```diff
@@ -0,0 +1,139 @@
...
+  }
+
+  PluginRegistry &pluginRegistry = PluginRegistry::instance();
+  KinetoPlugin_Registry cPluginRegistry = pluginRegistry.toCRegistry();
+
+  for (const auto &libFilePath : libFilePaths) {
+    // Clear error state
+    dlerror();
+
+    void *pHandle = dlopen(libFilePath.c_str(), RTLD_LAZY);
```
australopitek: Do you plan to add support for Windows?

briancoutinho: @australopitek  Not planning to in this PR but it should be an easy addition on top of this PR. Personally, I don't have a Windows setup to test things right now.

briancoutinho:
@australopitek  Still waiting to hear back from the maintainers at Meta. Though Kineto is part of PyTorch Foundation now, the codebase is not github first from what I know, it needs to be imported by Meta to push through. Hoping the maintainers at Meta get back on this.

@australopitek  If you describe your use-case for this interface too that might help make a stronger case. Thanks!

australopitek:
> @australopitek Still waiting to hear back from the maintainers at Meta. Though Kineto is part of PyTorch Foundation now, the codebase is not github first from what I know, it needs to be imported by Meta to push through. Hoping the maintainers at Meta get back on this.
> 
> @australopitek If you describe your use-case for this interface too that might help make a stronger case. Thanks!

@briancoutinho In response to my comment https://github.com/pytorch/kineto/pull/1148/files#r2454233130, we want to move XpuptiProfiler to external Intel repo, so it's independent from Kineto repository. Currently XpuptiProfiler is responsible for various traces (runtime, kernel, memset, etc.).
To maintain full functionality, plugin profiler should know about requested activities, otherwise it will always report all kinds of traces, even if only runtime is requested.

libkineto/include/KinetoDynamicPluginInterface.h
sraikund16: @briancoutinho Agreed, this should be easier to scale

sraikund16:
Taking a look, for some reason I was not getting notifications on this PR.

briancoutinho:
@australopitek Updated the start trace API to pass enabled Activity types, this should help configure plugins 👍 Also see the DynamicPluginTests.cpp, that verifies this works

Updates
1. Updated the start trace API to pass enabled Activity types
2. Exposed addDeviceInfo() to the plugin's trace builder API, so plugins can create new device rows.
3. Rebase to main

briancoutinho:
@sraikund16  Thanks for checking in 👍  
You will probably need to add new TARGET files.. 
Let me know if I can help with the format changes too, or please feel free to push changes to the PR :)

libkineto/include/KinetoDynamicPluginInterface.h
```diff
@@ -220,6 +239,12 @@ struct KinetoPlugin_ProfilerStart_Params {
   // Always set to KINETO_PLUGIN_PROFILER_START_PARAMS_UNPADDED_STRUCT_SIZE
   size_t unpaddedStructSize;
 
+  // [in] Enabled activity types.
+  KinetoPlugin_ProfileEventType *pEnabledActivityTypes;
+
+  // [in] Max length of pEnabledActivityTypes
+  size_t enabledActivityTypesMaxLen;
+
```
australopitek: I wonder if KinetoPlugin_ProfilerCreate_Params isn't a better place for these params. Both Cupti profiler and Xpupti profiler enable activities in configure phase, which happen before start https://github.com/pytorch/kineto/blob/main/libkineto/src/CuptiActivityProfiler.cpp#L1118

libkineto/libkineto_defs.bzl
```diff
@@ -70,6 +70,9 @@ def get_libkineto_cpu_only_srcs(with_api = True):
         "src/init.cpp",
         "src/output_csv.cpp",
         "src/output_json.cpp",
+        "src/dynamic_plugin/PluginLoader.h",
```
sraikund16, line 74: why are we not adding pluginutils.h?

libkineto/src/dynamic_plugin/PluginUtils.h
```diff
@@ -0,0 +1,150 @@
...
+namespace libkineto {
+
+// Utility functions for dynamic plugins
+
+static inline ActivityType convertToActivityType(
+    KinetoPlugin_ProfileEventType type) {
+  switch (type) {
+    case KINETO_PLUGIN_PROFILE_EVENT_TYPE_CPU_OP:
+      return ActivityType::CPU_OP;
+    case KINETO_PLUGIN_PROFILE_EVENT_TYPE_USER_ANNOTATION:
```
malfet, line 19: Could this code be codegenerated? (Or just use one enum)

libkineto/src/dynamic_plugin/PluginTraceBuilder.h
```diff
@@ -0,0 +1,270 @@
...
+// need to translate a few things around
+
+class PluginTraceBuilder {
+ public:
+  PluginTraceBuilder(TraceSpan span) {
+    buffer_ = std::make_unique<CpuTraceBuffer>();
+    buffer_->span = span;
+  }
+
+  int addEvent(const KinetoPlugin_ProfileEvent* pProfileEvent) {
```
malfet, line 29: Would be good to have some explanation on what `addEvent` return code is

libkineto/src/dynamic_plugin/PluginLoader.h
```diff
@@ -0,0 +1,138 @@
...
+               << strerror(errno);
+
+    return;
+  }
+
+  std::vector<std::string> libFilePaths;
+  try {
+    for (const auto& entry :
+         std::filesystem::directory_iterator(pPluginLibDirPath)) {
+      if (entry.is_regular_file() && entry.path().extension() == ".so") {
```
malfet: Looks like it constraints this to Linux, is that by design

libkineto/src/dynamic_plugin/PluginLoader.h
```diff
@@ -0,0 +1,138 @@
...
+
+  if (pPluginLibDirPath == nullptr) {
+    LOG(VERBOSE) << "Environment variable " << pPluginLibDirPathEnvVar
+                 << " not set";
+
+    return;
+  }
+
+  if (unsetenv(pPluginLibDirPathEnvVar) == -1) {
+    LOG(ERROR) << "Failed to unset environment variable "
```
malfet: This wouldn't work for distributed invocations, would it? (i.e. if this is called from parent process, that calls `fork()` and calls the same method from child processes

libkineto/src/dynamic_plugin/PluginLoader.h
```diff
@@ -0,0 +1,138 @@
...
+    int (*pfxRegister)(const KinetoPlugin_Registry* pRegistry) =
+        reinterpret_cast<int (*)(const KinetoPlugin_Registry* pRegistry)>(
+            dlsym(pHandle, "KinetoPlugin_register"));
+
+    if (pfxRegister == nullptr) {
+      char* pError = dlerror();
+      LOG(VERBOSE) << "Failed to find symbol KinetoPlugin_register() from "
+                   << libFilePath << " at dlsym() with error " << pError;
+
+      continue;
```
malfet, line 123: Don't one needs to `dlclose()` library here?

libkineto/src/dynamic_plugin/PluginProfiler.h
```diff
@@ -0,0 +1,494 @@
...
+    for (size_t i = 0; i < supportedActivityTypes.size(); i++) {
+      if (supportedActivityTypes[i] !=
+          KINETO_PLUGIN_PROFILE_EVENT_TYPE_INVALID) {
+        supportedActivities_.insert(
+            convertToActivityType(supportedActivityTypes[i]));
+      }
+    }
+  }
+
+  ~PluginProfiler() {}
```
malfet:
```suggestion
  ~PluginProfiler() = default;
```

libkineto/include/KinetoDynamicPluginInterface.h
```diff
@@ -0,0 +1,423 @@
...
+  (size_t)(offsetof(struct_type, last_field_name) +                      \
+           sizeof(((struct_type*)1000)->last_field_name))
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+// Event types that can be supported via plugin. Please sync
+// with ActivityType.h if you add new event types.
+enum KinetoPlugin_ProfileEventType {
```
malfet, line 22: Should we use `enum class`?

aaronenyeshi (approved):

libkineto/include/KinetoDynamicPluginInterface.h
briancoutinho: @australopitek Got it, yes I can move it to Create() makes sense. (TODO)

libkineto/libkineto_defs.bzl
briancoutinho: Looks like I missed this file, somehow the build works, will fix it (TODO)

libkineto/include/KinetoDynamicPluginInterface.h
briancoutinho: Actually we need to create a C interface because of the handling of C++ ABI by different compilers can be an issue.
So the whole dynamic plugin header is exporting a pure C view of the orginal C++ interface.
Hence we cannot use enum classes.

libkineto/src/dynamic_plugin/PluginLoader.h
briancoutinho: Ah, that makes sense, Let me remove this (TODO)

libkineto/src/dynamic_plugin/PluginLoader.h
briancoutinho: No we should try to support Windows. will update this extension to an OR to support dlls. 
We still need to add some windows specific loading to support Windows, that can be in a followup 
For my case I only need Linux but will be good to have the API support Windows too.

libkineto/src/dynamic_plugin/PluginProfiler.h
briancoutinho: TODO

libkineto/src/dynamic_plugin/PluginTraceBuilder.h
briancoutinho: TODO

libkineto/src/dynamic_plugin/PluginUtils.h
briancoutinho: Yea something that generates the 1:1 translation could be helpful. Not sure if there is some simpler C++ solution for this. Still need to use C enums here.

sraikund16:
@briancoutinho I think we should hold off on merging this PR. PyTorch in general uses static linking over dynamic to maintain portability/release stability. This would change that paradigm for Kineto. Also the PR itself adds a lot of boilerplate code based on the CUPTI Activity Profiler, which we may consider redesigning for more general use.

briancoutinho:
@sraikund16  
> PyTorch in general uses static linking over dynamic to maintain portability/release stability

I understand that this is a shift from the convention in PyTorch. My intent was to try to dogfood this dynamic approach gradually rather than make an immediate change. The changes are relatively independent and should not impact the backbone Kineto code. 
* Open to ways we can make this feature more restrictive like a runtime flag or build time flag.
* The source based approach works well for mono-repo style development but if users are picking fixed Torch versions it is quite hard to always upstream and open source profiling hooks. We thought this would make PyTorch Profiler more maintainable and extensible.
* The versioning strategy is also fairly simple and is leveraged in TensorFlow/Jax profiler ([reference](https://github.com/annarev/community/blob/09e231324c3b2b1f653c3f385fd2120edd460815/rfcs/20200612-stream-executor-c-api/C_API_versioning_strategy.md#backwards-compatibility)) The size of the struct is used to indicate the version, and avoid out of bounds accesses.

Please let me know if the above make sense, and share more on the concerns on portability/stability. Happy to discuss in a call as well.

 > Also the PR itself adds a lot of boilerplate code based on the CUPTI Activity Profiler, which we may consider redesigning for more general use.
 
Oh my changes are using the generic ActivityProfilerAPI which is not CUPTI specific afaik. The activity types have some historical attachment to CUPTI which we can evolve for sure.

yisitu:
@sraikund16 

>PyTorch in general uses static linking over dynamic to maintain portability/release stability.

Portability and release stability of the core PyTorch components will not be impacted by this PR since they are unmodified. 

>the PR itself adds a lot of boilerplate code based on the CUPTI Activity Profiler, which we may consider redesigning for more general use.

Would you mind clarifying -- which are the boilerplate code that based on CUPTI?  The PR uses general plugin and profiler concepts that are not specific to CUPTI; that would not impede the possibility of redesigning for general use in the future.

sraikund16:
> @sraikund16
> 
> > PyTorch in general uses static linking over dynamic to maintain portability/release stability.
> 
> Portability and release stability of the core PyTorch components will not be impacted by this PR since they are unmodified.
> 
> > the PR itself adds a lot of boilerplate code based on the CUPTI Activity Profiler, which we may consider redesigning for more general use.
> 
> Would you mind clarifying -- which are the boilerplate code that based on CUPTI? The PR uses general plugin and profiler concepts that are not specific to CUPTI; that would not impede the possibility of redesigning for general use in the future.

@briancoutinho @yisitu 

We understand that the backbone of Kineto is unchanged and that this is essentially parallel to it. However, we have historically suffered from trying to support to many side paths so we are trying to consolidate Kineto a bit more. Currently, Kineto is not very modular (as you can see from all the ifdefs in CuptiActivityProfiler) and we are accessing how to redesign to make it more generic. I imagine that this rework may break this implementation which is why I was suggesting we hold off until we sort that out.
 
In terms of the boilerplate comment, I was not saying that based on CUPTI but rather the CuptiActivityProfiler.cpp/h implementation. For example, a lot of the enums regarding activity type are essentially copied over.

briancoutinho:
@sraikund16 

I see. To summarize, **there is desire to refactor kineto to make it more modular. The concern is that adding this dynamic plugin capability will be an additional burden to the refactor**. I agree we need to modularize kineto! 

For the this PR the question boils down to whether the `IActivityProfiler` interface is subject to change. When I introduced this interface years ago the goal was indeed to make Kineto more modular. However, we did move CUPTI profile to this interface though. If the interface does not change the dynamic plugin I have shared here will just be another type of profiler plugin, and should be transparent to the refactor. 

> For example, a lot of the enums regarding activity type are essentially copied over.

Yes that is a good point. The enums need to be revisited, and this API would expose them making backward compatability harder. 

Please share with us a **timeline for this re-design** we can wait for it. I will address comments above in a bit to get it to good state.
If the timeline is several months, one suggestion is to make this a **build time disabled feature** and also comment that **the API is unstable**. Build off by default will not be a burden for refactor. I can publish updates to the plugin interface to match your redesign..

briancoutinho:
@malfet Thanks for sharing the feedback.
>Is it possible to do the changes in several increments? (For example, it feels fine to have an API call that disables Kineto initialization, but doing it via environment variable with some pre-defined name makes it less traceable) Same about having an environment variable that just scans the folder and loads extensions, which would be immediately flagged during next security review as fundamentally unsafe

Yes, I can break this up into 2-3 PRs and use ghstack. That will help with reviews as wel.
Let me make some final changes based on the feedback you and others have shared so they do not get lost. We can then switch to the broken down PRs.

>And pardon my shallow knowledge of the background behind this change, but it seems like your goal is to expose GPU_PM_COUNTER activity. If this is the case, why can't it be done via the existing Kineto/CUPTI integration mechanism?

Kind of, the goal is to enable a newer profiling technology that has lower overhead than CUPTI. It also happens to support counter streaming, so I added `GPU_PM_COUNTER` event type.

The counter events part was necessary as Kineto does not currently handle counter events in the output. The Chrome Trace format does [support counter events](https://docs.google.com/document/d/1CvAClvFfyA5R-PhYUmn5OOQtYMH4h6I0nSsKchNAySU/preview?tab=t.0#heading=h.msg3086636uq). Also, counters can be used for other device types too.

libkineto/include/KinetoDynamicPluginInterface.h
briancoutinho: Moved

briancoutinho:
Updates
- Moved enabledActivityTypes to create() from start() part of dynamic API @australopitek 
- Missing file in bzl fixed
- Plugin loader comments.
- Better documentation for return value `addEvent()` and others

briancoutinho:
Will split these changes out 👍

libkineto/include/KinetoDynamicPluginInterface.h
```diff
@@ -0,0 +1,448 @@
...
+  KINETO_PLUGIN_PROFILE_EVENT_TYPE_COLLECTIVE_COMM, // collective communication
+  KINETO_PLUGIN_PROFILE_EVENT_TYPE_GPU_PM_COUNTER, // GPU performance monitoring
+                                                   // counter
+
+  // PRIVATEUSE1 Activity types are used for custom backends.
+  // The corresponding device type is `DeviceType::PrivateUse1` in PyTorch.
+  KINETO_PLUGIN_PROFILE_EVENT_TYPE_PRIVATEUSE1_RUNTIME, // host side privateUse1
+                                                        // runtime events
+  KINETO_PLUGIN_PROFILE_EVENT_TYPE_PRIVATEUSE1_DRIVER, // host side privateUse1
+                                                       // driver events
```
KarhouTam, line 63: Hi, @briancoutinho. I am new to kineto, so this might be a dumb question. Is this dynamic plugin also friendly to the profiler for those out-of-tree accelerator backends (integrated via privateuse1 key)? For example, is it easy to let those backend developers add their own kineto activity types?

briancoutinho:
@*KarhouTam it is a great question 👍*

*Yes and no. Yes, that the dynamic approach can be leveraged for out of
tree accelerators.*

However, we do not want to add new activity types for every accelerator.
Instead we will likely define generic activity/event types like RUNTIME,
KERNEL, Memcpy etc. We can use aliasing for existing device specific types
so they continue working but we should eventually deprecate them.
In this PR I had it the old way but will split this PR into 2-3 PRs. Will
add the generic event types in a future change, soon in early December.

PR #1151, opened 2025-09-23
[XPUPTI] Add missing `stream` field in a trace.

tsocha:
Currently tools like HTA accept traces which are strictly targeted for CUDA.
This PR is a MVP which allow traces gathered on XPU to be accepted by such tools.
Adapter introduced in this PR will be extended with more capabilities.

### DIFF
--- a/libkineto/include/Config.h
+++ b/libkineto/include/Config.h
@@ -26,6 +26,7 @@ class Config : public AbstractConfig {
   Config& operator=(const Config&) = delete;
   Config(Config&&) = delete;
   Config& operator=(Config&&) = delete;
+  ~Config() override = default;
 
   // Return a full copy including feature config object
   std::unique_ptr<Config> clone() const {


libkineto/src/plugin/xpupti/XpuptiActivityProfiler.cpp
```diff
@@ -188,11 +188,23 @@ const std::set<ActivityType>& XPUActivityProfiler::availableActivities() const {
...
+    return env != nullptr && strcmp(env, "1") == 0;
+  }
+}
+
 std::unique_ptr<libkineto::IActivityProfilerSession>
 XPUActivityProfiler::configure(
     const std::set<ActivityType>& activity_types,
     const libkineto::Config& config) {
-  return std::make_unique<XpuptiActivityProfilerSession>(
+  const auto experimental_cuda_format_enabled = is_env_set("XPUPTI_USE_EXPERIMENTAL_CUDA_FORMAT");
```
gujinghui: 1. Why call it CUDA format? It's defined by CUDA or HTA?
2. What's the difference btw this format and original one? Looks like no more detail in this PR.
3. Why have to introduce a env var? Generally, it's not a good idea to depend on a env var...

tsocha: Ad. 1. It is the first change in a series. The goal is to produce our traces in a way that it looks like CUDA trace.
Ad. 2. The first step is to produce a `stream` param. I plan to add other changes like renaming some operators in our trace to match CUDA's naming.
Ad. 3. I don't like the env var either, but I didn't get a better idea how to parametrize it. Any feedback how to do it better is welcome.

gujinghui: > The goal is to produce our traces in a way that it looks like CUDA trace.

I have no idea about the cuda trace. Could you show related links or examples? Thanks.

> I don't like the env var either, but I didn't get a better idea how to parametrize it. Any feedback how to do it better is welcome.

Per this PR, I cannot understand your motivation, as there's no real change so far. I guess, finally, you will only keep one path, right? Could you show your design here?

tsocha: > > The goal is to produce our traces in a way that it looks like CUDA trace.
> 
> I have no idea about the cuda trace. Could you show related links or examples? Thanks.
> 
Some examples of differences:
1. CUDA produce a field named `stream` but XPU produce `sycl queue` and `l0 queue`.
    - This PR add `stream` field.
2. Operators for queue start and end and named differently:
    - `urEnqueueUSMMemcpy` vs `cudaLaunchKernelExC` - this is the next step to do.
3. For distributed communication kernels it have `nccl` prefix in CUDA.

> > I don't like the env var either, but I didn't get a better idea how to parametrize it. Any feedback how to do it better is welcome.
> 
> Per this PR, I cannot understand your motivation, as there's no real change so far. I guess, finally, you will only keep one path, right? Could you show your design here?

My motivation is to produce a trace which could be accepted in tools which does not have XPU support and expect traces produced by CUDA.
This PR introduce one simple change to add a missing field `stream` but I expect more changes.
Some of these changes can replace operators name or categories.
I don't want to break existing solution.
That's a reason why I have decided to create a separate class instead of modifying already existing one.

For some time there will be 2 paths but in the future we can remove/merge one of them.

gujinghui: @tsocha   These are not cuda format. These are API/interface names of cuda software stack, like `stream`, `cudaLaunchKernelExC`, `nccl`. I don't think we must show identical API names in the trace or UI. And we even cannot do that, as we cannot guarantee we are always able to find APIs in XPU software stack with same semantics.
If needed, we have to do some change in kineto or HTA to fit our requirement, instead of totally following CUDA for all, right?

> My motivation is to produce a trace which could be accepted in tools which does not have XPU support and expect traces produced by CUDA.

Instead, we should improve the tool to support XPU. If you're talking about HTA, we should submit PR to HTA if needed. But I assume in most case HTA should be device-agnostic?

tsocha: @gujinghui
I see why you don't want to introduce this adapter into kineto, I will focus on changes in HTA.

So another question:
Do you agree to produce a `stream` param in our traces?
HTA requires it, but we produce `l0_queue` and `sycl_queue`.
We can just reuse `sycl_queue` as our `stream`.
This is a minimal change and will fix many issues here and in potential other tools.

I can see that `stream` field is used by `CUPTI`, `ROCTRACER` and `AIUPTI`.

PS in pytorch code `stream` is already used in `xpu` context.
example:
```
stream1 = torch.xpu.current_stream()
stream_tmp = torch.xpu.Stream(torch.xpu.current_device())
stream2 = torch.xpu.Stream(torch.xpu.current_device())
```

tsocha: @gujinghui could you look at it? ☝️

gujinghui: Could you show the stream user case in HTA?
I believe it depends on which level of runtime concept we are going to show to user in HTA.
If HTA needs a PyTorch Stream, we can follow to align with CUDA path.
But if HTA needs a CUDA Stream specifically, we'd better expose sycl queue or L0 queue, instead.

tsocha: `stream` is commonly used in HTA: 110 hits in 18 files excluding examples and tests.

It's mainly used to group events and filter if kernel is on GPU.

In my opinion its abstract level is very similar to sycl_queue, it represent a series of related operators.
Comparing traces collected on CUDA and XPU seems to confirm it.

We had an ugly solution in HTA where we search for `sycl queue` and then remove `stream` and rename `sycl queue` -> `stream` after that, but it's an ugly solution in my opinion.

Since all other PTIs use `stream` field(`AIUPTI` just use 1 as its value) I believe we can add it into our `XPUPTI`.

PR #1153, opened 2025-09-29
Initial draft

moksiuc:


libkineto/src/plugin/xpupti/XpuptiActivityApi.cpp
```diff
@@ -24,6 +25,22 @@ XpuptiActivityApi& XpuptiActivityApi::singleton() {
   return instance;
 }
 
+XpuptiActivityApi::XpuptiActivityApi() {
+#ifdef HAS_XPUPTI
+  XPUPTI_CALL(ptiMetricsGetDevices(nullptr, &deviceCount_));
```
tsocha, line 28: When I see a `nullptr` I'm getting suspicious, why do you need it here?

libkineto/src/plugin/xpupti/XpuptiActivityApi.cpp
```diff
@@ -24,6 +25,22 @@ XpuptiActivityApi& XpuptiActivityApi::singleton() {
...
+
+  if (deviceCount_ > 0) {
+    auto devices = std::make_unique<pti_device_properties_t[]>(deviceCount_);
+    XPUPTI_CALL(ptiMetricsGetDevices(devices.get(), &deviceCount_));
+
+    devicesHandles_ = std::make_unique<pti_device_handle_t[]>(deviceCount_);
+    for (uint32_t i = 0; i < deviceCount_; ++i) {
+      devicesHandles_[i] = devices[i]._handle;
+    }
+  }
```
tsocha, line 38: `devices` can be an `std::vector`.

libkineto/src/plugin/xpupti/XpuptiActivityApi.cpp
```diff
@@ -184,117 +196,280 @@ void XpuptiActivityApi::bufferCompleted(
...
+
+void XpuptiActivityApi::enableScopeProfiler(const Config& cfg) {
+#ifdef HAS_XPUPTI
+#if PTI_VERSION_AT_LEAST(0, 14)
+  if (deviceCount_ == 0) {
+    throw std::runtime_error("No XPU devices available");
+  }
+
+  if (!scopeHandleOpt_) {
+    XPUPTI_CALL(ptiMetricsScopeEnable(&*scopeHandleOpt_));
```
tsocha: This look weird, you de-reference a pointer and then get it's address?
Why you can't just use a pointer here?

libkineto/src/plugin/xpupti/XpuptiActivityApi.cpp
```diff
@@ -184,117 +196,280 @@ void XpuptiActivityApi::bufferCompleted(
...
-      1, pti_api_group_id::PTI_API_GROUP_SYCL, urEnqueueMemBufferWrite_id));
-  XPUPTI_CALL(ptiViewEnableRuntimeApi(
-      1, pti_api_group_id::PTI_API_GROUP_SYCL, urEnqueueMemBufferCopy_id));
-  XPUPTI_CALL(ptiViewEnableRuntimeApi(
-      1, pti_api_group_id::PTI_API_GROUP_SYCL, urUSMHostAlloc_id));
-  XPUPTI_CALL(ptiViewEnableRuntimeApi(
-      1, pti_api_group_id::PTI_API_GROUP_SYCL, urUSMSharedAlloc_id));
-  XPUPTI_CALL(ptiViewEnableRuntimeApi(
-      1, pti_api_group_id::PTI_API_GROUP_SYCL, urUSMDeviceAlloc_id));
+  std::array<pti_api_id_runtime_sycl, 14> specifcRuntimeAPIsTracing = {
```
PawelSwider2000: Cant' this be constexpr?

libkineto/src/plugin/xpupti/XpuptiActivityApi.cpp
```diff
@@ -184,117 +196,280 @@ void XpuptiActivityApi::bufferCompleted(
...
 #endif
 }
 
+static auto safeCallPtiMetricScopeCalculateMetrics(
+    pti_scope_collection_handle_t scopeHandle,
+    void* bufferData,
+    size_t& recordsCount) {
+  pti_metrics_scope_record_t* records = nullptr;
+  XPUPTI_CALL(ptiMetricScopeCalculateMetrics(
+      scopeHandle, bufferData, &records, &recordsCount));
```
tsocha: Just to be sure, `ptiMetricScopeCalculateMetrics` will set a valid pointer into `records`?

libkineto/src/plugin/xpupti/XpuptiActivityApi.cpp
```diff
@@ -184,117 +196,280 @@ void XpuptiActivityApi::bufferCompleted(
...
+      records, recordsDeleter);
+}
+
+static auto safeCallMetricScopeGetDisplayInfo(
+    pti_scope_collection_handle_t scopeHandle,
+    pti_metrics_scope_record_t* record,
+    uint32_t& infoCount) {
+  pti_metric_scope_display_info_t* displayInfo = nullptr;
+  XPUPTI_CALL(ptiMetricScopeGetDisplayInfo(
+      scopeHandle, record, &displayInfo, &infoCount));
```
tsocha: As above.

libkineto/src/plugin/xpupti/XpuptiActivityApi.h
```diff
@@ -35,40 +33,47 @@ class XpuptiActivityApi {
...
 
  private:
-  int maxGpuBufferCount_{0};
   XpuptiActivityBufferMap allocatedGpuTraceBuffers_;
   std::unique_ptr<XpuptiActivityBufferMap> readyGpuTraceBuffers_;
   std::mutex mutex_;
-  std::atomic<uint32_t> tracingEnabled_{0};
   bool externalCorrelationEnabled_{false};
+  std::optional<pti_scope_collection_handle_t> scopeHandleOpt_;
+  std::unique_ptr<pti_device_handle_t[]> devicesHandles_;
```
tsocha, line 91: Do we need to store it that way?
`std::vector` could be more convenient.

libkineto/src/plugin/xpupti/XpuptiActivityApi.h
```diff
@@ -35,40 +33,47 @@ class XpuptiActivityApi {
...
+  uint32_t deviceCount_{0};
+  std::vector<const char*> activitiesXpuptiMetricsNames_;
```
tsocha: can we use `std::string` here?

libkineto/src/plugin/xpupti/XpuptiActivityHandlers.cpp
```diff
@@ -329,42 +250,34 @@ void XpuptiActivityProfilerSession::handleOverheadActivity(
...
     const pti_view_record_base* record,
-    ActivityLogger* logger) {
+    ActivityLogger& logger) {
   switch (record->_view_kind) {
     case PTI_VIEW_EXTERNAL_CORRELATION:
       handleCorrelationActivity(
           reinterpret_cast<const pti_view_record_external_correlation*>(
               record));
       break;
-#if PTI_VERSION_MAJOR > 0 || PTI_VERSION_MINOR > 10
```
PawelSwider2000, line 347: Are we not missing this? or it is handled in different place?

libkineto/src/plugin/xpupti/XpuptiActivityHandlers.cpp
```diff
@@ -1,28 +1,33 @@
...
 }
 
 void XpuptiActivityProfilerSession::checkTimestampOrder(
     const ITraceActivity* act1) {
-  const auto& it = correlatedPtiActivities_.find(act1->correlationId());
-  if (it == correlatedPtiActivities_.end()) {
-    correlatedPtiActivities_.insert({act1->correlationId(), act1});
+  auto insert_result =
+      correlatedPtiActivities_.insert({act1->correlationId(), act1});
+  if (insert_result.second) {
```
tsocha: This logic is hard to read.
What is inside insert_result?
Why we return if `insert_result.second` is true?

libkineto/src/plugin/xpupti/XpuptiActivityHandlers.cpp
```diff
@@ -39,20 +44,20 @@ void XpuptiActivityProfilerSession::checkTimestampOrder(
...
+    err_msg += " (" + std::to_string(act->timestamp());
     err_msg += " < " + std::to_string(captureWindowStartTime_);
-    err_msg += " or " + std::to_string(act.timestamp() + act.duration());
+    err_msg += " or " + std::to_string(act->timestamp() + act->duration());
     err_msg += " > " + std::to_string(captureWindowEndTime_);
     errors_.push_back(err_msg);
   }
-  return out_of_range;
+  return out_of_scope;
 }
```
tsocha, line 67: Switch from reference to pointer looks unnecessary.

libkineto/src/plugin/xpupti/XpuptiActivityHandlers.cpp
```diff
@@ -74,123 +79,38 @@ inline std::string handleToHexString(ze_handle_type handle) {
...
-  auto it = std::find(
-      sycl_queue_pool_.begin(), sycl_queue_pool_.end(), sycl_queue_id);
-  if (it != sycl_queue_pool_.end()) {
-    return std::distance(sycl_queue_pool_.begin(), it);
-  }
-  sycl_queue_pool_.push_back(sycl_queue_id);
-  return sycl_queue_pool_.size() - 1;
+  auto insert_result =
+      sycl_queue_pool_.insert({sycl_queue_id, sycl_queue_pool_.size()});
+  return insert_result.first->second;
```
tsocha: It's hard to read, you can use structured bindings and helper variables to document this code.

libkineto/src/plugin/xpupti/XpuptiActivityHandlers.cpp
```diff
@@ -378,4 +291,106 @@ void XpuptiActivityProfilerSession::handlePtiActivity(
   }
 }
 
+static std::string PtiValueToStr(
+    pti_metric_value_type valueType,
+    const pti_value_t& value) {
+  switch (valueType) {
+#define CASE(T, FIELD)            \
```
PawelSwider2000: This define here for me makes this function harder to read, simple switch case will be shorted and much easier to read

libkineto/src/plugin/xpupti/XpuptiActivityHandlers.cpp
```diff
@@ -211,105 +131,106 @@ inline std::string bandwidth(pti_view_memory_record_type* activity) {
   return duration == 0 ? "\"N/A\"" : fmt::format("{}", bytes * 1.0 / duration);
 }
 
-void XpuptiActivityProfilerSession::handleMemcpyActivity(
-    const pti_view_record_memory_copy* activity,
-    ActivityLogger* logger) {
+template <class pti_view_memory_record_type>
+void XpuptiActivityProfilerSession::handleRuntimeKernelMemcpyMemsetActivities(
```
tsocha, line 141: I'm not a fan of this huge function, I think that composition will work better here. Current version have many ifs and elses.

libkineto/src/plugin/xpupti/XpuptiActivityHandlers.cpp
```diff
@@ -329,42 +250,34 @@ void XpuptiActivityProfilerSession::handleOverheadActivity(
           activity->_overhead_duration_ns / overhead_activity->duration()));
   overhead_activity->addMetadata("overhead count", activity->_overhead_count);
 
-  if (outOfRange(*overhead_activity)) {
-    return;
+  if (!outOfScope(overhead_activity.get())) {
+    overhead_activity->log(logger);
   }
```
tsocha, line 269: I prefer the previous `if` condition.
I don't like negative ifs.

libkineto/src/plugin/xpupti/XpuptiActivityHandlers.cpp
```diff
@@ -378,4 +291,106 @@ void XpuptiActivityProfilerSession::handlePtiActivity(
...
+#define CASE(T, FIELD)            \
+  case PTI_METRIC_VALUE_TYPE_##T: \
+    return std::to_string(value.FIELD)
+
+    CASE(UINT32, ui32);
+    CASE(UINT64, ui64);
+    CASE(FLOAT32, fp32);
+    CASE(FLOAT64, fp64);
+
+#undef CASE
```
tsocha: this switch is so small that I doubt this macro is worth to add.

libkineto/src/plugin/xpupti/XpuptiActivityProfiler.cpp
```diff
@@ -64,24 +81,37 @@ void XpuptiActivityProfilerSession::stop() {
 void XpuptiActivityProfilerSession::toggleCollectionDynamic(const bool enable) {
   if (enable) {
     xpti_.enableXpuptiActivities(activity_types_);
+    if (scopeProfilerEnabled_) {
+      xpti_.startScopeActivity();
+    }
   } else {
+    if (scopeProfilerEnabled_) {
+      xpti_.stopScopeActivity();
+    }
```
tsocha, line 96: Why do you need to check if it's enabled?
Can't you just enable/disable it with no checking?

libkineto/src/plugin/xpupti/XpuptiActivityProfiler.cpp
```diff
@@ -176,34 +203,31 @@ const std::set<ActivityType> kXpuTypes{
...
-    int64_t ts_ms,
-    int64_t duration_ms,
-    const std::set<ActivityType>& activity_types,
-    const libkineto::Config& config) {
-  AsyncProfileStartTime_ = ts_ms;
-  AsyncProfileEndTime_ = ts_ms + duration_ms;
+std::unique_ptr<libkineto::IActivityProfilerSession> XPUActivityProfiler::
+    configure(
+        int64_t /*ts_ms*/,
+        int64_t /*duration_ms*/,
```
tsocha: Why not `[[maybe_unused]]`?

libkineto/src/plugin/xpupti/XpuptiActivityProfiler.h
```diff
@@ -53,35 +62,39 @@ class XpuptiActivityProfilerSession
  private:
   void checkTimestampOrder(const ITraceActivity* act1);
   void removeCorrelatedPtiActivities(const ITraceActivity* act1);
-  bool outOfRange(const ITraceActivity& act);
+  bool outOfScope(const ITraceActivity* act);
```
tsocha, line 75: Why switch to pointer?

libkineto/src/plugin/xpupti/XpuptiScopeProfilerConfig.cpp
```diff
@@ -0,0 +1,75 @@
...
+
+// number of scopes affect the size of counter data binary used by
+// the XPUPTI Profiler. these defaults can be tuned
+constexpr int KMaxAutoScopes = 1500; // supports 1500 kernels
+constexpr int KMaxUserScopes = 10; // enable upto 10 sub regions marked by user
+
+constexpr char kXpuptiProfilerMetricsKey[] = "XPUPTI_PROFILER_METRICS";
+constexpr char kXpuptiProfilerPerKernelKey[] =
+    "XPUPTI_PROFILER_ENABLE_PER_KERNEL";
+constexpr char kXpuptiProfilerMaxScopesKey[] = "XPUPTI_PROFILER_MAX_SCOPES";
```
tsocha: Why not `std::string`?

libkineto/src/plugin/xpupti/XpuptiScopeProfilerConfig.h
```diff
@@ -0,0 +1,92 @@
...
+  void printActivityProfilerConfig(std::ostream& s) const override;
+  void setActivityDependentConfig() override {}
+  static void registerFactory();
+
+ protected:
+  AbstractConfig* cloneDerived(AbstractConfig& parent) const override {
+    XpuptiScopeProfilerConfig* clone = new XpuptiScopeProfilerConfig(*this);
+    clone->parent_ = dynamic_cast<Config*>(&parent);
+    return clone;
+  }
```
tsocha, line 65: Why it have to be a raw pointer?

libkineto/src/plugin/xpupti/XpuptiActivityApi.cpp
moksiuc: It is API requirement

libkineto/src/plugin/xpupti/XpuptiActivityApi.cpp
moksiuc: Pti API requires raw pointer

libkineto/src/plugin/xpupti/XpuptiActivityApi.cpp
moksiuc: It is address of contents of optional

libkineto/src/plugin/xpupti/XpuptiActivityApi.cpp
moksiuc: Yes or leave nullptr. But then recordsCount would stay 0.

libkineto/src/plugin/xpupti/XpuptiActivityProfiler.h
moksiuc: To be consistent with other API using ITraceActivity

libkineto/src/plugin/xpupti/XpuptiActivityApi.cpp
tsocha: But `devices` is used only to feed `devicesHandles_`

libkineto/src/plugin/xpupti/XpuptiScopeProfilerConfig.cpp
moksiuc: What for, it is constant character literal.

libkineto/src/plugin/xpupti/XpuptiScopeProfilerConfig.h
moksiuc: What could be an alternative here ?

libkineto/src/plugin/xpupti/XpuptiActivityHandlers.cpp
moksiuc: It is more readable as you see dependency between type and field without going deep into the implementation.
It is like extracting some code into separate function.

libkineto/src/plugin/xpupti/XpuptiActivityApi.cpp
moksiuc: Pti API is C-API.

libkineto/src/plugin/xpupti/XpuptiActivityApi.cpp
moksiuc: I don't see advantage of using vector and accessing .data() from it as raw pointer

libkineto/src/plugin/xpupti/XpuptiActivityApi.cpp
moksiuc: Address is necessary as it is C-way of passing argument by reference

libkineto/src/plugin/xpupti/XpuptiActivityHandlers.cpp
moksiuc: It seems I've missed it

libkineto/src/plugin/xpupti/XpuptiActivityApi.cpp
tsocha: what about?
It return a reference and will be more descriptive.
```suggestion
    XPUPTI_CALL(ptiMetricsScopeEnable(&scopeHandleOpt_.value()));
```

moksiuc: *opt is exactly the same as opt.value(), has not disambiguity and is shorter

libkineto/src/plugin/xpupti/XpuptiActivityApi.cpp
moksiuc: I'll add if it can

libkineto/src/plugin/xpupti/XpuptiScopeProfilerConfig.cpp
tsocha: if else block from line 36 will be able to use `==` operator instead of compare.

libkineto/src/plugin/xpupti/XpuptiActivityApi.cpp
tsocha: It is the same and it's the reason why I suggests that.
We won't print this code but we will read it.
`&*` looks weird and `.value()` shows that we want an address of it.

libkineto/src/plugin/xpupti/XpuptiActivityProfiler.cpp
moksiuc: I'll try this

libkineto/src/plugin/xpupti/XpuptiActivityHandlers.cpp
tsocha: Resolved in another conversation.

libkineto/src/plugin/xpupti/XpuptiActivityHandlers.cpp
tsocha: You could also use `std::unordered_map` for it.

libkineto/src/plugin/xpupti/XpuptiScopeProfilerConfig.h
tsocha: Resolved offline.

libkineto/src/plugin/xpupti/XpuptiActivityHandlers.cpp
moksiuc: How ?
Depending on the type you have to access one of the fields of union

libkineto/src/plugin/xpupti/XpuptiActivityApi.cpp
moksiuc: No, &* is also straightforward as address of taking value of something

libkineto/src/plugin/xpupti/XpuptiScopeProfilerConfig.cpp
moksiuc: std::string has additional memory allocated not necessary for literals
The comparison would however be actually simpler. I'll try with string_view.

libkineto/src/plugin/xpupti/XpuptiActivityHandlers.cpp
tsocha: I see... I didn't realize it's an union.
I have thought that you take `PTI_METRIC_VALUE_TYPE_*` and return a predefined string...
Like `PTI_METRIC_VALUE_TYPE_UINT32` -> `ui32`.

This macro is hard to read.

moksiuc: You don't need to go into the detail of the macro.
Its usage shows relation between enum and field of union.

libkineto/src/plugin/xpupti/XpuptiActivityProfiler.cpp
```diff
@@ -176,34 +203,31 @@ const std::set<ActivityType> kXpuTypes{
     ActivityType::XPU_RUNTIME,
     ActivityType::EXTERNAL_CORRELATION,
     ActivityType::OVERHEAD,
+#if PTI_VERSION_AT_LEAST(0, 14)
+    ActivityType::XPU_SCOPE_PROFILER,
+#endif
```
australopitek: kXpuTypes are not used anywhere. Valid place for it is kineto_shim.cpp.

libkineto/src/plugin/xpupti/XpuptiActivityApi.cpp
```diff
@@ -9,12 +9,8 @@
 #include "XpuptiActivityApi.h"
 #include "XpuptiScopeProfilerConfig.h"
 
-#include <assert.h>
-#include <chrono>
```
tsocha, line 12: You have removed `chrono` include but it's used in line 112.

libkineto/src/plugin/xpupti/XpuptiActivityBuffer.h
```diff
@@ -1,9 +1,16 @@
...
+ * This source code is licensed under the BSD-style license found in the
+ * LICENSE file in the root directory of this source tree.
+ */
 
-#include "XpuptiProfilerMacros.h"
+#pragma once
 
-#include <assert.h>
-#include <stdlib.h>
+#include <stddef.h>
```
tsocha:
```suggestion
#include <cstddef>
```

libkineto/src/plugin/xpupti/XpuptiActivityHandlers.cpp
tsocha: You can use structural bindings to name `first` and `second`.

libkineto/src/plugin/xpupti/XpuptiActivityHandlers.cpp
tsocha: Did you miss this comment or do not agree with it?

libkineto/src/plugin/xpupti/XpuptiActivityHandlers.cpp
tsocha: It's probably done to avoid explicit early return. Resolved.

libkineto/src/plugin/xpupti/XpuptiActivityProfiler.cpp
```diff
@@ -9,53 +9,76 @@
...
       libkineto::timeSinceEpoch(std::chrono::high_resolution_clock::now());
+  if (scopeProfilerEnabled_) {
+    xpti_.startScopeActivity();
+  }
 }
 
 void XpuptiActivityProfilerSession::stop() {
+  if (scopeProfilerEnabled_) {
+    xpti_.stopScopeActivity();
+  }
```
tsocha, line 81: Similar case to line 90.
Do we need to check if scopeProfiler is enabled to disable it?

PR #1161, opened 2025-10-27
Add Safe Accesses to Rocproflogger

sraikund16:
Summary:
Saw vector access issues in OSS CI: https://github.com/pytorch/pytorch/actions/runs/18849910562/job/53786425781

It is likely this comes from the usage of .at() in this PR. Lets fix it by just having default values for these unexepected values

Differential Revision: D85609875


aaronenyeshi (approved):

PR #1164, opened 2025-11-03
Add Cuda Event to Kineto Trace

sraikund16:
Summary:
Adds event for CUDA 13.0+.

For 12.8 they add the event but there is no way of turning on timestamps.

Differential Revision: D84963573


aaronenyeshi (approved):

PR #1169, opened 2025-11-07
Fix repeated thread creation in CuptiActivityApi::teardownContext

leveretconey:
Running the following code causes a core dump:

```python
import os
os.environ["TEARDOWN_CUPTI"] = "1"

import torch

x = torch.randn(1024, 1024).cuda()

for i in range(5):
    with torch.profiler.profile() as p:
        y = x @ x

for i in range(5):
    y = x @ x
```

The issue occurs when we try to perform consecutive profiling sessions. In CuptiActivityApi::teardownContext, the next thread might be created before the previous thread finishes its work, so we need to prevent thread duplication.

### DIFF
--- a/libkineto/src/CuptiActivityApi.cpp
+++ b/libkineto/src/CuptiActivityApi.cpp
@@ -439,7 +439,11 @@ void CuptiActivityApi::teardownContext() {
   if (!tracingEnabled_) {
     return;
   }
+  if (tearingDown_) {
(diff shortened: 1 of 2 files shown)


leveretconey:
Hi maintainers,

Checking in on this PR - it's been a few weeks without any review. Is there anything I can do to help move this forward?

Thanks!

PR #1171, opened 2025-11-12
[AIU Profiler] Revert extract invocation number from AIU cb names

mcalman:
This post-processing step for AIU Kernel names is no longer needed and should be removed.


mcalman:
@sraikund16 @aaronenyeshi are either of you able to review this PR (along with PRs #1172 and #1182) since you reviewed the initial AIU profiler PR?

PR #1174, opened 2025-11-13
Introduce XPU scope profiler extending existing XPU profiler plugin

moksiuc:
**Summary:**

As XPU became a PyTorch built-in device, the profiler support is indispensable part of functionality completeness. In this PR, the XPU scope profiler is introduced by extending existing XPU profiler plugin. The XPU scope profiler is built on the foundation of intel PTI toolkit (https://github.com/intel/pti-gpu), and underlying SYCL runtime. It allows to gather XPU hardware metrics. The LIBKINETO_NOXPUPTI option is used to enable or disable the whole XPU profiler plugin during kineto build stage.

**Changes:**
- Added new ActivityType : XPU_SCOPE_PROFILER, enabling the new scope profiler
- Outputs XPU hardware metrics from the new scope profiler in Perfetto counters display mode ("C")
- Added gtest


moksiuc:
@EikanWang, @gujinghui

libkineto/src/plugin/xpupti/FindSYCLToolkit.cmake
```diff
@@ -105,6 +105,5 @@ if((NOT SYCL_INCLUDE_DIR) OR (NOT SYCL_LIBRARY_DIR) OR (NOT SYCL_LIBRARY))
 endif()
 
 message(DEBUG "The SYCL compiler is ${SYCL_COMPILER}")
-message(DEBUG "The SYCL Flags are ${SYCL_FLAGS}")
```
gujinghui: Why remove this message?

gujinghui:
@moksiuc  It's great that we are going to update our PTI integration code, and introduce new profiler path.
Could you help address below questions?
1. The alternative of ScopeProfiler is the RangeProfiler for CUDA? Looks like the RangeProfler is not enabled in PyTorch by default so far. Do you know why?
2. This PR is too huge to review. Can we split it to several PRs? For example, one PR for code refactor or cleanup per kineto or PTI changes, one or two PRs for ScopeProfiler, one PR for ChromeTraceLogger enhancement, and add test cases for each PRs.
3. BTW, CUDA provides CUDA_DRIVER activity to trace the driver actions. We should provide L0 actions as the counterpart, right? I remember, PTI should be able to do that. Do we have plan to cover it?
https://github.com/pytorch/kineto/blob/1e30d37905ddc1f9bd5e8de8d5e1148b3689f918/libkineto/src/ActivityType.cpp#L29

libkineto/src/plugin/xpupti/FindSYCLToolkit.cmake
moksiuc: SYCL_FLAGS is not used in any of cmake files

moksiuc:
> 1. The alternative of ScopeProfiler is the RangeProfiler for CUDA? Looks like the RangeProfler is not enabled in PyTorch by default so far. Do you know why?
It is enabled by providing experimental_config=_ExperimentalConfig(...). I don't know why it is this way but we are enabling our profiler the same way. One of the reasons may be that Range/Scope profiler requires parameters like HW metrics names that are passed through _ExperimentalConfig.

moksiuc:
> 3. BTW, CUDA provides CUDA_DRIVER activity to trace the driver actions. We should provide L0 actions as the counterpart, right? I remember, PTI should be able to do that. Do we have plan to cover it?
>    https://github.com/pytorch/kineto/blob/1e30d37905ddc1f9bd5e8de8d5e1148b3689f918/libkineto/src/ActivityType.cpp#L29

For sure not in this PR. I'll add this to our list of tasks.

moksiuc:
> 2. This PR is too huge to review. Can we split it to several PRs? For example, one PR for code refactor or cleanup per kineto or PTI changes, one or two PRs for ScopeProfiler, one PR for ChromeTraceLogger enhancement, and add test cases for each PRs.

Extracted clean up and adding config for scope profiler to separate PR's.
This one should be much smaller afterwards.
Currently I don't see further areas of extracting separate PR's as what would remain is full scope profiler implementation with tests and we'd like not to introduce half of the implementation that is not working functionally.

gujinghui (approved):

gujinghui (changes requested):
This PR is split to #1177, #1180, and more.

gujinghui:
@moksiuc Let's close this PR.

moksiuc:
@gujinghui this is the core of the scope profiler. When 2 smaller parts are merged this one would have only core profiler left.

moksiuc:
@EikanWang, @gujinghui
Please review.
2 extracted parts are already merged.
Now this change contains only code directly related to scope profiler.

moksiuc:
Tests are run on pytorch part:
https://github.com/pytorch/pytorch/pull/165766

libkineto/src/plugin/xpupti/XpuptiActivityApi.cpp
```diff
@@ -241,12 +349,22 @@ void XpuptiActivityApi::enableXpuptiActivities(
 #endif
         break;
 
+      case ActivityType::XPU_SCOPE_PROFILER:
+#if PTI_VERSION_AT_LEAST(0, 15)
+        scopeProfilerEnabled = true;
+#else
+        throw std::runtime_error(
+            "Intel® oneAPI version required to use scope profiler is at least 2025.3.1");
```
gujinghui: Can we give the PTI version here, instead of oneAPI package version?

libkineto/test/xpupti/compute/CMakeLists.txt
```diff
@@ -0,0 +1,55 @@
+# Copyright (c) Meta Platforms, Inc. and affiliates.
```
gujinghui, line 1: @chuanqi129 If we want to append these cases in XPU CI scope, how should we do?

chuanqi129: Hi @gujinghui , we don't have xpu ci test for kineto repo, do you mean put those test cases into stock pytorch? If so, we may evaluate whether can put them into https://github.com/pytorch/pytorch/tree/main/test/cpp/profiler

gujinghui (approved):
LGTM. I assume you already verified it on local real machine, right? @moksiuc

libkineto/test/xpupti/compute/CMakeLists.txt
gujinghui: yes. @moksiuc please sync with @chuanqi129 to make it. Thanks.

moksiuc:
> LGTM. I assume you already verified it on local real machine, right? @moksiuc

Yes, I did.

libkineto/src/plugin/xpupti/XpuptiActivityApi.cpp
moksiuc: I had previously PTI version but is it not clearer for the user to have one api version ?
Has the user possibility to install any PTI version he wants ?

gujinghui (approved):

gujinghui:
@sraikund16  could you help review this PR? Thanks.

libkineto/src/plugin/xpupti/XpuptiActivityApi.cpp
moksiuc: Added required PTI version in message

gujinghui:
@malfet @sraikund16 could you help review this PR? Thanks.

moksiuc:
@malfet @sraikund16 could you review, please ?

libkineto/include/ITraceActivity.h
```diff
@@ -53,6 +53,10 @@ struct ITraceActivity {
   virtual const std::string getMetadataValue(const std::string& key) const {
     return "";
   }
+  virtual const std::vector<std::pair<std::string, std::string>> getMetadata()
+      const {
+    return {};
+  }
```
divyanshk: Is there a way we can get away without updating the interface ?

libkineto/src/plugin/xpupti/XpuptiActivityApi.cpp
```diff
@@ -202,8 +204,115 @@ static void enableSpecifcRuntimeAPIsTracing() {
 }
 #endif
 
-void XpuptiActivityApi::enableXpuptiActivities(
+#if PTI_VERSION_AT_LEAST(0, 15)
```
divyanshk: Nit: `XPUPTI_VERSION_AT_LEAST` would be a better name for this macro. I can update that later (not relevant for this PR).

divyanshk (commented):
Thanks for the PR. I'm ramping up on this space. I'm curious to understand how much does it differ from the existing XPU profiler - maybe it should exist as a separate class in itself ? How can I learn more about the scope profiler ?

Also, thank you for adding tests.

moksiuc:
> Thanks for the PR. I'm ramping up on this space. I'm curious to understand how much does it differ from the existing XPU profiler - maybe it should exist as a separate class in itself ? How can I learn more about the scope profiler ?
> 
> Also, thank you for adding tests.

This PR adds new activity ActivityType::XPU_SCOPE_PROFILER, accepted by XPU profiler. With this new mode enabled XPU profiler additionally gathers requested HW metrics.
I don't think the separate class is necessary as it just adds additional mode to a few existing so far.
The new mode is a XPU alternative for what is implemented in class CuptiRangeProfiler.

libkineto/src/plugin/xpupti/XpuptiActivityApi.cpp
moksiuc: Ok, thanks for hint. I can update it if I have to make any other update in the code and push new commit.

moksiuc: After thinking about it XPUPTI_VERSION_AT_LEAST might be misleading as it is not xpupti version (that is part of kineto) but the PTI component version that xpupti uses underneath. Actually there is whole PTI api referenced starting with pti* prefix and PTI_VERSION_AT_LEAST refers to version of that api.

gujinghui:
> > Thanks for the PR. I'm ramping up on this space. I'm curious to understand how much does it differ from the existing XPU profiler - maybe it should exist as a separate class in itself ? How can I learn more about the scope profiler ?
> > Also, thank you for adding tests.
> 
> This PR adds new activity ActivityType::XPU_SCOPE_PROFILER, accepted by XPU profiler. With this new mode enabled XPU profiler additionally gathers requested HW metrics. I don't think the separate class is necessary as it just adds additional mode to a few existing so far. The new mode is a XPU alternative for what is implemented in class CuptiRangeProfiler.

Hi @divyanshk ,

For more details of CuptiRangeProfiler, please see this page. https://deepwiki.com/pytorch/kineto/3.1-nvidia-gpu-support-(cupti)#4-cupti-range-profiler-integration

@moksiuc is going to provide similar functionalities on XPU in this PR.

libkineto/include/ITraceActivity.h
moksiuc: @divyanshk I've found that existing metadataJson() method is enough so I've restored the original interface.

scotts:
Hi, @moksiuc, I'm also a new maintainer to this library. Thanks for all of the recent cleanup PRs!

On this PR, I have a similar concern that @divyanshk does on the code structuring. Specifically, this code uses `#ifdef`s to conditionally add member functions, member functions and member structs. This library has a history of doing that, but we want to start untangling that as it's difficult to maintain.

We should be able to achieve the same functionality by defining `XpuptiActivityScopeApi` (or a similar name) that derives from `XpuptiActivityApi`. Then we can use the `#ifdef`s to guard the entire files, which is much easier to reason about. You may need to still modify the `XpuptiActivityApi` interface itself a little so that it all works cleanly with `XpuActivityProfiler`, but that seems reasonable.

moksiuc:
@divyanshk , @scotts I've split files and moved the new code to new files. Please review.

divyanshk (commented):
@moksiuc  @gujinghui Can you help me understand a few things, thanks
1.) Do we have a different intel PTI API that is being leveraged for the range profiler ? How stable and well-supported is that API ? I ask because CUPTIRangeProfiler uses a different CUPTI API (`CUPTI Profiling API`) from the canonical profiler. Also that API is marked for deprecation so we are thinking of deprecating and deleting the CUPTIRangeProfiler.
2.) Can you point me to the official documentation for scope profiler API for Intel?
3.) Deferring code restructuring concerns to @scotts, I am starting to think `ActivityType` might be over-used here; because activitytype should be signifying profiling activities/events but `CUDA_PROFILER_RANGE` and `XPU_SCOPE_PROFILER` are independent profilers.  Thoughts?

moksiuc:
@divyanshk, 
1. The PTI API used for XPU Scope Profiler has been introduced in version 0.15 of https://github.com/intel/pti-gpu. According to their API policy it would always stay backwards compatible so it would not change in a way breaking the compatibility.
2. The documentation is not yet ready as the feature is not yet available - it is only in this PR. It would be prepared as soon as the feature is available.
3. Maybe CUDA_PROFILER_RANGE works differently but XPU_SCOPE_PROFILER is not independent profiler but additional mode of current profiler. It works independently on other profiling modes and gathers additional HW metrics.

moksiuc:
@scotts, could you review restructured code, please ?

moksiuc:
Whole cleanup extracted to separate PR in case it is preferred:
https://github.com/pytorch/kineto/pull/1256

moksiuc:
I noticed a lot of formatting issues. Could someone help what version of clang-format is used on CI ?
My formatting is obtained with lintrunner. clang-format in version 18.1.3.
I tried earlier version but it doesn't work with current .clang-format contents, for ex version 14 gives an error:
.clang-format:64:1: error: unknown key 'Macros'

libkineto/include/ActivityType.h
```diff
@@ -59,7 +61,8 @@ ActivityType toActivityType(const std::string& str);
 
 // Return an array of all activity types except COUNT
 constexpr int activityTypeCount = (int)ActivityType::ENUM_COUNT;
-constexpr int defaultActivityTypeCount = (int)ActivityType::OPTIONAL_ACTIVITY_TYPE_START;
+constexpr int defaultActivityTypeCount =
+    (int)ActivityType::OPTIONAL_ACTIVITY_TYPE_START;
```
gujinghui: @moksiuc  Let's remove all code changes due to format issue. Otherwise, too many changes will obstruct reviewer's view. We can have separate PR to fix format issues.

libkineto/src/plugin/xpupti/XpuptiActivityApiAndSessionVersionSelector.h
```diff
@@ -0,0 +1,31 @@
...
+#include "XpuptiProfilerMacros.h"
+
+namespace KINETO_NAMESPACE {
+
+#if PTI_VERSION_AT_LEAST(0, 15)
+
+class XpuptiActivityApiV2;
+
+#define XPUPTI_ACTIVITY_API XpuptiActivityApiV2
+#define XPUPTI_ACTIVITY_PROFILER_SESSION XpuptiActivityProfilerSessionV2
```
gujinghui: @moksiuc It’s not a good idea to use XPUPTI_ACTIVITY_API and XPUPTI_ACTIVITY_PROFILER_SESSION here. I understand we’re addressing feedback from @divyanshk and @scotts to remove #ifdefs, and I fully agree that we shouldn’t rely on #ifdef to define the activity API types. That’s why having XpuptiActivityApiV2 derive from XpuptiActivityApi in this PR makes sense.

However, it’s still reasonable to use #ifdef at the call sites where V2-only APIs are invoked. Compared to XPUPTI_ACTIVITY_API, putting the #ifdef directly around the V2 calls tends to improve readability and makes it clearer which code paths are version-specific.

libkineto/include/ActivityType.h
moksiuc: ok, good point

libkineto/src/plugin/xpupti/XpuptiActivityApiAndSessionVersionSelector.h
moksiuc: Done

libkineto/src/plugin/xpupti/XpuptiActivityProfilerSession.cpp
```diff
@@ -125,19 +125,23 @@ std::vector<libkineto::ResourceInfo> XpuptiActivityProfilerSession::
...
 void XpuptiActivityProfilerSession::pushUserCorrelationId(uint64_t id) {
-  xpti_.pushCorrelationID(id, XpuptiActivityApi::CorrelationFlowType::User);
+  xpti_.pushCorrelationID(
+      id, SELECT_VERSION(XpuptiActivityApi)::CorrelationFlowType::User);
 }
 
 void XpuptiActivityProfilerSession::popUserCorrelationId() {
-  xpti_.popCorrelationID(XpuptiActivityApi::CorrelationFlowType::User);
+  xpti_.popCorrelationID(
+      SELECT_VERSION(XpuptiActivityApi)::CorrelationFlowType::User);
```
gujinghui: Why the CorrelationFlowType must be decorated by different class names? It only be defined in XpuptiActivityApi, right? Any compilation error?

libkineto/src/plugin/xpupti/XpuptiScopeProfilerApi.cpp
```diff
@@ -0,0 +1,198 @@
...
+
+#endif
+
+namespace KINETO_NAMESPACE {
+
+SELECT_VERSION(XpuptiActivityApi) &
+    SELECT_VERSION(XpuptiActivityApi)::singleton() {
+  static SELECT_VERSION(XpuptiActivityApi) instance;
+  return instance;
+}
```
gujinghui: Can we use the `#ifdef` here, instead of SELECT_VERSION?

libkineto/src/plugin/xpupti/XpuptiActivityProfiler.cpp
```diff
@@ -66,8 +66,11 @@ std::unique_ptr<libkineto::IActivityProfilerSession> XPUActivityProfiler::
     configure(
         const std::set<ActivityType>& activity_types,
         const libkineto::Config& config) {
-  return std::make_unique<XpuptiActivityProfilerSession>(
-      XpuptiActivityApi::singleton(), name(), config, activity_types);
+  return std::make_unique<SELECT_VERSION(XpuptiActivityProfilerSession)>(
+      SELECT_VERSION(XpuptiActivityApi)::singleton(),
```
gujinghui: Can we use the #ifdef here, instead of SELECT_VERSION?

libkineto/src/plugin/xpupti/XpuptiActivityProfilerSession.cpp
```diff
@@ -44,7 +44,7 @@ std::unordered_set<std::string_view>
 
 // =========== Session Constructor ============= //
 XpuptiActivityProfilerSession::XpuptiActivityProfilerSession(
-    XpuptiActivityApi& xpti,
+    SELECT_VERSION(XpuptiActivityApi) & xpti,
```
gujinghui: Can we also pass the ref of XpuptiActivityApi obj, but do dynamic_cast for XpuptiActivityApi to XpuptiActivityApiV2 if needed in the implementation?

libkineto/src/plugin/xpupti/XpuptiActivityProfilerSession.h
```diff
@@ -23,17 +23,16 @@
...
 
 class XpuptiActivityProfilerSession : public libkineto::IActivityProfilerSession {
  public:
   XpuptiActivityProfilerSession() = delete;
-  XpuptiActivityProfilerSession(XpuptiActivityApi& xpti,
-                                const std::string& name,
-                                const libkineto::Config& config,
-                                const std::set<ActivityType>& activity_types);
+  XpuptiActivityProfilerSession(
+      SELECT_VERSION(XpuptiActivityApi) & xpti,
```
gujinghui: Can we also pass the ref of XpuptiActivityApi obj, but do dynamic_cast for XpuptiActivityApi to XpuptiActivityApiV2 if needed in the implementation?

libkineto/src/plugin/xpupti/XpuptiActivityProfilerSession.h
```diff
@@ -111,13 +110,34 @@ class XpuptiActivityProfilerSession : public libkineto::IActivityProfilerSession
 
   libkineto::getLinkedActivityCallback cpuActivity_;
 
-  XpuptiActivityApi& xpti_;
+  SELECT_VERSION(XpuptiActivityApi) & xpti_;
```
gujinghui: ditto.

libkineto/src/plugin/xpupti/XpuptiProfilerMacros.h
```diff
@@ -32,4 +32,12 @@ using namespace libkineto;
 
 using DeviceIndex_t = int8_t;
 
+#if PTI_VERSION_AT_LEAST(0, 15)
+#define SELECT_VERSION(obj) obj##V2
+#else
+#define SELECT_VERSION(obj) obj
+#endif
+
+class SELECT_VERSION(XpuptiActivityApi);
```
gujinghui: Can we use `#ifdef` in the code, instead of a macro?

moksiuc: It would be again too many #ifdefs throughout the code reducing the readability.

libkineto/src/plugin/xpupti/XpuptiActivityProfilerSession.h
moksiuc: We know the type statically during compilation. There is no need to deduce it in runtime.

libkineto/src/plugin/xpupti/XpuptiActivityProfilerSession.cpp
moksiuc: We know the type statically during compilation. There is no need to deduce it in runtime.

libkineto/src/plugin/xpupti/XpuptiActivityProfilerSession.cpp
moksiuc: Removed decorator

libkineto/src/plugin/xpupti/XpuptiScopeProfilerApi.cpp
moksiuc: Removed SELECT_VERSION, but #ifdef not added not to bring them back in large amount

libkineto/src/plugin/xpupti/XpuptiActivityProfiler.cpp
moksiuc: Removed SELECT_VERSION, but #ifdef not added not to bring them back in large amount

libkineto/src/plugin/xpupti/XpuptiActivityProfilerSession.h
moksiuc: Done

libkineto/src/plugin/xpupti/XpuptiProfilerMacros.h
moksiuc: Removed SELECT_VERSION, but #ifdef not added not to bring them back in large amount

libkineto/src/plugin/xpupti/XpuptiActivityHandlersV2.cpp
```diff
@@ -0,0 +1,123 @@
...
+    case PTI_METRIC_VALUE_TYPE_BOOL8:
+      scopeActivity->addMetadata(metricName, value.b8 ? "true" : "false");
+      return;
+
+    default:
+      break;
+  }
+}
+
+void XpuptiActivityProfilerSessionV2::handleScopeRecord(
```
gujinghui: May I know why we split the impl of XpuptiActivityProfilerSession into two CPP files?
Should we merge them together in future?

libkineto/src/plugin/xpupti/XpuptiActivityProfiler.cpp
```diff
@@ -66,8 +66,14 @@ std::unique_ptr<libkineto::IActivityProfilerSession> XPUActivityProfiler::
...
-  return std::make_unique<XpuptiActivityProfilerSession>(
-      XpuptiActivityApi::singleton(), name(), config, activity_types);
+#if PTI_VERSION_AT_LEAST(0, 15)
+  using XpuptiActivityProfilerSessionT = XpuptiActivityProfilerSessionV2;
+#else
+  using XpuptiActivityProfilerSessionT = XpuptiActivityProfilerSession;
+#endif
+
+  return std::make_unique<XpuptiActivityProfilerSessionT>(
+      XpuptiActivityApiT::singleton(), name(), config, activity_types);
```
gujinghui:
```suggestion
#if PTI_VERSION_AT_LEAST(0, 15)
    return std::make_unique<XpuptiActivityProfilerSessionV2>(
          XpuptiActivityApiV2::singleton(), name(), config, activity_types);
#else
    return std::make_unique<XpuptiActivityProfilerSession>(
          XpuptiActivityApi::singleton(), name(), config, activity_types);
#endif
```

libkineto/src/plugin/xpupti/XpuptiActivityProfilerSession.cpp
```diff
@@ -44,7 +44,7 @@ std::unordered_set<std::string_view>
 
 // =========== Session Constructor ============= //
 XpuptiActivityProfilerSession::XpuptiActivityProfilerSession(
-    XpuptiActivityApi& xpti,
+    XpuptiActivityApiT& xpti,
```
gujinghui: XpuptiActivityProfilerSession will not use any APIs from XpuptiActivityApiV2? Why we need to refine the class name here?

libkineto/src/plugin/xpupti/XpuptiActivityProfilerSession.h
```diff
@@ -23,17 +23,16 @@
...
 
 class XpuptiActivityProfilerSession : public libkineto::IActivityProfilerSession {
  public:
   XpuptiActivityProfilerSession() = delete;
-  XpuptiActivityProfilerSession(XpuptiActivityApi& xpti,
-                                const std::string& name,
-                                const libkineto::Config& config,
-                                const std::set<ActivityType>& activity_types);
+  XpuptiActivityProfilerSession(
+      XpuptiActivityApiT& xpti,
```
gujinghui: XpuptiActivityProfilerSession will not use any APIs from XpuptiActivityApiV2? Why we need to refine the class name here?

libkineto/src/plugin/xpupti/XpuptiActivityProfilerSession.h
```diff
@@ -111,13 +110,34 @@ class XpuptiActivityProfilerSession : public libkineto::IActivityProfilerSession
 
   libkineto::getLinkedActivityCallback cpuActivity_;
 
-  XpuptiActivityApi& xpti_;
+  XpuptiActivityApiT& xpti_;
```
gujinghui: XpuptiActivityProfilerSession will not use any APIs from XpuptiActivityApiV2? Why we need to refine the class name here?

libkineto/src/plugin/xpupti/XpuptiProfilerMacros.h
```diff
@@ -32,4 +32,12 @@ using namespace libkineto;
 
 using DeviceIndex_t = int8_t;
 
+#if PTI_VERSION_AT_LEAST(0, 15)
+class XpuptiActivityApiV2;
+using XpuptiActivityApiT = XpuptiActivityApiV2;
+#else
+class XpuptiActivityApi;
+using XpuptiActivityApiT = XpuptiActivityApi;
+#endif
```
gujinghui: not clear to me... why we have to use the macro to redefine the API at build stage...

libkineto/src/plugin/xpupti/XpuptiActivityApiV2.cpp
```diff
@@ -0,0 +1,197 @@
...
+XpuptiActivityApiT& XpuptiActivityApiT::singleton() {
+  static XpuptiActivityApiT instance;
+  return instance;
+}
+
+#if PTI_VERSION_AT_LEAST(0, 15)
+
+XpuptiActivityApi& XpuptiActivityApi::singleton() {
+  return XpuptiActivityApiV2::singleton();
+}
```
gujinghui:
```suggestion
#if PTI_VERSION_AT_LEAST(0, 15)
XpuptiActivityApiV2& XpuptiActivityApiV2::singleton() {
  static XpuptiActivityApiV2 instance;
  return instance;
}
```

libkineto/src/plugin/xpupti/XpuptiProfilerMacros.h
gujinghui: We are able to make sure which version of XpuptiActivityApi will be created in libkineto/src/plugin/xpupti/XpuptiActivityProfiler.cpp. We do not need the refine the class name at build time, right?

libkineto/src/plugin/xpupti/XpuptiActivityHandlersV2.cpp
moksiuc: All V2 related files are separate from 'V1'. I don't suppose there would be merged.

libkineto/src/plugin/xpupti/XpuptiActivityProfiler.cpp
moksiuc: Done

libkineto/src/plugin/xpupti/XpuptiActivityProfilerSession.h
moksiuc: done

libkineto/src/plugin/xpupti/XpuptiActivityProfilerSession.h
moksiuc: done

libkineto/src/plugin/xpupti/XpuptiProfilerMacros.h
moksiuc: It is not a macro. But anyway it is no longer necessary.

libkineto/src/plugin/xpupti/XpuptiActivityApiV2.cpp
moksiuc: done

libkineto/src/plugin/xpupti/XpuptiActivityProfilerSession.cpp
moksiuc: done

libkineto/src/plugin/xpupti/XpuptiActivityHandlersV2.cpp
moksiuc: It was requested in the review to separate it to have #ifdef sections for whole files rather then a lot of fragments in the file.

libkineto/src/plugin/xpupti/XpuptiActivityApiV2.cpp
```diff
@@ -0,0 +1,208 @@
...
+#if PTI_VERSION_AT_LEAST(0, 15)
+
+XpuptiActivityApiV2& XpuptiActivityApiV2::singleton() {
+  static XpuptiActivityApiV2 instance;
+  return instance;
+}
+
+#else
+
+XpuptiActivityApi& XpuptiActivityApi::singleton() {
```
gujinghui: why we have to put this func in V2 cpp file?

libkineto/src/plugin/xpupti/XpuptiActivityApi.cpp
```diff
@@ -14,11 +14,6 @@ namespace KINETO_NAMESPACE {
 
 constexpr size_t kBufSize(4 * 1024 * 1024);
 
-XpuptiActivityApi& XpuptiActivityApi::singleton() {
-  static XpuptiActivityApi instance;
-  return instance;
-}
```
gujinghui, line 21: why we move this func to V2 cpp file?

libkineto/src/plugin/xpupti/XpuptiActivityApiV2.cpp
```diff
@@ -0,0 +1,208 @@
...
+  static XpuptiActivityApi instance(false);
+  return instance;
+}
+
+#endif
+
+#if PTI_VERSION_AT_LEAST(0, 15)
+
+XpuptiActivityApi& XpuptiActivityApi::singleton() {
```
gujinghui: why we still need this entry point? We always call singleton with the decorator of exact class name, right?
Like below,
`  return std::make_unique<XpuptiActivityProfilerSessionV2>(
      XpuptiActivityApiV2::singleton(), name(), config, activity_types);`

Is there other place where we must call the singleton with XpuptiActivityApi?

libkineto/src/plugin/xpupti/XpuptiActivityApi.h
```diff
@@ -24,7 +24,8 @@ class XpuptiActivityApi {
  public:
   enum CorrelationFlowType { Default, User };
 
-  XpuptiActivityApi() = default;
+  XpuptiActivityApi(bool scopeProfilerActivityAccepted)
+      : scopeProfilerActivityAccepted_(scopeProfilerActivityAccepted) {}
```
gujinghui: Looks like I see why we need this flag now. We use this flag at runtime to reduce the `#ifdef` in the code, right?

libkineto/src/plugin/xpupti/XpuptiActivityHandlersV2.cpp
gujinghui: got it. thanks.

libkineto/src/plugin/xpupti/XpuptiActivityApiV2.cpp
moksiuc: Returned to V1 file

libkineto/src/plugin/xpupti/XpuptiActivityApiV2.cpp
moksiuc: For example:
kineto/libkineto/src/plugin/xpupti/XpuptiActivityApi.cpp:27
if (!singleton().externalCorrelationEnabled_) {

libkineto/src/plugin/xpupti/XpuptiActivityApi.h
moksiuc: Changed to #ifdef

libkineto/src/plugin/xpupti/XpuptiActivityApi.cpp
moksiuc: Returned to V1 file

libkineto/src/plugin/xpupti/XpuptiActivityApiV2.cpp
gujinghui: Sounds like the code is wrong. The original design of singleton is not good. Why we need to use singleton() to call the member function, or to access to member data here?

gujinghui: > For example: kineto/libkineto/src/plugin/xpupti/XpuptiActivityApi.cpp:27 if (!singleton().externalCorrelationEnabled_) {

The code should be `if (externalCorrelationEnabled_) {`, right? We do not need to call singleton() again.

moksiuc: Cleaned up singleton usage.
However it must stay inside bufferRequestedTrampoline and bufferCompletedTrampoline as they are passed as C function pointers to
ptiViewSetCallbacks(bufferRequestedTrampoline, bufferCompletedTrampoline);
So they are called back without an object and it has to be taken from singleton() member.

libkineto/src/plugin/xpupti/XpuptiActivityApi.h
```diff
@@ -32,8 +32,8 @@ class XpuptiActivityApi {
 
   static XpuptiActivityApi& singleton();
 
-  static void pushCorrelationID(int id, CorrelationFlowType type);
-  static void popCorrelationID(CorrelationFlowType type);
```
gujinghui, line 36: Do we know why these two APIs were originally made static?
Are they called directly via the class in PyTorch?

moksiuc: No idea but currently there is no need they are static.

gujinghui: Looks like cupti and other pti have same static APIs. If the change is not necessary, let's revert it.

libkineto/src/plugin/xpupti/XpuptiActivityProfiler.cpp
```diff
@@ -66,8 +66,17 @@ std::unique_ptr<libkineto::IActivityProfilerSession> XPUActivityProfiler::
     configure(
         const std::set<ActivityType>& activity_types,
         const libkineto::Config& config) {
+#if PTI_VERSION_AT_LEAST(0, 15)
+  return std::make_unique<XpuptiActivityProfilerSessionV2>(
+      XpuptiActivityApi::singleton(),
+      XpuptiActivityApiV2::singleton(),
```
gujinghui: I think it's not a good idea to keep v1 an v2 separate class. May I know why we do not refine them by following the composition design module?

libkineto/src/plugin/xpupti/XpuptiActivityApi.h
moksiuc: Maybe their need it static. It is not needed in our case. And it even can't be static as it accesses externalCorrelationEnabled_ member.

libkineto/src/plugin/xpupti/XpuptiActivityProfiler.cpp
moksiuc: There is no need to keep V1 inside V2 as they have no connection. V1 is kept inside session and V2 inside sessionV2.
You wanted separated singletons recently so they are actually separate. And member V1 inside V2 (composition) would be never used as V2 doesn't reference V1 at all.

gujinghui: Yes, you're right. There's no connection btw V1 and V2. Then, we should not call it V2. Otherwise, the code is very weird and confused here. Let's rename it XpuptiScopeProfilerApi. What do you think?

libkineto/src/plugin/xpupti/XpuptiActivityApi.h
gujinghui: Since V1 and V2 will be handled separately, we can revert all changes to the V1 .h and .cpp files because they are not necessary for this PR. This will make the PR clearer and easier to review.

moksiuc: Separated to https://github.com/pytorch/kineto/pull/1313

libkineto/src/plugin/xpupti/XpuptiActivityProfiler.cpp
moksiuc: Let's try

gujinghui (approved):
@moksiuc  please resolve the change conflicts. Thanks.

libkineto/src/plugin/xpupti/XpuptiActivityApi.cpp
```diff
@@ -244,6 +244,15 @@ void XpuptiActivityApi::enableXpuptiActivities(
         XPUPTI_CALL(ptiViewEnable(PTI_VIEW_DRIVER_API));
         break;
 
+      case ActivityType::XPU_SCOPE_PROFILER:
+#if PTI_VERSION_AT_LEAST(0, 15)
+#else
+        throw std::runtime_error(
```
gujinghui, line 253: When does this error occur?

libkineto/src/plugin/xpupti/XpuptiActivityProfiler.cpp
```diff
@@ -66,8 +66,17 @@ std::unique_ptr<libkineto::IActivityProfilerSession> XPUActivityProfiler::
     configure(
         const std::set<ActivityType>& activity_types,
         const libkineto::Config& config) {
+#if PTI_VERSION_AT_LEAST(0, 15)
+  return std::make_unique<XpuptiActivityProfilerSessionV2>(
+      XpuptiActivityApi::singleton(),
+      XpuptiScopeProfilerApi::singleton(),
```
gujinghui: Why we have to make XpuptiScopeProfilerApi as a singleton here?
If XpuptiScopeProfilerApi is only used in XpuptiActivityProfilerSessionV2, why we do not make it part of XpuptiActivityProfilerSessionV2? At least we can make it static in XpuptiActivityProfilerSessionV2?

libkineto/src/plugin/xpupti/XpuptiActivityApi.cpp
moksiuc: When you compile with PTI<0.15 and try to use XPU_SCOPE_PROFILER activity enabling scope profiler.
Then instead of silent ignoring you got meaningful error message.

libkineto/src/plugin/xpupti/XpuptiActivityProfiler.cpp
moksiuc: We'd like to keep architecture consistent. If in V1 there is separate api and session, we'd like to keep the same idea in V2. Api is communicating with PTI.

libkineto/src/plugin/xpupti/XpuptiActivityApi.cpp
moksiuc: So the case is SessionV1 with XPU_SCOPE_PROFILER.

libkineto/src/plugin/xpupti/XpuptiActivityProfiler.cpp
moksiuc: However it doesn't need to be singleton and can be just a member of SessionV2

libkineto/src/plugin/xpupti/XpuptiActivityApi.cpp
gujinghui: How does PTI handle this case? The scope profiler is not supported by PTI earlier than 0.15. Then, some error should happen in PTI?

moksiuc: This is not the case. If we build with PTI<0.15 we'll build sessionV1 and apiV1 only. There would not be available any code related to scope profiler so without this change providing ActivityType::XPU_SCOPE_PROFILER would be just silently ignored.

gujinghui (approved):

gujinghui (approved):

gujinghui:
@divyanshk  This PR is ready now. Could you help review? Thanks.

libkineto/include/ActivityType.h
```diff
@@ -43,14 +43,15 @@ enum class ActivityType {
...
-  COLLECTIVE_COMM = 23, // collective communication
+  XPU_SCOPE_PROFILER = 23, // XPUPTI Profiler scope for performance metrics
+  COLLECTIVE_COMM = 24, // collective communication
 
   // PRIVATEUSE1 Activity types are used for custom backends.
   // The corresponding device type is `DeviceType::PrivateUse1` in PyTorch.
-  PRIVATEUSE1_RUNTIME = 24, // host side privateUse1 runtime events
-  PRIVATEUSE1_DRIVER = 25, // host side privateUse1 driver events
+  PRIVATEUSE1_RUNTIME = 25, // host side privateUse1 runtime events
+  PRIVATEUSE1_DRIVER = 26, // host side privateUse1 driver events
```
divyanshk: We shouldn't change the ordering for COLLECTIVE_COMM, PRIVATEUSE1_RUNTIME, PRIVATEUSE1_DRIVER.

libkineto/test/xpupti/XpuptiTestUtilities.cpp
```diff
@@ -226,9 +226,33 @@ RunProfilerTest(
...
+    bool nameStartsWithMetrics = pActivity->name().find("metrics:") == 0;
+    auto [metricsCount, metricsMask] =
         CountMetricsInString(metrics, pActivity->metadataJson());
 
+    switch (pActivity->type()) {
+      case KN::ActivityType::CONCURRENT_KERNEL:
+        if (nameStartsWithMetrics)
+          goto label_scope;
+        else
+          goto label_default;
```
divyanshk: Let's avoid goto!

libkineto/src/output_json.cpp
```diff
@@ -653,16 +653,51 @@ void ChromeTraceLogger::handleActivity(const libkineto::ITraceActivity& op) {
   sanitizeStrForJSON(op_name);
   sanitizeForNonReadableChars(op_name);
 
-  // clang-format off
   ts = transToRelativeTime(ts);
-  fmt::print(traceOf_, R"JSON(
+
+  if (op.type() == ActivityType::XPU_SCOPE_PROFILER) {
+    std::string metricsStr = op.metadataJson();
```
divyanshk: lets do `sanitizeStrForJSON(metricsStr)`

libkineto/src/output_json.cpp
```diff
@@ -653,16 +653,51 @@ void ChromeTraceLogger::handleActivity(const libkineto::ITraceActivity& op) {
...
+  {{
+    "name": "{}",
+    "ph": "C",
+    "ts": {}.{:03},
+    "pid": {},
+    "tid": {},
+    "args": {{ {} }}
+  }},)JSON",
+        // clang-format on
+        activityName.substr(0, activityName.find('_')),
```
divyanshk: lets just hard-code this to `xpu` instead of extracting it from `activityName`. or better just use the full`activityName`  ?

libkineto/include/ActivityType.h
```diff
@@ -43,14 +43,15 @@ enum class ActivityType {
   HPU_OP = 20, // HPU host side runtime event
   XPU_RUNTIME = 21, // host side xpu runtime events
   XPU_DRIVER = 22, // host side xpu driver events
-  COLLECTIVE_COMM = 23, // collective communication
+  XPU_SCOPE_PROFILER = 23, // XPUPTI Profiler scope for performance metrics
```
divyanshk: I know we have a `CUDA_PROFILER_RANGE` already. Ignoring that for a moment.

Just thinking from the basic, do we need a new activity for a new profiler mode? can't we do with the existing XPU_* activities ?

I asked this earlier but want to learn more.

libkineto/libkineto_defs.bzl
```diff
@@ -49,9 +49,12 @@ def get_libkineto_xpupti_srcs(with_api = True):
     return [
         "src/plugin/xpupti/XpuptiActivityApi.cpp",
         "src/plugin/xpupti/XpuptiActivityHandlers.cpp",
+        "src/plugin/xpupti/XpuptiActivityHandlersV2.cpp",
```
divyanshk: @scotts you had concerns on `xxx.cpp` and `xxxV2.xpp`

libkineto/src/plugin/xpupti/XpuptiActivityApi.cpp
```diff
@@ -245,6 +245,15 @@ void XpuptiActivityApi::enableXpuptiActivities(
         XPUPTI_CALL(ptiViewEnable(PTI_VIEW_DRIVER_API));
         break;
 
+      case ActivityType::XPU_SCOPE_PROFILER:
+#if PTI_VERSION_AT_LEAST(0, 15)
```
divyanshk, line 249: Nit: this is a little confusing to read, maybe just a comment on what happens on the if condition and hows its handled.

divyanshk (commented):
Left some comments. Important one is in `ActivityType.h`

I need to look again at `output_json.cpp`.

libkineto/src/plugin/xpupti/XpuptiActivityApi.cpp
gujinghui: Sounds reasonable. @moksiuc, could you please add comments to the if condition path?

libkineto/include/ActivityType.h
gujinghui: @moksiuc  Let's put the new item at the end of the list if still needed.

libkineto/libkineto_defs.bzl
gujinghui: @scotts  Could you share your thoughts on this?

We are considering splitting the new implementations into separate V2 C++ files. Otherwise, we would need to keep all implementations in the original C++ file and use `#if / #else / #endif` blocks within the file.

Do you have any suggestions? Thanks.

libkineto/include/ActivityType.h
moksiuc: Usually enums are used by name and should not depend on their values.
But I can avoid renumbering and move newly added enum to the end.

libkineto/test/xpupti/XpuptiTestUtilities.cpp
moksiuc: Although goto is almost always discouraged in this case it gave the simplest and cleanest solution. I can avoid it however.

libkineto/src/output_json.cpp
moksiuc: Done

libkineto/src/output_json.cpp
moksiuc: Full activity name is too long and looks bad in perfetto. Hardcoded xpu.

libkineto/include/ActivityType.h
moksiuc: It is a separate mode of gathering HW metrics and needs separate activity to enable it.
Planned GPU_PM_COUNTER could be used in the future (https://github.com/pytorch/kineto/pull/1212);

libkineto/libkineto_defs.bzl
moksiuc: Until no specific concern is provided I can't do anything here.

libkineto/src/plugin/xpupti/XpuptiActivityApi.cpp
moksiuc: Comment added

libkineto/libkineto_defs.bzl
scotts: @gujinghui, @divyanshk, I think it's definitely better to have new implementations in separate classes and files.

But I dislike just appending `V2` to the name of a thing - that doesn't communicate much about why it exists. Based on the fact that it's the activity handler for the "scope" functionality, I think we should probably name it `XpuptiScopeActivityHandlers`, which would then be reflected in all of the file names as well. This would apply to all of the new classes that have `V2` at the end, they should probably be `XpuScope*` instead.

libkineto/src/plugin/xpupti/XpuptiScopeProfilerApi.cpp
```diff
@@ -0,0 +1,188 @@
...
+    ~safe_pti_scope_collection_handle_t() noexcept {
+  try {
+    XPUPTI_CALL(ptiMetricsScopeDisable(handle_));
+  } catch (...) {
+    exceptFromDestructor_ = std::current_exception();
+  }
+}
+
+void XpuptiScopeProfilerApi::enableScopeProfiler(const Config& cfg) {
+#ifdef HAS_XPUPTI
```
scotts: Can we just wrap entire files with `#ifdef HAS_XPUTI`? Or, better yet, do we even need it in XPU specific files? The build system should prevent compiling XPU specific files when it's not enabled.

moksiuc: There is the same technique used in already present XpuptiActivityApi.cpp file.
I can check it but I don't think it should be a part of this PR. I'll prepare the separate one to clean this up in whole xpupti files.

libkineto/libkineto_defs.bzl
moksiuc: ScopeApi and ScopeHandlers are names would correctly describe the functionality.
ScopeSession matches less as it also contains Session through derivation.

gujinghui: @scotts   @divyanshk 
We correct the names now. There is no V2 in all files. Could you help review again? Thanks.

gujinghui:
@divyanshk  @scotts all comments have been addressed. Could you help review again? Thanks.

gujinghui (approved):

divyanshk (approved):
Thanks for addressing the comments! Looks good to me!

moksiuc:
Failure in the test
Linux CUDA build and test / pr-test / linux-cuda-build-and-test (pull_request)
seems unrelated to my change:
=================================== FAILURES ===================================
_______________________ TestProfiler.test_dynamic_toggle _______________________
Traceback (most recent call last):
  File "/opt/conda/lib/python3.11/unittest/case.py", line 57, in testPartExecutor
    yield
  File "/opt/conda/lib/python3.11/unittest/case.py", line 623, in run
    self._callTestMethod(testMethod)
  File "/opt/conda/lib/python3.11/unittest/case.py", line 579, in _callTestMethod
    if method() is not None:
       ^^^^^^^^
  File "/pytorch/kineto/pytorch/torch/testing/_internal/common_utils.py", line 3448, in wrapper
    method(*args, **kwargs)
  File "/pytorch/kineto/pytorch/torch/testing/_internal/common_utils.py", line 1821, in wrapper
    fn(*args, **kwargs)
  File "/pytorch/kineto/pytorch/test/profiler/test_profiler.py", line 2312, in test_dynamic_toggle
    self.assertTrue(all(device not in e.name for e in p1.events()))
  File "/opt/conda/lib/python3.11/unittest/case.py", line 715, in assertTrue
    raise self.failureException(msg)
AssertionError: False is not true

gujinghui:
@scotts   Could you help review this PR again? We already addressed all comments. Thanks.

aostrowski-hbn:
@scotts @divyanshk Can we merge?

PR #1172, opened 2025-11-12
[AIU Profiler] Allow AIU profiling via env vars on newer PyTorch versions

mcalman:
Fixes a bug where setting the environment variable `ProfilerActivity=PrivateUse1` wouldn't enable the PrivateUse1 activities for AIU as intended.

This is because `ActivityType::PRIVATEUSE1_RUNTIME` and `ActivityType::PRIVATEUSE1_DRIVER` are enabled by default when passing `torch.profiler.ProfilerActivity.CPU` to `torch.profiler.profile` on newer versions of PyTorch (see: [here](https://github.com/pytorch/pytorch/blob/078d475d3bb104823e70ce975c2ee0d4d2fb0952/torch/csrc/profiler/kineto_shim.cpp#L29)).

PR #1182, opened 2025-11-21
[AIU Profiler] Change activity types

mcalman:
1. Changes the activity types from `cuda_driver` and `cuda_runtime` to `privateuse1_driver` and `privateuse1_runtime` respectively.
2. Switches from `cuda_driver` to `gpu_memset` for memset activities so that they are properly accounted for in the Step Time Breakdown in TensorBoard.


mcalman:
@malfet @aaronenyeshi can we merge this?

#1171, #1172, and #1229 also include other small fixes to the AIU profiler. If possible, can you review these as well?

PR #1198, opened 2025-12-16
Refactor of handling version metadata parameters for devices and add XPU support

PawelSwider2000:
Refactor of existing `CuptiActivityProfiler::logGpuVersions` and extracting it into separate class `DeviceVersionLogger` which should be overwritten by each device. Such approach allow for much easier creating support for this feature for various devices, it allow also to keep the device specific code separated from the rest of code.

PR converts existing CUDA and ROCTRACER implementations to new format and add support for XPU devices via `XpuVersionLogger` class. 

For XPU following information will be added to logs:
```
      "pti_version": "<version>",
      "level_zero_version": <version>,
      "sycl_compiler_version": <version>,
```

### DIFF
--- a/libkineto/src/CuptiActivityProfiler.cpp
+++ b/libkineto/src/CuptiActivityProfiler.cpp
@@ -21,6 +21,7 @@
 #include <unordered_map>
 #include <vector>
 #include "ApproximateClock.h"
+#include "VersionLogger.h"
 
 #ifdef HAS_CUPTI
 #include <cupti.h>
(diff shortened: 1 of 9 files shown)


libkineto/src/CuptiActivityProfiler.cpp
```diff
@@ -300,7 +257,7 @@ void CuptiActivityProfiler::processTraceInternal(ActivityLogger& logger) {
       LOGGER_OBSERVER_ADD_METADATA(pair.first, pair.second);
     }
   }
-  for (auto& pair : versionMetadata_) {
+  for (auto& pair : versionLogger_ ? versionLogger_->getVersionMetadata() : std::unordered_map<std::string, std::string>{}) {
```
moksiuc, line 260: if (versionLogger_) for ... would be simpler

moksiuc (changes requested):

libkineto/src/VersionLogger.cpp
```diff
@@ -0,0 +1,78 @@
...
+std::unique_ptr<DeviceVersionLogger> selectDeviceVersionLogger(
+    std::recursive_mutex& mutex) {
+#ifdef HAS_CUPTI
+  return std::make_unique<CudaVersionLogger>(mutex);
+#elif HAS_ROCTRACER
+  return std::make_unique<HipVersionLogger>(mutex);
+#elif HAS_XPUPTI
+  return std::make_unique<XpuVersionLogger>(mutex);
+#else
+  return nullptr;
```
tsocha, line 74: I would like to avoid `nullptr`, could you add default behavior for CPU?

libkineto/src/plugin/xpupti/CMakeLists.txt
```diff
@@ -42,12 +42,16 @@ if(TARGET Pti::pti_view)
...
     set(XPU_xpupti_LIBRARY ${SYCL_LIBRARY})
     list(APPEND XPU_xpupti_LIBRARY ${PTI_LIBRARY})
-    set(XPU_xpupti_LIBRARY ${XPU_xpupti_LIBRARY} PARENT_SCOPE)
 
     # find dependent include
     list(APPEND XPUPTI_INCLUDE_DIR ${PTI_INCLUDE_DIR})
     set(XPUPTI_INCLUDE_DIR ${XPUPTI_INCLUDE_DIR} PARENT_SCOPE)
 
+    # find level zero library
+    find_library(LEVEL_ZERO_LIBRARY ze_loader)
```
tsocha, line 51:
```suggestion
    find_library(LEVEL_ZERO_LIBRARY ze_loader REQUIRED)
```

libkineto/src/plugin/xpupti/XpuptiVersionLogger.cpp
```diff
@@ -0,0 +1,47 @@
...
+#include <level_zero/ze_api.h>
+#include <plugin/xpupti/XpuptiVersionLogger.h>
+#include <pti/pti_view.h>
+#include <sycl/version.hpp>
+namespace KINETO_NAMESPACE {
+namespace {
+
+std::string getLevelZeroVersion() {
+  uint32_t version = 0;
+  auto result = zeInit(0);
```
tsocha, line 18: Check if we need this `0` here.
```suggestion
  auto result = zeInit();
```

tsocha (changes requested):
+1 @moksiuc comment

libkineto/src/plugin/xpupti/XpuptiVersionLogger.cpp
```diff
@@ -0,0 +1,47 @@
...
+  if (result == ZE_RESULT_SUCCESS) {
+    uint32_t driverCount = 0;
+    zeDriverGet(&driverCount, nullptr);
+    if (driverCount > 0) {
+      ze_driver_handle_t driver;
+      zeDriverGet(&driverCount, &driver);
```
aostrowski-hbn, line 24: This will only work if driverCount == 1. Instead of a single `driver` we should pass an `vector` of `driverCount` elements.

libkineto/src/plugin/xpupti/XpuptiVersionLogger.h
```diff
@@ -0,0 +1,19 @@
...
+ * LICENSE file in the root directory of this source tree.
+ */
+
+#include <VersionLogger.h>
+
+namespace KINETO_NAMESPACE {
+
+class XpuVersionLogger : public DeviceVersionLogger {
+ public:
+  XpuVersionLogger(std::recursive_mutex& mutex) : DeviceVersionLogger(mutex) {}
```
aostrowski-hbn, line 15: Almost always, single-argument constructors should be marked `explicit` to avoid accidental implicit conversions.

Same goes for implementations for other vendors.

PR #1206, opened 2025-12-16
Add Lintrunner to Kineto

sraikund16:
Summary: Adds implementation for lintrunner in kineto. Now OSS users can run lintrunner instead of internal changes being needed.

Differential Revision: D89324115

### DIFF
--- a/.lintrunner.toml
+++ b/.lintrunner.toml
@@ -1,0 +1,85 @@
+[[linter]]
+code = 'CLANGFORMAT'
+include_patterns = [
+    'libkineto/**/*.cpp',
+    'libkineto/**/*.h',
+    'libkineto/**/*.cc',
+    'libkineto/**/*.hpp',
+]
+exclude_patterns = [
+    'libkineto/third_party/**',
+    'libkineto/build/**',
+]
+command = [
+    'python3',
+    'tools/linter/adapters/clangformat_linter.py',
+    '--binary=clang-format',
+    '@{{PATHSFILE}}'
+]
+is_formatter = true
+
+[[linter]]
+code = 'NEWLINE'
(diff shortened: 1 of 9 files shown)


sraikund16:
@moksiuc This PR should be useful in formatting in OSS

.lintrunner.toml
```diff
@@ -0,0 +1,85 @@
...
+command = [
+    'python3',
+    'tools/linter/adapters/clangformat_linter.py',
+    '--binary=clang-format',
+    '@{{PATHSFILE}}'
+]
+is_formatter = true
+
+[[linter]]
+code = 'NEWLINE'
```
scotts, line 22: I'm surprised we need this special command, including the Python script of ours that we're calling. This surprise also applies to newlines, spaces and tabs. Doesn't clangformat just handle all of that? I think that if we just call the clangformat script, all of this below will be handled. Or am I missing something?

aaronenyeshi (changes requested):

aaronenyeshi (approved):

.lintrunner.toml
sraikund16: They allow you to add some custom rules that clang-format doesn't add. You can see all the linters here for PyTorch itself. We are essentially doing a subset of them: https://github.com/pytorch/pytorch/tree/main/tools/linter/adapters

scotts: @sraikund16, understood that we can add custom rules. But my question is doesn't clang-format already handle spaces, newlines and tabs? That is, given that we're running clang-format, I would expect these additional rules to be a no-op.

briancoutinho:
@sraikund16  Thanks for adding this!

PR #1214, opened 2026-01-06
update cmake to make it mutually exclusive for backends

divyanshk:
Right now `cmakelists.txt` treats each backend independently. I think if someone has multiple SDK installed it will try to include multiple backends.

Note: right now we still try to detect AIUPTI unconditionally. Not updating that logic in this PR.


scotts (approved):
This is much better - thank you!

PR #1221, opened 2026-01-19
Profile XPU_DRIVER activities similarly to CUDA_DRIVER

moksiuc:
Add XPU_DRIVER activity to the profiler so it reports XPU L0 driver activities.
It is counterpart to CUDA_DRIVER activity.

Add gtest.


moksiuc:
@gujinghui @EikanWang

PR #1222, opened 2026-01-20
Add Linux CPU-only CI build and tests

scotts:
This PR:

1. Renames existing CUDA build and test to `linux_cuda.yml`.
2. Separates the build and Kineto unit tests step.
3. Creates `linux_cpu.yml` for CPU-only build and test.

Note that the CPU-only tests run on linux.12xlarge. If we use a smaller system, the PyTorch build will take much longer. Initially, I was using smaller systems while still doing parallel builds and the PyTorch build kept getting killed around 20 minutes in because of OOMs.

I also experimented with not pulling a docker image and trying to use test-infra in a different way. That's still something we should do, but what I was trying was not obviously going to work quickly. We can figure that out later, as I think there's value in getting these CI jobs up as soon as possible.

The CPU-only build and test is also a copy-paste of the existing CUDA tests, with some minor changes. I think that's also fine for now. I think cleaning that up is connected to using test-infra better. Again, I'd like to merge this quickly so we can make sure we have these signals and do this better later (but soon).

### DIFF
--- a/.github/workflows/linux_cpu.yml
+++ b/.github/workflows/linux_cpu.yml
@@ -62,13 +62,13 @@ jobs:
           # Note that we are making these CPU-only builds through cmake flags.
           docker exec -t -w "/kineto/build_static" "${container_name}" bash -c "
             set -eux
-            cmake -DLIBKINETO_NOCUPTI=1 -DKINETO_LIBRARY_TYPE=static ../libkineto/
+            cmake -DLIBKINETO_NOCUPTI=1 -DLIBKINETO_NOROCTRACER=1 -DLIBKINETO_NOXPUPTI=1 -DLIBKINETO_NOAIUPTI=1 -DKINETO_LIBRARY_TYPE=static ../libkineto/
             make -j
           "
 
           docker exec -t -w "/kineto/build_shared" "${container_name}" bash -c "
             set -eux
-            cmake -DLIBKINETO_NOCUPTI=1 -DKINETO_LIBRARY_TYPE=shared ../libkineto/
+            cmake -DLIBKINETO_NOCUPTI=1 -DLIBKINETO_NOROCTRACER=1 -DLIBKINETO_NOXPUPTI=1 -DLIBKINETO_NOAIUPTI=1 -DKINETO_LIBRARY_TYPE=shared ../libkineto/
             make -j
           "
 


.github/workflows/linux_cpu.yml
```diff
@@ -0,0 +1,142 @@
...
+          echo GITHUB_SHA        = $GITHUB_SHA
+          echo GITHUB_REF        = $GITHUB_REF
+
+      - name: Build libkineto (static and shared library)
+        run: |
+          container_name=$(docker run \
+            ${GPU_FLAG:-} \
+            --tty \
+            --detach \
+            --shm-size=32gb \
```
divyanshk, line 48: Was this chosen for a specific reason ? I don't know what is the setting in torchdata or codec flows so curious

jiannanWang:
LGTM. Let's wait for the CIs to finish.

.github/workflows/linux_cpu.yml
```diff
@@ -0,0 +1,142 @@
...
+            set -eux
+            mkdir build_static
+            mkdir build_shared
+            pip install cmake
+          "
+
+          # Note that we are making these CPU-only builds through cmake flags.
+          docker exec -t -w "/kineto/build_static" "${container_name}" bash -c "
+            set -eux
+            cmake -DLIBKINETO_NOCUPTI=1 -DKINETO_LIBRARY_TYPE=static ../libkineto/
```
divyanshk: We should add other flags as well ? https://github.com/divyanshk/kineto/blob/main/libkineto/CMakeLists.txt#L90

(not a big fan of this - I can consolidate them later)

.github/workflows/linux_cpu.yml
scotts: Pure copy-pasta from the CUDA files. :) I started trying to refactor this a bit, but that become its own thing, and I decided to cut my losses and get this up as-is. I'm not sure if it's worth refactoring this bit in particular, since we may move to using test-infra in a way that they manage the docker container for us.

.github/workflows/linux_cpu.yml
scotts: In principle, yes, in practice, we get CPU-only builds. I also know those are not available in this docker container. I think we should also *verify* that we have CPU-only builds, but I'd like that also to be a part of refactoring these later.

scotts: Ya know what, let's just add the flags. :)

scotts: Related: we might want to define some form of `CPU_ONLY` flag that we can set that inside the build we error if any GPUs were set, and then explicitly set all known GPU flags to 0.

PR #1231, opened 2026-01-21
Add basic Rocm CI build and test

scotts:
Depends on #1222.

### DIFF
--- a/.github/workflows/linux_rocm.yml
+++ b/.github/workflows/linux_rocm.yml
@@ -6,7 +6,6 @@ on:
       - ciflow/rocm/*
     branches:
       - main
-  pull_request:
 
 env:
   DOCKER_IMAGE: "pytorch/manylinux2_28-builder:rocm7.2"


.github/workflows/linux_rocm.yml
```diff
@@ -0,0 +1,136 @@
...
+      - main
+  pull_request:
+
+# This is the docker image that pytorch/pytorch uses for ROCM CI.
+env:
+  DOCKER_IMAGE: "ghcr.io/pytorch/ci-image:pytorch-linux-noble-rocm-n-py3"
+
+jobs:
+  pr-test:
+    runs-on: linux.rocm.gpu.gfx942
```
jithunnair-amd: Should be `linux.rocm.gpu.gfx942.1` if you want a single-GPU runner

scotts: @jithunnair-amd, will that enable the job to run? It hasn't been successfully scheduled yet.

scotts:
@jithunnair-amd, my workflow is now running, but I haven't been able to successfully pull a ROCm image. Can you advise me on what image to pull and through what mechanisms?

.github/workflows/linux_rocm.yml
jithunnair-amd: Yes

.github/workflows/linux_rocm.yml
```diff
@@ -0,0 +1,142 @@
...
+
+on:
+  push:
+    branches:
+      - main
+  pull_request:
+
+# This is the docker image that pytorch/pytorch uses for ROCM CI.
+env:
+  DOCKER_IMAGE: "ghcr.io/pytorch/ci-image:pytorch-linux-noble-rocm-n-py3"
```
jithunnair-amd: @scotts If we want to use the same image that's used by PyTorch CI for ROCm, there's a sha suffix that is also part of the tag e.g. `pytorch-linux-noble-rocm-n-py3-f2bf68fc1a48c33b01facc4b2e0227aa3da827e0`. However, this is ever-changing, and is *not* maintainable. Alternatives:
1. Work with @atalman to update the publishing workflow, so that every time it pushes out a new image for `pytorch-linux-noble-rocm-n-py3-*` tag, it also publishes it as `pytorch-linux-noble-rocm-n-py3-latest` tag or something.
2. Use a different image such as `pytorch/manylinux2_28-builder:rocm7.2`, which will still be a maintenance burden to update the ROCm versions, but with less frequency than the SHA updates in the CI image. That part can again be addressed by automatically publishing the same image with a tag `pytorch/manylinux2_28-builder:rocm-latest`. Also, this image usually updates ROCm version *before* the CI image, so you might have potential days of mismatch in the ROCm version between pytorch CI and kineto CI, if that matters.

scotts: @jithunnair-amd, thanks for taking a look!

Since my short-term goal is to just get some tests up and running to check regressions, I'm happy doing the easiest thing for now. I tried pulling `pytorch/manylinux2_28-builder:rocm7.2`, but I'm getting these failures: https://github.com/pytorch/kineto/actions/runs/21452690799/job/61785302020?pr=1231

It looks like I don't have permissions. Is this because this is running in pytorch/kineto and not pytorch/pytorch? There's also some `aws: command not found` errors, but these same patterns work fine for the CPU and CUDA workflows. Any help is greatly appreciated!

jithunnair-amd: It's because it's assuming the image is on ghcr.io, whereas it's actually on dockerhub...

The "aws: command not found" errors can be ignored, but you'll probably need an AWS login via OIDC step to access any S3 buckets as part of the workflow e.g. https://github.com/pytorch/pytorch/blob/2b610902b46d0cbbb64923544f37feec0139cdfa/.github/actions/setup-rocm/action.yml#L115

jithunnair-amd: You'll also need to add a permissions section to your workflow with `id-token: write` for the OIDC step to work successfully e.g.
https://github.com/pytorch/pytorch/blob/2b610902b46d0cbbb64923544f37feec0139cdfa/.github/workflows/_rocm-test.yml#L73

jithunnair-amd: I see you're already calling `setup-rocm` in your workflow, so I think it'll try to do the AWS OIDC login step as well. But you'll probably still need the permissions clause.

.github/workflows/linux_rocm.yml
```diff
@@ -2,9 +2,10 @@ name: Linux ROCm build and test
 
 on:
   push:
+    tags:
+      - ciflow/rocm/*
```
jithunnair-amd, line 6: @scotts A few more steps to get this system working:
1. Create a `ciflow/rocm` label on kineto repo (to be used on PRs)
2. Add pytorch-probot.yaml to kineto repo like https://github.com/meta-pytorch/monarch/blob/main/.github/pytorch-probot.yml

jithunnair-amd: I can't add labels on Kineto PRs. Please try adding the `ciflow/rocm` label and we can see if it works as expected

.github/workflows/linux_rocm.yml
```diff
@@ -0,0 +1,141 @@
+name: Linux ROCm build and test
+
+on:
+  push:
+    tags:
+      - ciflow/rocm/*
+    branches:
+      - main
+  pull_request:
```
atalman:
```suggestion
```

atalman: Lets remove pull_request and try to land this

PR #1236, opened 2026-01-27
D91430608: [Kineto] Add PROFILE_CAPTURE_OVERLOAD_NAMES to Kineto OD config

ryanzhang22:
Summary: This option is available in auto-trace (under ExperimentalConfig -> capture_overload_names) but not in OD. By enabling it, the profile is instructed to include ATen overload names.

Differential Revision: D91587687


aaronenyeshi (approved):

PR #1237, opened 2026-01-27
Introduce callback registry for callback management

divyanshk:
### Summary
**Central source of truth** — Consolidates scattered CUPTI callback metadata (flow correlation, blocklist status) into a single registry. This will replace hardcoded checks throughout the codebase.

**Minimal API** — Exposes only 2 methods: requiresFlowCorrelation(domain, cbid) for CPU→GPU trace arrows and isBlocklisted(domain, cbid) for filtering noisy callbacks. Although we can add more, the overall idea should be to keep it minimal.

**O(1) lookups** — Uses separate std::unordered_map per domain (RUNTIME/DRIVER) with range support for memory operations, initialized lazily as a singleton.

**Usage** — TBD but this would be initialized before the profiling stage when deciding which callbacks to hook on to; and then used later when activity buffers are collected and are to be processed. 

This PR in itself brings no material change to kineto. Future PR will refactor files to make use of this registry.

Note: this is different from `CuptiCallbackApi`'s `registerCallback` which register generic function pointers to be called when CUPTI fires. With `CuptiCallbackRegistry`, we want to store metadata/properties about CUPTI callbacks to help with post-processing (filtering, flow correlation checks, etc).

### DIFF
--- a/libkineto/libkineto_defs.bzl
+++ b/libkineto/libkineto_defs.bzl
@@ -13,6 +13,7 @@ def get_libkineto_cupti_srcs(with_api = True):
     return [
         "src/CuptiActivityApi.cpp",
         "src/CuptiCallbackApi.cpp",
+        "src/CuptiCallbackRegistry.cpp",
         "src/CuptiEventApi.cpp",
         "src/CuptiMetricApi.cpp",
         "src/CuptiRangeProfiler.cpp",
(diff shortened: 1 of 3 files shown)


scotts:
Can you show some examples of what using this new code? That is, this is the foundation for a future refactoring. Can you show some before and after for that refactoring?

divyanshk (commented):
Landed separately here https://github.com/pytorch/kineto/commit/f93df2c6640c2734de1e255210ff928dbb7ca0a3

Closing this PR!

PR #1244, opened 2026-02-03
Refactor CUDA workflow to better use test-infra

scotts:
The workflow on main:

1. Manually specifies a docker image that happens to be from torchbench. I do not know if it is maintained.
2. Does not specify the CUDA version; I do not know what version is being used.
3. Manually invokes `docker` to execute all commands.

The new workflow uses [pytorch/test-infra/.github/workflows/linux_job_v2.yml@main](https://github.com/pytorch/test-infra/blob/a79eb7c044afe4ceb7699acf3ba38e4bd19e4811/.github/workflows/linux_job_v2.yml) which allows us to specify:

1. That we're using `cuda` as our GPU.
2. The CUDA version.
3. We don't have to decide a docker image; test-infra does that for us.
4. We also don't need to explicitly setup CUDA.

We should be able to eventually refactor the CPU and ROCm workflows into a single workflow with a support matrix.

**However**, there is one major downside: each step is part of a giant script, so when there is a failure, it's more work for us to find out what went wrong. I do think this tradeoff is worth it.

Note that this PR also deselects `TestProfiler::test_python_gc_event`. That's being tracked with #1253.


aaronenyeshi (approved):

PR #1255, opened 2026-02-11
[BUG] Clear GPU activities during warmup for iteration-based profiling

zyzshishui:
## Problem
In `RunloopState::Warmup`, Kineto only cleared GPU activities for timestamp-driven flow (`currentIter < 0`), but not for iteration-driven flow (`currentIter >= 0`).

For long iteration-based warmup phases, warmup GPU events can accumulate in backend buffers.  
This can lead to dropping events from the active window (identified on ROCm/roctracer, where traces ended up CPU-only).

## Root Cause
`GenericActivityProfiler::performRunLoopStep()` did not clear warmup GPU activities in iteration-based mode.

## Fix
In the warmup branch, clear GPU activities when:
- profiling is iteration-based, and
- `currentIter >= 0`

to avoid warmup-only GPU events consuming buffer capacity before active collection starts.

### DIFF
--- a/libkineto/src/GenericActivityProfiler.cpp
+++ b/libkineto/src/GenericActivityProfiler.cpp
@@ -833,14 +833,12 @@ time_point<system_clock> GenericActivityProfiler::performRunLoopStep(
     case RunloopState::Warmup:
       VLOG(1) << "State: Warmup";
       warmup_done = derivedConfig_->isWarmupDone(now, currentIter);
-      if (!cpuOnly_ && derivedConfig_->isProfilingByIteration() &&
-          currentIter >= 0) {
-        clearGpuActivities();
-      }
-      // Flushing can take a while so avoid doing it close to the start time
-      if (!cpuOnly_ && currentIter < 0 &&
+      // Flushing can take a while so avoid doing it close to the start time.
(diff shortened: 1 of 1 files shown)


libkineto/src/GenericActivityProfiler.cpp
```diff
@@ -833,6 +833,10 @@ time_point<system_clock> GenericActivityProfiler::performRunLoopStep(
     case RunloopState::Warmup:
       VLOG(1) << "State: Warmup";
       warmup_done = derivedConfig_->isWarmupDone(now, currentIter);
+      if (!cpuOnly_ && derivedConfig_->isProfilingByIteration() &&
```
divyanshk: This if check and the one after seems to have an overlap - can we combine them, or make it cleaner ?

zyzshishui:
> This fix seems correct! I will kick off some more testing. Left a small comment.

Good catch, combined them into a single if. The combined condition looks a bit non-obvious because the two original checks are asymmetric. LMK if you prefer two separate but more explicit checks for readability, or feel free to edit this pr yourself

divyanshk (commented):
Cool, making some changes  - will update this PR accordingly.

Issue #949, opened 2024-06-13
Upgrade to CUDA 12.4 is causing segfaults in 4 Range Profiler Tests


sraikund16:
After upgrading the CUDA image to 12.4 we are having segfault failures in the following tests:
	 19 - CuptiRangeProfilerApiTest.asyncLaunchUserRange (SEGFAULT)
	 20 - CuptiRangeProfilerApiTest.asyncLaunchAutoRange (SEGFAULT)
	 24 - CuptiRangeProfilerTest.UserRangeTest (SEGFAULT)
	 25 - CuptiRangeProfilerTest.AutoRangeTest (SEGFAULT)
	 
This is what the backtrace looks like:
Program received signal SIGSEGV, Segmentation fault.
0x00007ffff730f047 in ?? () from /usr/local/cuda-12.2/lib64/libcupti.so.12
(gdb) bt
#0  0x00007ffff730f047 in ?? () from /usr/local/cuda-12.2/lib64/libcupti.so.12
#1  0x00007ffff72f728a in cuptiFinalize () from /usr/local/cuda-12.2/lib64/libcupti.so.12
#2  0x0000000000448a6e in libkineto::CuptiRBProfilerSession::deInitCupti()::{lambda()#1}::operator()() const [clone .isra.0] ()
#3  0x0000000000448cf9 in libkineto::CuptiRBProfilerSession::~CuptiRBProfilerSession() ()
#4  0x000000000041fd20 in libkineto::MockCuptiRBProfilerSession::~MockCuptiRBProfilerSession() ()
#5  0x000000000041d9a7 in CuptiRangeProfilerTest_UserRangeTest_Test::TestBody() ()
#6  0x00000000004cac2d in void testing::internal::HandleExceptionsInMethodIfSupported<testing::Test, void>(testing::Test*, void (testing::Test::*)(), char const*) ()
#7  0x00000000004bd21e in testing::Test::Run() [clone .part.0] ()
#8  0x00000000004bd935 in testing::TestInfo::Run() [clone .part.0] ()
#9  0x00000000004be395 in testing::TestSuite::Run() [clone .part.0] ()
#10 0x00000000004bfb27 in testing::internal::UnitTestImpl::RunAllTests() ()
#11 0x00000000004bddfb in testing::UnitTest::Run() ()
#12 0x00000000004169a4 in main ()
(gdb) 

cc @briancoutinho

nadavrot:
I can reproduce this issue with 12.4 and 12.8.

[...]

PR #1262, opened 2026-02-18
Re-enable unit tests

scotts:
These tests were disabled because of #949, which has since been resolved.


libkineto/test/CMakeLists.txt
```diff
@@ -122,9 +116,11 @@ target_link_libraries(PidInfoTest PRIVATE
 gtest_discover_tests(PidInfoTest)
 
 # CuptiProfilerApiTest
+# TODO: Re-enable once runtime crash in cuptiDeviceGetChipName is fixed.
```
scotts, line 119: I looked into re-enabling this test as well, but:

1. It's a failure independent of #949.
2. The test actually crashes now.

I'd rather leave that to a follow-up.

PR #1264, opened 2026-02-18
Enable CUPTI teardown by default for CUDA 12.6+

sraikund16:
Summary: Previously, CUPTI teardown (calling cuptiFinalize() after profiling) was only enabled when the TEARDOWN_CUPTI=1 env var was explicitly set. This meant users on CUDA 12.6+ still experienced post-profiling QPS degradation unless they knew to set the env var. This change makes teardown the default for CUDA 12.6+, where it is known to work reliably, while keeping it off by default for older versions. The TEARDOWN_CUPTI env var continues to override the default in either direction for all CUDA versions.

Differential Revision: D93661386


divyanshk (approved):

aaronenyeshi (approved):

PR #1313, opened 2026-03-18
Clean XpuptiActivityApi::singleton() and virtual keyword usage

moksiuc:
Stop methods [push|pop]CorrelationID being static and getting api singleton.
Remove unnecessary virtual keywords.


moksiuc:
@EikanWang, @gujinghui

gujinghui:
The original code aligns with CuPTI and others. Let's keep it as is unless we must have it in future.

moksiuc:
> The original code aligns with CuPTI and others. Let's keep it as is unless we must have it in future.

Maybe it is necessary in their case.
In our code it is unjustified and should be like after proposed modification. 
The modified parts are entirely internal xpu code and have no impact on CuPTI and others code.

gujinghui:
This change is not necessary. Our code is aligned with other device vendors. Considering future change in kineto main, let's do not adopt this change, and close this PR.

PR #1318, opened 2026-03-20
Fix missing wait_on_cuda_event_record_corr_id in Event Synchronize activities

jiannanWang:
CUPTI provides no ordering guarantees for activity records in its buffers. SYNCHRONIZATION records (Event Synchronize, Stream Wait Event) can appear before the CUDA_EVENT record that populates the waitEventMap() with the source stream and correlation ID.

  The existing code performed the waitEventMap() lookup eagerly at processing time. Stream Wait Event was already deferred (for stream-filtering purposes), so it happened to work — but Event Synchronize was logged immediately. When its CUDA_EVENT record hadn't been seen yet,  wait_on_cuda_event_record_corr_id and wait_on_stream were left as -1.

  Fix:
  - Move the waitEventMap() lookup inside the deferred lambda so it runs after all CUDA_EVENT records have been processed.
  - Defer both Event Synchronize and Stream Wait Event (change isWaitEventSync → isEventSync for the defer condition).
  - Add an isWaitEvent flag to DeferredLogEntry so that only Stream Wait Event entries are filtered out on unseen streams, while Event Synchronize entries are always logged.

  Test

  SyncEventCorrIdOutOfOrder — places SYNCHRONIZATION records before their CUDA_EVENT record in the mock buffer and verifies that wait_on_cuda_event_record_corr_id and wait_on_stream are correctly populated for both sync types.

Results
Before fix:
Screenshot 2026-04-02 at 3 46 57 PM

After fix:
Screenshot 2026-04-02 at 3 47 16 PM


libkineto/src/CuptiActivityProfiler.cpp
```diff
@@ -360,21 +360,22 @@ void CuptiActivityProfiler::handleCudaSyncActivity(
...
-    auto maybe_wait_event_info =
-        getWaitEventInfo(activity->contextId, activity->cudaEventId);
-    if (maybe_wait_event_info) {
-      src_stream = maybe_wait_event_info->stream;
-      src_corrid = maybe_wait_event_info->correlationId;
-    }
-  }
+  auto ctx_id = activity->contextId;
+  auto event_id = activity->cudaEventId;
+  auto sync_type = activity->type;
```
scotts: Nit: for most types, including number types, I prefer stating their explicit type rather than using `auto`. It's often useful information for the reader. I tend to use `auto` when the actual type is cumbersome to write and read, such as the iterator on a container.

libkineto/test/CuptiActivityProfilerTest.cpp
```diff
@@ -183,13 +183,30 @@ struct MockCuptiActivityBuffer {
...
+    activities.push_back(reinterpret_cast<CUpti_Activity*>(&act));
+  }
+
+  void addCudaEventActivity(
+      int64_t correlation,
+      uint32_t eventId,
+      uint32_t streamId = 1,
+      uint32_t contextId = 0) {
+    auto& act = *(CUpti_ActivityCudaEventType*)malloc(
+        sizeof(CUpti_ActivityCudaEventType));
```
scotts: We already have an established pattern of calling a family of `createActivity<>()` functions. Let's create another one that takes the extra parameters we need.

I don't like that here and in the existing `createActivity<>()` definition we're calling `malloc()` instead of `new`, but let's deal with that later. There might be a valid reason, so it's better to handle in a follow-up PR.

libkineto/test/CuptiActivityProfilerTest.cpp
```diff
@@ -635,6 +652,97 @@ TEST_F(CuptiActivityProfilerTest, SyncTrace) {
...
+  gpuOps->addSyncActivity(
+      start_time_ns + 120,
+      start_time_ns + 140,
+      kEvtSyncCorrId,
+      CUPTI_ACTIVITY_SYNCHRONIZATION_TYPE_EVENT_SYNCHRONIZE,
+      -1,
+      kEventId);
+  // CUDA_EVENT record comes AFTER the sync records
+  gpuOps->addCudaEventActivity(kRecordCorrId, kEventId, 1, 0);
+  cuptiActivities_.activityBuffer = std::move(gpuOps);
```
scotts, line 713: Can we put a comment before this sequence explaining what's going on and the important relationship we're establishing? I kinda get it, but I'm not sure on the details. A lot of this is just me learning how all of this works! :)

My understanding at the moment:

1. start_time + 10: CUDA_LAUNCH_KERNEL activity indicates we initiated that on the CPU side (maybe?).
2. start_time + 30: kernel activity indicates that the kernel started on GPU
3. start_time + 100: kernel on GPU issued stream wait
4. start_time + 120: kernel on GPU issued synchronize

And then we record the relationship between these events in the buffer _after_ all of this happens. My questions:

1. Is my narrative above correct?
2. We use `kEventId` for both sync activities. That surprises me, because I would have assumed that was unique for each event. But it's the "correlation" IDs that are unique for each event. Can you help explain the relationship between them?

libkineto/src/CuptiActivityProfiler.cpp
jiannanWang: Talking about types, I noticed there are implicit type conversion in the current kineto code base (e.g., int32 to uint32, int64 to uint32). Is that something we should fix? Thanks!

libkineto/test/CuptiActivityProfilerTest.cpp
jiannanWang: createActivity must take in start_ns, end_ns, correlation as arguments. However some events doesn't have start_ns and end_ns. That's why the existing function addCorrelationActivity also use malloc instead of createActivity. Therefore we could create another funciton without start_ns and end_ns and replace the malloc in addCorrelationActivity and here to use the new function.

libkineto/test/CuptiActivityProfilerTest.cpp
jiannanWang: For (1), it is correct. The key to this test is to make a sequence where wait events is provided before synchronization in the activity buffer.

For (2), event ID and correlation ID are different.
The correlation ID is provided by CUPTI and is unique per activity.
The event ID is to label a CUDA event object and is reusable (the same event can be recorded/waited on multiple times).

Example:
corrId=100: cudaEventRecord(event42)        — record we want to reference
corrId=101: cudaStreamWaitEvent(event42)    — the wait event
corrId=200: cudaEventRecord(event42)        — a later re-record of the same event

All three use the same event ID because they refer to the same CUDA event object (event42). But since these are three separate activities, they have three different correlation IDs.

libkineto/src/CuptiActivityProfiler.cpp
scotts: Yes, please! I tried to take care of all of those recently, but it looks like a lot are still left over.

libkineto/test/CuptiActivityProfilerTest.cpp
scotts: @jiannanWang, yes, let's create a new version of `createActivity<>()` that takes the parameters that we need. We can implement it using `malloc()`, since that's the existing pattern. In a future PR, let's refactor the `malloc()` call to use `new` instead.

libkineto/test/CuptiActivityProfilerTest.cpp
scotts: Interesting, that's the opposite terminology I think I would have used! I need to read up on this more. Thanks for the explanation! :)

libkineto/src/CuptiActivityProfiler.cpp
```diff
@@ -397,12 +398,15 @@ void CuptiActivityProfiler::handleCudaSyncActivity(
         setGpuActivityPresent(true);
       };
 
-  if (isWaitEventSync(activity->type)) {
-    // Defer logging wait event syncs till the end so we only
-    // log these events if a stream has some GPU kernels on it.
+  if (isEventSync(activity->type)) {
```
ryanzhang22, line 397: nit: since we defined sync_type above, should we use it here?

libkineto/test/CuptiActivityProfilerTest.cpp
scotts: @jiannanWang, calling out that an explanatory comment before the sequence of events would be helpful. We don't need the comment to explain the sequence, as that's in the code. Rather, the comment should call out the particular thing that's unusual. In your words:
> The key to this test is to make a sequence where wait events is provided before synchronization in the activity buffer.

libkineto/src/CuptiActivityProfiler.cpp
jiannanWang: Good catch! Actually I realized that sync_type can be inferred from the activity, so it isn’t necessary in this function. I removed sync_type and the other related variables.

libkineto/test/CuptiActivityProfilerTest.cpp
jiannanWang: Done! I'll also make sure to follow that in future added tests.

PR #1375, opened 2026-04-23
Fix fork-time deadlock between ConfigLoader and threadSafeForkPrepare

robieta:
Summary:
Fix a deadlock when `torch._thread_safe_fork` (used by `torch.compile`'s
`SubprocPool`) triggers `threadSafeForkPrepare()` while kineto's ConfigLoader
background thread is active.

The deadlock occurs because:
1. `threadSafeForkPrepare()` acquires exclusive `guardLock`, then calls
   `SingletonVault::destroyInstances()`
2. The `kShutdownConfigLoaderOnSingletonVaultDestroy` callback fires and calls
   `ConfigLoader::stopThread()`, which tries to `join()` the ConfigLoader thread
3. The ConfigLoader thread is blocked trying to acquire a shared
   `ThreadSafeForkGuard` in `DynoConfigLoader::readBaseConfig()` — but cannot
   because `guardLock` is held exclusively by the forking thread

This is a classic AB-BA deadlock: Thread A holds guardLock and waits for
Thread B to join; Thread B waits for guardLock (shared) and cannot exit.

The fix splits `stopThread()` into a non-blocking `signalStop()` (sets
`stopFlag_` + notifies condvar) and the existing `stopThread()` (signal + join).
The singleton vault callback now uses `signalStop()`. The ConfigLoader thread
exits cooperatively on its next loop iteration after `guardLock` is released.
The actual join is deferred to `~ConfigLoader()` at static-destruction time.

Differential Revision: D102071920

### DIFF
--- a/libkineto/src/ConfigLoader.cpp
+++ b/libkineto/src/ConfigLoader.cpp
@@ -166,13 +166,17 @@ void ConfigLoader::startThread() {
   }
 }
 
+void ConfigLoader::signalStop() {
+  stopFlag_ = true;
+  {
+    std::lock_guard<std::mutex> lock(updateThreadMutex_);
+    updateThreadCondVar_.notify_one();
+  }
+}
+
 void ConfigLoader::stopThread() {
+  signalStop();
   if (updateThread_) {
-    stopFlag_ = true;
-    {
-      std::lock_guard<std::mutex> lock(updateThreadMutex_);
-      updateThreadCondVar_.notify_one();
-    }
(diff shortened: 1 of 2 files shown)


robieta:
```
-- Initializing git submodules...
  error: expected submodule path 'third_party/kineto' not to be a symbolic link
  CMake Error at cmake/PreBuildSteps.cmake:52 (message):
    Git submodule initialization failed.  Please run:
```
CPU build failure looks unrelated to my changes.

Issue #1377, opened 2026-04-24
Project won't build due to unused result warning/error


bwalshe:
I tried building the project locally using the instructions [here](git push --set-upstream origin fix_ignore_return_value) but this failed because there are several `warn_unused_result`s in the test files and the `wall` compilation flag has been set. 

I'm on Ubuntu and compiling for CUPTI and using v3.28.3 of CMake. When I run `cmake` I get the following:

```
-- The CXX compiler identification is GNU 13.3.0
-- The C compiler identification is GNU 13.3.0
-- Detecting CXX compiler ABI info
-- Detecting CXX compiler ABI info - done
-- Check for working CXX compiler: /usr/bin/c++ - skipped
-- Detecting CXX compile features
-- Detecting CXX compile features - done
-- Detecting C compiler ABI info
-- Detecting C compiler ABI info - done
-- Check for working C compiler: /usr/bin/cc - skipped
-- Detecting C compile features
-- Detecting C compile features - done
-- Found Python3: /usr/bin/python3 (found version "3.12.3") found components: Interpreter 
--  ROCM_SOURCE_DIR = 
-- Found CUDAToolkit: /usr/local/cuda/targets/x86_64-linux/include (found version "13.2.78") 
-- Performing Test CMAKE_HAVE_LIBC_PTHREAD
-- Performing Test CMAKE_HAVE_LIBC_PTHREAD - Success
-- Found Threads: TRUE  
-- Found CUPTI
-- Found NVPERF: Using built-in FindCUDAToolkit target
--  Building with cupti
-- {fmt} version: 11.2.0
-- Build type: Release
-- Kineto: FMT_SOURCE_DIR = /home/brian/github/kineto/libkineto/third_party/fmt
--  ROCTRACER_INCLUDE_DIR = /include/roctracer
--  DYNOLOG_INCLUDE_DIR = /home/brian/github/kineto/libkineto/third_party/dynolog/
--  IPCFABRIC_INCLUDE_DIR = /home/brian/github/kineto/libkineto/third_party/dynolog//dynolog/src/ipcfabric/
CMake Deprecation Warning at third_party/dynolog/third_party/json/CMakeLists.txt:1 (cmake_minimum_required):
  Compatibility with CMake < 3.5 will be removed from a future version of
  CMake.

  Update the VERSION argument <min> value or use a ...<max> suffix to tell
  CMake that the project does not need compatibility with older versions.


[...]

PR #1378, opened 2026-04-24
Fix warn_unused_result warnings in Kineto test files

ryanzhang22:
Summary:
https://github.com/pytorch/kineto/issues/1377

This user's build system is building these tests and is set up to treat all warnings as errors, so these `warn_unused_result` warnings are causing build failure (which is valid).

Differential Revision: D102340142

### DIFF
--- a/libkineto/test/CuptiActivityProfilerTest.cpp
+++ b/libkineto/test/CuptiActivityProfilerTest.cpp
@@ -75,6 +75,14 @@ const TraceSpan& defaultTraceSpan() {
   static TraceSpan span(0, 0, "Unknown", "");
   return span;
 }
+
+#ifdef __linux__
+void createTempTraceFile(char* filename) {
+  const int fd = mkstemps(filename, 5);
+  ASSERT_GE(fd, 0) << "mkstemps failed for " << filename;
+  close(fd);
+}
+#endif
 } // namespace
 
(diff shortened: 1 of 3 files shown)


bwalshe:
FYI, I know it isn't the focus of this MR, but there is an `fd` opened on line 651 of [CuptiActivityProfilerTest.cpp](https://github.com/pytorch/kineto/blob/main/libkineto/test/CuptiActivityProfilerTest.cpp#L651C7-L651C10)   which does not appear to get closed.

ryanzhang22:
> FYI, I know it isn't the focus of this MR, but there is an `fd` opened on line 651 of [CuptiActivityProfilerTest.cpp](https://github.com/pytorch/kineto/blob/main/libkineto/test/CuptiActivityProfilerTest.cpp#L651C7-L651C10) which does not appear to get closed.

Thanks, is this causing build failures in your setup too? I'll put out a separate PR for it.

bwalshe:
> Thanks, is this causing build failures in your setup too? I'll put out a separate PR for it.

It's not causing any problems for me. It's just something I noticed while looking at the code. 

Actually, there is a function `checkTracefile(const char* filename)` which is used sometimes and then other times the tests directly open the file and check they have at least 100 bytes. If the function was used consistently then the fd would always be closed.

PR #1388, opened 2026-05-04
Replace LIBKINETO_NO* build flags with KINETO_BACKEND

scotts:
Selecting a GPU backend (or no GPU at all) in libkineto's CMake build required setting up to three flags phrased in the negative: `LIBKINETO_NOCUPTI`, `LIBKINETO_NOROCTRACER`, `LIBKINETO_NOXPUPTI`. A CPU-only build had to pass all three: (`-DLIBKINETO_NOCUPTI=1 -DLIBKINETO_NOROCTRACER=1 -DLIBKINETO_NOXPUPTI=1`) The flags were also inconsistent with each other: only `LIBKINETO_NOCUPTI` was a real `option()`; `LIBKINETO_NOXPUPTI` was undeclared and required `if(DEFINED ... AND NOT ...)` everywhere it was checked.

**Why**

The flags grew up incrementally as each backend was added, and the `NO*` framing matched how the surrounding code already gated CUPTI. Meanwhile the C++ source uses positive macros (`HAS_CUPTI`, `HAS_ROCTRACER`, `HAS_XPUPTI`), so the negation lived only in the CMake/CI/PyTorch-glue layer. The mutual exclusivity of the backends was enforced by an `if/elseif` chain on three independent variables instead of being expressed directly.

**Approach**

Introduce a single `KINETO_BACKEND` variable taking one of `cpu`, `cuda`, `rocm`, `xpu` (default `cpu`). The variable directly expresses the "exactly one backend" constraint. Selection is now explicit: if the user picks `cuda`/`rocm`/`xpu` and the corresponding toolkit isn't available, configuration fails fast instead of silently degrading.

The C++ source is unchanged — `HAS_CUPTI`/`HAS_ROCTRACER`/`HAS_XPUPTI` are still emitted by the build, just driven by `KINETO_BACKEND` rather than the inverted flags.

A backwards-compatibility shim in `libkineto/CMakeLists.txt` translates the legacy `LIBKINETO_NO*` flags into `KINETO_BACKEND` so existing callers (notably `pytorch/cmake/Dependencies.cmake`) keep building during the changeover. The shim is marked with a TODO and will be removed once PyTorch migrates to `KINETO_BACKEND`.

The CI scripts under `.github/scripts/config_*.sh` are updated to use the new flag. The existing PyTorch build and test CI jobs check that this will still work when we release these changes upstream.


libkineto/CMakeLists.txt
```diff
@@ -50,53 +50,104 @@ if(NOT ROCM_SOURCE_DIR)
...
-  add_subdirectory(src/plugin/xpupti)
-else()
-  set(LIBKINETO_NOXPUPTI ON)
+unset(_KINETO_BACKEND_VALID)
+message(STATUS "KINETO_BACKEND = ${KINETO_BACKEND}")
+
+# Warn if legacy LIBKINETO_NO* flags disagree with the cached KINETO_BACKEND.
+# The shim only fires on initial configure; once KINETO_BACKEND is in the
+# cache, subsequent legacy-flag changes are silently ignored. Detect that and
+# tell the user. TODO: remove with the rest of the legacy shim.
```
NicolasHug, line 87: you're being too nice!

NicolasHug (approved):

libkineto/CMakeLists.txt
scotts: I'm hoping the window of time this shim will need to exist is less than a week!

scotts:
The ROCm failures are pre-existing. See https://github.com/pytorch/kineto/actions/runs/25532405262/job/74941277250?pr=1391 on https://github.com/pytorch/kineto/pull/1391.

PR #1392, opened 2026-05-08
Update the rocm api filtering

mwootton:
There is a list of chatty hip apis that are suppressed during rocm trace collection.
Updated an api name.
Added some new apis.
Removed some hipEvent apis we might want to see.

### DIFF
--- a/libkineto/src/RocprofLogger.cpp
+++ b/libkineto/src/RocprofLogger.cpp
@@ -380,7 +380,6 @@ int RocprofLogger::toolInit(
   apiList.add("hipPeekAtLastError");
   apiList.add("hipModuleGetFunction");
 
-
   // Get a vector of the enabled api calls
   auto apis = apiList.allEnabled();
 
--- a/libkineto/src/RocprofLogger.cpp
+++ b/libkineto/src/RocprofLogger.cpp
@@ -373,12 +373,13 @@ int RocprofLogger::toolInit(
   apiList.add("__hipPushCallConfiguration");
   apiList.add("__hipPopCallConfiguration");
   apiList.add("hipCtxSetCurrent");
-  apiList.add("hipEventRecord");
-  apiList.add("hipEventQuery");
-  apiList.add("hipGetDeviceProperties");
+  apiList.add("hipGetDevicePropertiesR0600");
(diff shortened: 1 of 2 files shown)


libkineto/src/RocprofLogger.cpp
```diff
@@ -373,12 +373,13 @@ int RocprofLogger::toolInit(
...
-  apiList.add("hipEventQuery");
-  apiList.add("hipGetDeviceProperties");
+  apiList.add("hipGetDevicePropertiesR0600");
+  apiList.add("hipGetDeviceCount");
+  apiList.add("hipDeviceGetAttribute");
+  apiList.add("hipRuntimeGetVersion");
   apiList.add("hipPeekAtLastError");
   apiList.add("hipModuleGetFunction");
-  apiList.add("hipEventCreateWithFlags");
+
```
scotts, line 382: @mwootton, our internal linter is complaining about the added new line here.

libkineto/src/RoctracerLogger.cpp
```diff
@@ -342,13 +342,13 @@ void RoctracerLogger::startLogging() {
...
+    loggedIds_.add("hipGetDevicePropertiesR0600");
+    loggedIds_.add("hipGetDeviceCount");
+    loggedIds_.add("hipDeviceGetAttribute");
+    loggedIds_.add("hipRuntimeGetVersion");
     loggedIds_.add("hipPeekAtLastError");
     loggedIds_.add("hipModuleGetFunction");
-    loggedIds_.add("hipEventCreateWithFlags");
-    loggedIds_.add("hipGetDeviceCount");
-    loggedIds_.add("hipDevicePrimaryCtxGetState");
+
```
scotts, line 351: @mwootton, and here.

mwootton:
2 bytes saved!  Thanks,
Were there directions for how to pre-lint someplace?

scotts:
@mwootton, thanks for the fix!

No, unfortunately, we don't have a good linter solution yet. What I need to do is turn off the internal linter, and then rely on an external linter. We will just reuse the one that's used in PyTorch. That's on me to get that all working properly.

PR #1393, opened 2026-05-08
[xpupti]: emit GPU_USER_ANNOTATION for user record_function ranges

SlawomirLaba:
The XPU PTI plugin already wires user correlation IDs into PTI via
ptiViewPushExternalCorrelationId / PTI_VIEW_EXTERNAL_KIND_CUSTOM_1 and
populates userCorrelationMap_ in handleCorrelationActivity(), but it
never produced any GPU_USER_ANNOTATION events from that information,
so user record_function() ranges did not appear on the device timeline.

After a GPU activity (kernel, memcpy, memset) is logged, look up its
correlation id in userCorrelationMap_ and resolve the originating CPU
activity via the linked-activity callback. The first resolved hit on a
given (device, stream, user_corr_id) emplaces a synthesized
GPU_USER_ANNOTATION GenericTraceActivity; subsequent hits on the same
key widen its [startTime, endTime] so the annotation spans the union
of all GPU activities on that stream that share the user correlation
id, matching the behavior of GpuUserEventMap::insertOrExtendEvent in
the generic profiler. The accumulated annotations are logged once at
the end of processTrace and the per-session map is cleared between
iterations.

The lookup is gated on a successful callback resolution rather than a
separate dedup set, so a transient nullptr return (CPU side not yet
visible when the first GPU activity is processed) is retried on later
GPU activities for the same range.

Only emit the synthesized event when the caller requested
ActivityType::GPU_USER_ANNOTATION, to avoid producing extra events for
clients that do not opt in.

Tests
-----
Extend RunProfilerTest with optional parameters (userCorrelationId,
linkedCpuActivity, linkedActivityCallback) so existing test scaffolding
can drive a session that pushes/pops a user correlation id around the
XPU workload and resolves it back to a CPU-side activity through the
linked-activity callback. Existing tests are unaffected (defaults are
0 / nullptr).

Add XpuptiProfilerTest.GpuUserAnnotation, which enables
GPU_USER_ANNOTATION alongside the existing GPU/runtime activities,
runs the XPU compute helper inside a user correlation range, and
asserts that exactly one synthesized "user_function" annotation is
emitted per participating GPU stream (memcpy stream and kernel
stream).

Add XpuptiProfilerTest.GpuUserAnnotationLinkedActivityRetry, which
uses a callback that returns nullptr on the first lookup and the
linked CPU activity thereafter, and verifies that the plugin still
emits one annotation per stream -- pinning down the
retry-until-resolved contract on the linked-activity callback.


scotts:
cc: @gujinghui, @chuanqi129

gujinghui (approved):

SlawomirLaba:
cc: @scotts 

the internal review is complete
please trigger CI jobs

aostrowski-hbn (approved):

aostrowski-hbn:
@ryanzhang22 @scotts please import and review

ryanzhang22 (approved):

libkineto/src/plugin/xpupti/XpuptiActivityHandlers.cpp
```diff
@@ -295,6 +296,39 @@ void XpuptiActivityProfilerSession::handleRuntimeKernelMemcpyMemsetActivities(
     return;
   }
   trace_activity->log(logger);
+
+  if constexpr (!handleRuntimeActivities) {
```
ryanzhang22, line 309: nit: It would be nice to have some comments for this code (and for the rest of this file, for that matter)

libkineto/test/xpupti/XpuptiProfilerTest.cpp
```diff
@@ -9,13 +9,126 @@
...
+void expectTwoUserAnnotations(const KN::CpuTraceBuffer& buffer) {
+  std::set<std::pair<int64_t, int64_t>> streamKeys;
+  unsigned annotationCount = 0;
+  for (const auto& activity : buffer.activities) {
+    if (activity->type() != KN::ActivityType::GPU_USER_ANNOTATION) {
+      continue;
+    }
+    ++annotationCount;
+    EXPECT_EQ(activity->correlationId(), kUserCorrId);
+    ASSERT_NE(activity->linkedActivity(), nullptr);
```
ryanzhang22, line 121: actually, `linkedActivity` is `cpuAct` from the function above right? It looks like `linkedActivity` is just a raw pointer to `cpuAct` constructed in `RunProfilerTest`, are we sure that it survives to this point?

libkineto/src/plugin/xpupti/XpuptiActivityHandlers.cpp
aostrowski-hbn: I fully agree. We're considering a refactor, so most probably we'll add comments then if that's okay.

libkineto/test/xpupti/XpuptiProfilerTest.cpp
aostrowski-hbn: Looks dangling to me - good catch! I don't have permissions to push to this branch, so let's wait for Slawomir to get back from his vacation to fix this.

SlawomirLaba: yes good catch, it is fixed now

aostrowski-hbn (approved):

gujinghui:
@ryanzhang22  The issue has been addressed. Could you please land it again? Thanks.

ryanzhang22:
LGTM, but Scott is on PTO until Monday and I need an internal approval to finish the land process. Will update when he comes back.

PR #1335, opened 2026-03-30
Add support for oneCCL and Synchronization events.

tsocha:
- Add communication and synchronization profiling support — new activity handlers for collective communication events and device synchronization events
- new `XPU_SYNC` ActivityType
- Add integration tests — test suite for the new communication and synchronization handlers with proper MSVC test discovery via gtest_add_tests
- Harden error handling — log warnings instead of throwing when `PTI_VIEW_COMMUNICATION` is unsupported
- Fix Windows/MSVC build issues — fix `XpuptiProfilerTestLib` linking, ExternalProject Debug/Release mismatch, general MSVC compilation fixes
- correct `XPUPTI_BUILD_FLAG` logging in CMake

Commits marked with 🤖 are made with usage of gen AI.

PR #1394, opened 2026-05-11
[xpupti] Fix Windows build support for xpupti tests and multi-config generators.

tsocha:
It's a part of https://github.com/pytorch/kineto/pull/1335 1/3

* Refactor test/xpupti/CMakeLists.txt to handle Windows and multi-config generators (VS, Ninja Multi-Config):
  - Force Ninja as the inner ExternalProject generator on Windows and propagate $<CONFIG> via CONFIGURE_COMMAND for multi-config
  - Use WINDOWS_EXPORT_ALL_SYMBOLS and /FORCE:UNRESOLVED so the shared test library can build against ComputeOnXpu defined in the compute executable
  - Pass LINK_LIBRARY_NAME/LINK_LIBRARY_DIR instead of a hardcoded lib<name>.so path; inner project resolves via find_library with config-aware PATH_SUFFIXES
  - Forward SYCL lib path to the inner linker on Windows
  - Skip -Wl,--export-dynamic on Windows in compute/CMakeLists.txt

* Rename make_test to make_sycl_test to reflect SYCL-based tests.

* Fix XPUPTI_BUILD_FLAG so -DHAS_XPUPTI is visible in both the current and parent scopes (was only set in PARENT_SCOPE).

* Reorder id/linked/metadata assignments in XpuptiActivityProfilerSession::handleRuntimeKernelMemcpyMemsetActivities and inline the linkedActivity lookup — no behavior change.

* Include src/ActivityBuffers.h in XpuptiTestUtilities.cpp and hoist iterator declarations out of the for-init to satisfy MSVC.

PR #1396, opened 2026-05-11
[xpupti] Add XCCL collective communication activity tracing to XPU plugin

tsocha:
It's a part of https://github.com/pytorch/kineto/pull/1335 3/3

Enable PTI_VIEW_COMMUNICATION collection in the XPU PTI plugin so oneCCL host-side collective operations show up in Kineto traces. Events are emitted as COLLECTIVE_COMM activities named with an "xccl::" prefix and carry the PTI communicator id as metadata.

- Gate new code paths on PTI_VERSION_AT_LEAST(0, 17)
- Wire enable/disable of PTI_VIEW_COMMUNICATION in XpuptiActivityApi
- Add handleCommunicationActivity for pti_view_record_comms records
- Add unit tests covering naming, field mapping, and out-of-range drop
- Document INTEL_LIBITTNOTIFY64 requirement in libkineto/README.md


tsocha:
@gujinghui please review it.

libkineto/src/plugin/xpupti/XpuptiActivityApi.cpp
```diff
@@ -260,6 +261,16 @@ void XpuptiActivityApi::enableXpuptiActivities(
         XPUPTI_CALL(ptiViewEnable(PTI_VIEW_COLLECTION_OVERHEAD));
         break;
 
+#if PTI_VERSION_AT_LEAST(0, 17)
+      case ActivityType::COLLECTIVE_COMM: {
+        auto rc = ptiViewEnable(PTI_VIEW_COMMUNICATION);
+        if (rc != PTI_SUCCESS) {
```
gujinghui: Why we do not follow the existing code style to use XPUPTI_CALL macro?

tsocha: `XPUPTI_CALL` macro throw an error.
I wanted to Log a `WARNING` because oneCCL is not supported on Windows.
I think that the standard `XPUPTI_CALL` can be used. I will fix it.

libkineto/README.md
```diff
@@ -73,6 +73,8 @@ For more information on how to run on-demand profiling, please refer to the Dyno
 
 The default trace output is a JSON file that can be visualized in Chrome Trace Viewer or Perfetto. The trace output is generated by the `ChromeTraceLogger` instance. The `ChromeTraceLogger` writes to a JSON file using `std::ofstream` in `output_json.cpp` to maximize performance during export. This instance is created by the `ActivityProfilerController` and is stored in the `ActivityLoggerFactory` alongside its protocol. Using this schema, Kineto supports multiple trace output formats.
 
+- Intel XCCL: to enable collecting of oneCCL host events, `INTEL_LIBITTNOTIFY64` enviroment variable have to be set as path to `pti_view.so` location.
+
```
scotts, line 75: Why does this need to be in the general instructions? Is this something that's covered in Intel's other docs? I'd like to keep this file short and general.

tsocha: This env variable is required by ITT which is used by PTI to collect these events.
Without this variable user won't see oneCCL events in his trace.

I wanted to expose this information to avoid confusion of the Kineto user.

We are working to remove this requirement in the future but due to performance requirements of PTI integration it's not ready yet.

I could create a new README file inside `xpupti` plugin directory but I'm afraid that it will be hidden.

What do you think?

gujinghui: @tsocha  Do we have public doc to explain this env var?

tsocha: There is some documentation related to V-tune and ITT:
https://www.intel.com/content/www/us/en/docs/vtune-profiler/user-guide/2025-4/attaching-itt-apis-to-a-launched-application.html

Unfortunately it's not directly related to Kineto.
In our case the PTI is a collector.

There is also a documentation of tests for ITT:
https://github.com/intel/ittapi/blob/master/src/ittnotify_refcol/README.md
Again, it's at least 2 levels lower than Kineto and we should not expect that user will know that he need to set an env variable for ITT -> which is used by PTI -> which is used by Kineto.

The best documentation I was able to get is `PTI` Known Issues section of the documentation:
https://intel.github.io/pti-gpu/knownissues.html

I want to avoid confusion from the user.
If the main `README.md` is not a good place for it, maybe we should create another one in `XPUPTI` plugin directory.

tsocha: @scotts
`README.md` is clean again. Now, we emit a runtime warning.

gujinghui (approved):

gujinghui:
@scotts please help review this PR. Thanks.

libkineto/README.md
jfedorov: @scotts  Can you please review this PR. as it said README.md is clean 
and all additional requirements will be documented at [Intel PTI special page](https://intel.github.io/pti-gpu/knownissues.html) in a way specific to PyTorch profiler 
thanks! 

att @gujinghui

gujinghui:
@scotts  @divyanshk @ryanzhang22  
Could you please help review this PR? I would appreciate it if it could be included in the next PyTorch release. Thanks.

ryanzhang22 (approved):

libkineto/src/plugin/xpupti/XpuptiActivityApi.cpp
```diff
@@ -175,6 +178,31 @@ void XpuptiActivityApi::bufferCompleted(
 }
 #endif
 
+namespace {
```
ryanzhang22, line 181: @tsocha can we include this in a `#ifdef HAS_XPUPTI` block? It doesn't seem used outside anywhere that isn't already guarded by this

libkineto/src/plugin/xpupti/XpuptiActivityHandlers.cpp
```diff
@@ -297,6 +297,35 @@ void XpuptiActivityProfilerSession::handleRuntimeKernelMemcpyMemsetActivities(
   trace_activity->log(logger);
 }
 
+void XpuptiActivityProfilerSession::handleCommunicationActivity(
+    const pti_view_record_comms* activity,
+    ActivityLogger& logger) {
+  const auto& activity_record = *activity;
+  const std::string activity_name{activity_record._name};
```
ryanzhang22, line 303: is this guaranteed to be non-null?

libkineto/src/plugin/xpupti/XpuptiActivityApi.cpp
tsocha: This `#ifdef` will be removed in this PR: https://github.com/pytorch/kineto/pull/1394
It's already filtered on `CMake` level.

libkineto/src/plugin/xpupti/XpuptiActivityHandlers.cpp
tsocha: `PTI` should never return `nullptr`.
We plan to refactor this file, currently it's a common pattern here.

ryanzhang22:
@tsocha ack on the responses. Can you take a look at the XPU CI failure before I lint and merge internally.

tsocha:
@ryanzhang22 
> @tsocha ack on the responses. Can you take a look at the XPU CI failure before I lint and merge internally.

I have fixed issue on the CI.
I also made some fixes for Windows.
Some common tests use Linux specific code which break Windows build of tests.

libkineto/test/xpupti/XpuptiActivityHandlersTest.cpp
```diff
@@ -0,0 +1,156 @@
...
+#include "src/plugin/xpupti/XpuptiProfilerMacros.h"
+
+#include <gtest/gtest.h>
+
+namespace KN = KINETO_NAMESPACE;
+using namespace libkineto;
+
+// Mock XpuptiActivityApi that delivers hand-crafted PTI records
+// through the virtual processActivities without needing PTI runtime.
+class MockXpuptiActivityApi : public KN::XpuptiActivityApi {
```
ryanzhang22, line 23: do you need to mock the other methods to make this actually independent of PTI? `enableXpuptiActivities` for example looks like it's called in the constructor of `XpuptiActivityProfilerSession` and has a bunch of PTI enable calls.

tsocha: Unfortunately these other methods are not virtual.
I wanted to avoid modifying the production code just for tests.

PR #1398, opened 2026-05-11
Fix ROCm HtoD memcpy stream attribution

sanrise:
Summary:
**Problem**
AMD traces can render HtoD memcpy activity as if every copy happened on stream 0, even while the nearby GPU work is clearly using real non-zero streams. That makes the trace misleading for the main thing users are trying to understand: whether host-to-device copies overlap with active GPU work, and which stream context issued the copy.

**Why**
The ROCm async memory copy record can arrive without a useful queue id, so Kineto receives the copy as queue 0. At the same time, the HIP runtime stream context around the same work can still identify the real stream that issued the operation. Before this change, Kineto published the unusable queue 0 value directly, so the viewer collapsed those HtoD copies onto stream 0.

**Fix**
This change repairs only the safe case. The shared ROCm stream/queue helper learns an unambiguous HIP stream to non-zero ROCm queue mapping from correlated GPU activity, then uses that mapping to backfill HtoD memcpy rows that would otherwise render as stream 0. Existing non-zero memcpy queues are preserved, and ambiguous mappings stay on stream 0 instead of guessing.

**Caveat [from Michael Wootton]**
This is a Kineto rendering/backfill fix, not proof that an SDMA copy physically executed on the compute queue shown in the UI. Some memory copies may be serviced by SDMA without a queue id, and some may be serviced by a blit kernel. Mapping a queue-less copy onto the HIP stream's GPU queue is therefore virtual attribution: useful for making the trace readable and preserving stream-context overlap, but not a replacement for ROCm reporting an explicit copy queue or for a future dedicated pseudo-queue representation for queue-less copies.

Reviewed By: scotts

Differential Revision: D103952982

### DIFF
--- a/libkineto/src/RocmActivityProfiler.h
+++ b/libkineto/src/RocmActivityProfiler.h
@@ -12,6 +12,8 @@
 //                       need this guard in the header file.
 #ifdef HAS_ROCTRACER
 
+#include <cstdint>
+
 #include "GenericActivityProfiler.h"
 #include "RocLogger.h"
 #ifndef ROCTRACER_FALLBACK
--- a/libkineto/src/RocmStreamQueue.h
+++ b/libkineto/src/RocmStreamQueue.h
@@ -1,0 +1,131 @@
+/*
+ * Copyright (c) Meta Platforms, Inc. and affiliates.
+ * All rights reserved.
+ *
+ * This source code is licensed under the BSD-style license found in the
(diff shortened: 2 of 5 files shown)


sanrise:
@mwootton please take a look.

mwootton:
That looks pretty straightforward.  Just a little light remapping.  If the test passes, it seems fine to me.

PR #1399, opened 2026-05-13
Remove LIBKINETO_NO* compatibility shim

scotts:
The CMake refactor (#1388) replaced `LIBKINETO_NOCUPTI`, `LIBKINETO_NOROCTRACER`, and `LIBKINETO_NOXPUPTI` with a single `KINETO_BACKEND` variable, but kept a shim that translated the legacy flags during the cross-repo migration window. https://github.com/pytorch/pytorch/pull/183388 migrated PyTorch's `cmake/Dependencies.cmake` and `torch/csrc/autograd/init.cpp` have now both been migrated to use `KINETO_BACKEND` / `HAS_*` directly. The shim has no remaining caller.

**Approach**

Delete the shim block and the cached-vs-new-legacy-flags conflict warning that existed only to ease the migration. Replace with a simpler guard that warns when `KINETO_BACKEND` wasn't set at all and the build defaulted to `cpu`. This catches stragglers passing legacy `LIBKINETO_NO*` flags as a side effect — they'll be silently ignored, `KINETO_BACKEND` will default to `cpu`, and the new warning will tell them how to pick a backend explicitly. The new code carries no knowledge of the legacy flag names, so the kineto repo is now free of all `LIBKINETO_NO*` references.

The warning fires only on initial configure (subsequent re-runs find `KINETO_BACKEND` populated in the cache) and on a freshly-wiped build directory.

### DIFF
--- a/libkineto/CMakeLists.txt
+++ b/libkineto/CMakeLists.txt
@@ -55,17 +55,11 @@ endif()
 # Selection is explicit: if the chosen toolkit isn't available, configuration
 # fails. Use KINETO_BACKEND=cpu to build without GPU support.
 
-# Backwards-compat shim: derive KINETO_BACKEND from legacy LIBKINETO_NO* flags
-# when the user hasn't set KINETO_BACKEND directly.
-# TODO: remove after PyTorch migrates to KINETO_BACKEND.
+# Track whether the user explicitly set KINETO_BACKEND before the default
+# fires below, so we can warn when we fall back to cpu.
+set(_kineto_backend_explicit ON)
 if(NOT DEFINED CACHE{KINETO_BACKEND} AND NOT DEFINED KINETO_BACKEND)
-  if(DEFINED LIBKINETO_NOROCTRACER AND NOT LIBKINETO_NOROCTRACER)
-    set(KINETO_BACKEND "rocm" CACHE STRING "")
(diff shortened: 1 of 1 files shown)


libkineto/CMakeLists.txt
```diff
@@ -81,29 +75,12 @@ endif()
...
-      "Legacy flags are ignored. Pass -DKINETO_BACKEND=${_legacy_implied_backend} "
-      "explicitly or remove the build directory to switch backends.")
-  endif()
-  unset(_legacy_implied_backend)
+if(NOT _kineto_backend_explicit)
+  message(WARNING
+    "KINETO_BACKEND was not set; defaulting to 'cpu'. "
+    "Pass -DKINETO_BACKEND={cpu,cuda,rocm,xpu} explicitly to silence this warning.")
 endif()
+unset(_kineto_backend_explicit)
```
NicolasHug, line 83: Curious why it's needed to `unset`?

NicolasHug (approved):

libkineto/CMakeLists.txt
scotts: Not strictly needed, but my understanding is that it's good CMake hygiene to unset temporary-use variables so they don't pollute the rest of the build. In this case, if we didn't unset it, it would also be visible in all of the other builds we kick off from here (XPU plugin, everything in `third_party`, tests and benchmarks).

PR #1414, opened 2026-05-28
Expose ActivityLoggerFactory via public API

trichmo:
This is a working example of the changes necessary to expose registering a LoggerFactory for custom output formatting of Kineto events on save.


libkineto/include/libkineto.h
```diff
@@ -154,4 +156,13 @@ class LibkinetoApi {
 // Singleton
 LibkinetoApi& api();
 
+// Register a custom output format logger with a protocol prefix.
+// Example: registerLoggerFactory("perfetto", [](const std::string& path) {
+//   return std::make_unique<PerfettoLogger>(path);
+// });
+// Users can then call trace.save("perfetto:///tmp/trace.pftrace")
```
scotts: Let's keep this comment generic. Something like `"foo"` and `FooLogger`.

scotts:
Thanks, @trichmo! I think adding this to the public API is useful. Let's add some unit tests for the new API.

libkineto/include/libkineto.h
```diff
@@ -42,20 +42,24 @@ bool hasTestEnvVar();
...
 namespace libkineto {
 
+class ActivityLogger;
 class Config;
 class ConfigLoader;
 
 struct CpuTraceBuffer {
   template <class... Args>
   void emplace_activity(Args&&... args) {
-    activities.emplace_back(std::make_unique<GenericTraceActivity>(std::forward<Args>(args)...));
```
trichmo, line 51: Sorry for the noise here, lintrunner wasn't very contained and wanted these changes

scotts: Can you undo the lint-based changes? We're not actually using the open source linter right now.

trichmo:
@scotts It doesn't seem as though thread safety is something kineto is supporting. I had some tests for that but have removed them - if you'd like me to update for thread safety for registerlogger, I have some stashed changes but they touch existing code a bit instead of just an additive pr.

trichmo:
@scotts Can you rerun CI?
I think the failing test should be fixed by https://github.com/pytorch/pytorch/pull/186832
Not sure if your changes also need to land https://github.com/pytorch/kineto/pull/1428 - I do get these explicitly skipped tests failing locally for me.

scotts:
@trichmo, can you pull the latest to get the CI improvements?

libkineto/include/libkineto.h
```diff
@@ -154,4 +156,11 @@ class LibkinetoApi {
 // Singleton
 LibkinetoApi& api();
 
+// Register a custom output format logger with a protocol prefix.
+// Example: registerLoggerFactory("foo", [](const std::string& path) {
+//   return std::make_unique<FooLogger>(path);
+// });
+// Users can then call trace.save("foo:///tmp/trace.pftrace")
```
scotts: The filename should be generic; something like `trace.foo` to match the rest of the example.

scotts:
Thanks for the tests! I think we should do a lot less, testing, though. This is a simple addition, I think we should just keep:

1. BasicPublicAPI
2. ProtocolCaseInsensitive
3. UnregisteredProtocolThrows
4. OverwriteProtocolWarning
5. BuiltInFileProtocolStillWorks

The rest, I think, are redundant or test unrelated functionality.

trichmo:
Makes sense, trimmed it down. Thanks!

libkineto/include/libkineto.h
scotts: @trichmo, this is the only remaining issue.

trichmo:
@scotts gentle ping if things look okay, could you kick off CI?
Thanks!

scotts:
@trichmo, this is causing some internal build failures because we're including `Logger.h` in places that it previously was not included. That indicates we have real problems with that file, but I think in the short term if we can avoid adding it as an include, that would speed up merging. I can try commandeering and fixing later in the week if you're not able to address it.

PR #1416, opened 2026-05-29
Repair ROCm kernel stream attribution (#1416)

sanrise:
Summary:

**Problem**

ROCm Kineto traces can show kernels from independent HIP streams on the wrong stream, making unrelated worker activity look mixed together.

**Why**

ROCProfiler can deliver async kernel dispatch records where multiple kernels share one correlation and another runtime launch correlation is missing. Trusting correlation alone means Kineto can collapse or spread kernels onto the wrong stream. The repaired ROCm stream id is a 64-bit HIP stream value, and GPU user annotations also need to preserve that full value so Perfetto keeps annotations nested on the same lane as the kernels they cover.

**Fix**

Keep the HIP stream and demangled kernel name from runtime callbacks, carry dispatch thread and external stream data on async rows, and repair ambiguous duplicated kernel correlations only when runtime thread or canonical kernel name maps to exactly one stream. Preserve the full 64-bit resource id when creating generic GPU user annotations so `Stage::...` annotations stay aligned with repaired ROCm kernel lanes. The implementation limits overhead by building copy and kernel repair maps only for activity types requested by the trace, building the kernel-name map only when duplicated async kernel correlations are present, and tracking pushed stream correlations in cached kernel args instead of adding another thread-local hash set per kernel launch. If the evidence is ambiguous, Kineto preserves the original async stream instead of guessing.

Differential Revision: D104473177

### DIFF
--- a/libkineto/include/GenericTraceActivity.h
+++ b/libkineto/include/GenericTraceActivity.h
@@ -142,7 +142,7 @@ class GenericTraceActivity : public ITraceActivity {
   int64_t endTime{0};
   int32_t id{0};
   int32_t device{0};
-  int32_t resource{0};
+  int64_t resource{0};
   int32_t threadId{0};
   ActivityType activityType;
   std::string activityName;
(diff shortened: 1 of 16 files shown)


sanrise:
@mwootton please take a look, similar to #1398

PR #1418, opened 2026-06-02
Rocprofiler: fix missing grid and block dims in output

mwootton:
Under rocprofiler-sdk, hipModuleLaunchKernel has different argument names than hipLaunchKernel.  Add those so we can get grid and block dims.

The odd _inl.h pattern includes anonymous thread_local storage, a recipe for disaster.  Avoid disaster; heed it's own warning.


mwootton:
Oh, TU problem was actually fixed in https://github.com/pytorch/kineto/pull/1357
I managed to stumble into it before merging.  Never hurts to fix stuff twice.

sanrise:
Copying my internal comment:

worth making this testable? Today this is a lambda assigned to a variable. So its not testable - i.e. it's a local variable, not a function with external linkage.

we could turn it into an ordinary named function, and declare it in the header. As a benefit, `RocmActivityProfilerTest.cpp` can `#include "RocprofLogger.h"` and just call `extract_kernel_args(...)` directly. 

So we could test like:

```
// RocmActivityProfilerTest.cpp
TEST(RocprofLogger, ModuleLaunchGridBlockDims) {
  kernel_args args;
  auto feed = <const char* name, uint32_t val|&> {
    extract_kernel_args(/*kind*/0, /*op*/0, /*argnum*/0,
                        &val, /*ind*/0, /*type*/"uint32",
                        name, /*str*/"", /*deref*/0, &args);
  };
  // hipModuleLaunchKernel arg names
  feed("gridDimX", 2);  feed("gridDimY", 3);  feed("gridDimZ", 4);
  feed("blockDimX", 5); feed("blockDimY", 6); feed("blockDimZ", 7);

  // grid arrives via workgroupSize, block via gridSize (crossed by design)
  EXPECT_EQ(args.workgroupSize.x, 2);  // -> grid
  EXPECT_EQ(args.workgroupSize.y, 3);
  EXPECT_EQ(args.workgroupSize.z, 4);
  EXPECT_EQ(args.gridSize.x, 5);       // -> block
  EXPECT_EQ(args.gridSize.y, 6);
  EXPECT_EQ(args.gridSize.z, 7);
}
```

Otherwise LGTM

scotts:
@sanrise, @mwootton, let's handle that in a follow-up, if we think it's worthwhile.

PR #1421, opened 2026-06-02
Improvements to RocprofLogger

jrmadsen:
- Check return codes of rocprofiler-sdk API calls and log errors
- Avoid using .handle for opaque types directly
- Improve finalization variable lifetime and ordering
  - use placement new into buffer in .bss section to prevent destruction of `GlobalContext` singleton
  - Implement `RocprofLogger::endTracing()` to force invocation of finalization function
- Remove superfluous logging_ and registered_ variables
  - can be handled by queries to rocprofiler-sdk API

### DIFF
--- a/libkineto/src/RocprofLogger.cpp
+++ b/libkineto/src/RocprofLogger.cpp
@@ -9,17 +9,24 @@
 #include "RocprofLogger.h"
 
 #include <rocprofiler-sdk/context.h>
+#include <rocprofiler-sdk/cxx/hash.hpp>
 #include <rocprofiler-sdk/cxx/name_info.hpp>
+#include <rocprofiler-sdk/cxx/operators.hpp>
 #include <rocprofiler-sdk/fwd.h>
 #include <rocprofiler-sdk/marker/api_id.h>
 #include <rocprofiler-sdk/registration.h>
 #include <rocprofiler-sdk/rocprofiler.h>
 
-#include <time.h>
-#include <unistd.h>
 #include <chrono>
 #include <cstring>
+#include <ctime>
 #include <mutex>
+#include <type_traits>
+#include <unistd.h>
+#include <unordered_map>
+#include <vector>
+
+#include <fmt/format.h>
 
 #include "ApproximateClock.h"
 #include "Demangle.h"
@@ -29,6 +36,20 @@
 using namespace libkineto;
 using namespace std::chrono;
 using namespace RocLogger;
+
+#define ROCPROFSDK_CHECK_AND_LOG(LEVEL, RESULT)                                \
+  {                                                                            \
+    rocprofiler_status_t CHECKSTATUS = RESULT;                                 \
+    if (CHECKSTATUS != ROCPROFILER_STATUS_SUCCESS) {                           \
+      LOG(LEVEL) << fmt::format(                                               \
+          "[RocprofLogger] '{}' failed with error code {}: {}", #RESULT,       \
+          static_cast<std::underlying_type_t<rocprofiler_status_t>>(           \
+              CHECKSTATUS),                                                    \
+          rocprofiler_get_status_string(CHECKSTATUS));                         \
+    }                                                                          \
+  }
+
+#define ROCPROFSDK_CHECK(RESULT) ROCPROFSDK_CHECK_AND_LOG(ERROR, RESULT)
 
 namespace {
 
@@ -40,7 +61,8 @@ using kernel_name_map_t =
     std::unordered_map<rocprofiler_kernel_id_t, std::string>;
 using rocprofiler::sdk::buffer_name_info;
 using rocprofiler::sdk::callback_name_info;
-using agent_info_map_t = std::unordered_map<uint64_t, rocprofiler_agent_v0_t>;
+using agent_info_map_t =
+    std::unordered_map<rocprofiler_agent_id_t, rocprofiler_agent_v0_t>;
 
 // extract copy args
 struct copy_args {
@@ -249,8 +271,12 @@ class RocprofApiIdList : public ApiIdList {
   std::unordered_map<std::string, size_t> nameMap_;
 };
 
+constexpr auto null_context_id = rocprofiler_context_id_t{0};
+
 struct GlobalContext {
   rocprofiler_client_id_t* clientId{nullptr};
+  rocprofiler_client_finalize_t finalizer = nullptr;
+
   rocprofiler_tool_configure_result_t cfg = rocprofiler_tool_configure_result_t{
       sizeof(rocprofiler_tool_configure_result_t),
       &RocprofLogger::toolInit,
@@ -258,8 +284,8 @@ struct GlobalContext {
       nullptr};
 
   // Contexts
-  rocprofiler_context_id_t utilityContext = {0};
-  rocprofiler_context_id_t context = {0};
+  rocprofiler_context_id_t utilityContext = null_context_id;
+  rocprofiler_context_id_t context = null_context_id;
 
   // Buffers
   rocprofiler_buffer_id_t buffer = {};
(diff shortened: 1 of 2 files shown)


jrmadsen:
I have signed the CLA.

scotts:
@jrmadsen, thank you! The bot isn't picking up that you signed the CLA. We're also getting a failure on the ROCm PyTorch tests, but it may be an infra issue, so I've kicked it off again. Can you also rebase from main? We've made some testing improvements in the past week that may also help with that.

libkineto/src/RocprofLogger.cpp
```diff
@@ -262,13 +288,16 @@ struct GlobalContext {
...
 };
 
+auto global_context_buffer = std::array<std::byte, sizeof(GlobalContext)>{};
+
 GlobalContext& getGlobalContext() {
-  static GlobalContext instance;
-  return instance;
+  // placement new into .bss section to prevent destruction issues on shutdown.
+  static auto *instance = new (global_context_buffer.data()) GlobalContext{};
+  return *instance;
```
sanrise, line 314: This is a legit, common trick (leaky singleton / immortal singleton).

Reserve a chunk of raw bytes big enough to hold a GlobalContext - the compiler sees plain bytes, not an object, so there's no destructor attached to it. Then construct a GlobalContext in this exact memory we reserved - instead of allocating fresh. Thus we get a real, live object, but since nothing owns it as a typed object with a destructor, it's never destroyed. 

The problem: 
`std::array<std::byte, ...>` only guarantees 1-byte alignment, because a single byte can sit anywhere. So the reserved buffer can start at any address, say one ending in ...3. Then placement-new drops a GlobalContext (and its std::mutex) onto that odd address. Now the mutex is misaligned: it's sitting somewhere the standard says it's never allowed to. That's undefined behavior. In practice it ranges from "works by luck" to torn reads, failed atomics, or crashes, and it's the kind of thing that's fine on one toolchain/arch and explodes on another.

Suggested patch:
```
alignas(GlobalContext) auto global_context_buffer =
    std::array<std::byte, sizeof(GlobalContext)>{};
```

`alignas(GlobalContext)` forces the buffer's starting address to satisfy whatever GlobalContext's strictest member needs (the mutex, etc). Now placement-new lands the object on a legal address and the undefined behavior is gone.

mwootton:
@scotts Are there any other changes @jrmadsen or I need to work on here?  Thanks.

libkineto/src/RocprofLogger.cpp
scotts: @sanrise, I think we should be even simpler:
```
  GlobalContext& getGlobalContext() {
    static GlobalContext* instance = new GlobalContext{};  // intentionally leaked
    return *instance;
  }
```
I'd rather code be obviously correct, and the alignment operator with placement new is not obvious. We also intentionally leak global singletons in other places. If intentionally leaking the global context becomes a problem, we can deal with it then - but I doubt it will.

libkineto/src/RocprofLogger.cpp
```diff
@@ -30,6 +37,20 @@ using namespace libkineto;
...
+  {                                                                            \
+    rocprofiler_status_t CHECKSTATUS = RESULT;                                 \
+    if (CHECKSTATUS != ROCPROFILER_STATUS_SUCCESS) {                           \
+      LOG(LEVEL) << fmt::format(                                               \
+          "[RocprofLogger] '{}' failed with error code {}: {}", #RESULT,       \
+          static_cast<std::underlying_type_t<rocprofiler_status_t>>(           \
+              CHECKSTATUS),                                                    \
+          rocprofiler_get_status_string(CHECKSTATUS));                         \
+    }                                                                          \
+  }
```
scotts: It's safer to wrap this in the conventional `do { ... } while (0)`.

libkineto/src/RocprofLogger.cpp
```diff
@@ -771,40 +806,61 @@ void RocprofLogger::ensureRegistered() {
...
       VLOG(0) << "rocprofiler-sdk tool registration completed successfully";
-      singleton().registered_ = true;
     } else {
       LOG(WARNING) << "rocprofiler_force_configure failed with status "
-                   << result;
+                   << result << ": " << rocprofiler_get_status_string(result);
     }
   } else if (status == 1) {
-    singleton().registered_ = true;
   }
```
scotts, line 845: This is now an empty branch - we should delete it.

libkineto/src/RocprofLogger.cpp
```diff
@@ -331,36 +358,40 @@ int RocprofLogger::toolInit(
   auto& globalContext = getGlobalContext();
   globalContext.name_info = rocprofiler::sdk::get_callback_tracing_names();
   globalContext.buff_name_info = rocprofiler::sdk::get_buffer_tracing_names();
+  globalContext.finalizer = finialize_func;
```
scotts: We have two error paths, but the finalizer is still set. I think that's invalid - we should probably only set the finalizer after we're certain init will succeed.

Also, might as well fix the typo: `finialize_func` -> `finalize_func`

libkineto/src/RocprofLogger.cpp
```diff
@@ -370,8 +401,6 @@ int RocprofLogger::toolInit(
   apiList.add("hipGetDevice");
   apiList.add("hipSetDevice");
   apiList.add("hipGetLastError");
-  apiList.add("__hipPushCallConfiguration");
-  apiList.add("__hipPopCallConfiguration");
```
scotts, line 386: Is removing these purposeful?

scotts:
@mwootton, I added some other minor comments.

mwootton:
I pinged @jrmadsen.  Removing the items from the filter list looks like a mistake, those are pure noise AFAIK.  They wrap each kernel launch api.

libkineto/src/RocprofLogger.cpp
jrmadsen: Yes, these are part of the `HIP_COMPILER_API` and y'all are only tracing the `HIP_RUNTIME_API`.

libkineto/src/RocprofLogger.cpp
jrmadsen: > @sanrise, I think we should be even simpler:
> 
> ```
>   GlobalContext& getGlobalContext() {
>     static GlobalContext* instance = new GlobalContext{};  // intentionally leaked
>     return *instance;
>   }
> ```
> 
> I'd rather code be obviously correct, and the alignment operator with placement new is not obvious. We also intentionally leak global singletons in other places. If intentionally leaking the global context becomes a problem, we can deal with it then - but I doubt it will.

@scotts the placement new approach tends to have better performance for the same reasons that a static global variable tends to have better performance than a heap allocated variable... fixed address at compile time which generally leads to predictability/data locality/cache friendliness. Given that the global context is frequently utilized, this seems like the better approach. Furthermore, the `GlobalContext` object will not show up in a leak sanitizer whereas the leaked pointer will.

jrmadsen:
> I pinged @jrmadsen. Removing the items from the filter list looks like a mistake, those are pure noise AFAIK. They wrap each kernel launch api.

It was not a mistake. Those `__hip*` functions are part of the `HIP_COMPILER_API` and y'all are only tracing the `HIP_RUNTIME_API`.

libkineto/src/RocprofLogger.cpp
jrmadsen: Marking this as resolved but if you really want the `new GlobalContext{}` instead of the placement new @scotts, please lmk.

libkineto/src/RocprofLogger.cpp
```diff
@@ -398,62 +428,83 @@ int RocprofLogger::toolInit(
...
+    ROCPROFSDK_CHECK(rocprofiler_context_is_valid(globalContext.context, &isValid));
     if (isValid == 0) {
-      globalContext.context.handle = 0; // Can't destroy it, so leak it
+      // rocprofiler-sdk will cleanup all client contexts if init does not
+      // return 0
+      globalContext.context = null_context_id;
       return -1;
     }
   }
-  rocprofiler_stop_context(globalContext.context);
```
jrmadsen, line 446: Contexts are not started automatically... there is no need to stop this if it was never started.

scotts:
@jrmadsen, the lint job is failing. Can you fix?

PR #1424, opened 2026-06-05
map kernel and copy events to logical stream ID instead of HSA queue ID (#1424)

hannaxu:
Summary:

Previously, `resourceId()` propagated the HSA queue ID for kernel events, rather than the logical stream ID. The AMD default is `GPU_MAX_HW_QUEUES=4`, which follows why stream interleaving was noticeable only when number of workers (total) >= 5 which is common on models with remote (additional remote workers). Therefore, there was stream interleaving as we were mapping > 5 streams to 4 HSA queues. The separate `backfillAsyncCopyStreams` for H2D/D2D was also propagating HW queue IDs. 

Switch to int64_t for stream ids to avoid truncation; HW queue IDs fit fine in int32. Return streamId with queueId fallback in`resourceId()`. Recover the real HIP stream from the correlated runtime row, resolving both kernels and copies to stream ID via `backfillAsyncStreams`. Remove separate `backfillAsyncCopyStreams`. Async rows carry only the HW queue, so we map correlation --> HIP stream from the runtime rows.

*Process*
1. Infer stream from queue when a queue maps unambiguously to one stream
2. Remap raw 64-bit HIP stream pointers to per-device indices since Perfetto misrenders large id values onto the same track
3. Kernels and copies share one per-device index space, so ops on the same HIP stream land on the same track. `resourceId()` returns the resolved streamId with a queueId <--> unambiguous stream map fallback

Add the HSA queue ID as a field in the trace, but not as the track provenance in next diff D107568026

Differential Revision: D107306915

### DIFF
--- a/libkineto/include/GenericTraceActivity.h
+++ b/libkineto/include/GenericTraceActivity.h
@@ -142,7 +142,7 @@ class GenericTraceActivity : public ITraceActivity {
   int64_t endTime{0};
   int32_t id{0};
   int32_t device{0};
-  int32_t resource{0};
+  int64_t resource{0};
   int32_t threadId{0};
   ActivityType activityType;
   std::string activityName;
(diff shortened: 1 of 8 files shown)


sanrise (approved):
LGTM, thanks for adding this -

PR #1425, opened 2026-06-05
surface HSA queue ID (#1425)

hannaxu:
Summary:

Having the HSA queue ID is still important information. Surface it as a field to be easily queried across, rather than UI track provenance.

Reviewed By: sanrise

Differential Revision: D107568026


sanrise (approved):
Parent diff remaps stream for display but leaves queue untouched, so this surfaces the raw HSA queue while keeping logical stream placement intact.

PR #1428, opened 2026-06-09
Speedup and fix PyTorch CI build

scotts:
**Problem**
Kineto's PyTorch-build CI jobs timeout. Two phases dominate the run:

1. Installing cmake takes ~62 minutes because setup.sh adds the conda-forge channel with strict priority, which forces conda's classic solver to re-solve the entire base environment onto conda-forge.
2. The PyTorch build then recompiles all ~8k objects from scratch on every run because no compiler cache is configured.

Addressing that also exposed more problems:

1. Infinite directory creation.
2. New actual failing tests.

**Fixes**
1. Install cmake from PyPI instead of conda. The cmake wheel ships a prebuilt binary and installs in seconds, and this removes conda from the critical path for all nine workflows that source setup.sh.
3. Enable sccache backed by PyTorch's S3 cache for the PyTorch build. Most Kineto PRs touch only Kineto, so the bulk of PyTorch's objects become cache hits on warm runs.
4. We also eliminate an infinite directory problem where we symlinked Kineto in the cloned PyTorch, but we cloned PyTorch in the Kineto directory.
5. Manually deselect the failing tests. We're created https://github.com/pytorch/kineto/issues/1430 and https://github.com/pytorch/kineto/issues/1429 to track.

### DIFF
--- a/.github/scripts/config_rocm.sh
+++ b/.github/scripts/config_rocm.sh
@@ -46,11 +46,8 @@ KINETO_USE_SCCACHE=0
 # shellcheck disable=SC2034
 DESELECTED_TESTS=(
   test/profiler/test_profiler.py::TestExperimentalUtils::test_fuzz_symbolize
-  # The device-generic TestProfilerDevice suite (pytorch/pytorch#182434) does
-  # not hold on GPU runners: its CUDA instances error during setup (on ROCm/HIP
-  # one also asserts a CUDA-specific event count), and forked_process deadlocks
-  # once a GPU runtime is initialized. A class node ID deselects all its methods
-  # because pytest matches --deselect by node-ID prefix. Remove when fixed upstream.
+
+  # https://github.com/pytorch/kineto/issues/1429
   test/profiler/test_profiler.py::TestProfilerDeviceCUDA
   test/profiler/test_profiler.py::TestProfilerDeviceCPU::test_forked_process_cpu
 )
--- a/.github/scripts/config_rocm.sh
+++ b/.github/scripts/config_rocm.sh
@@ -46,4 +46,11 @@ KINETO_USE_SCCACHE=0
 # shellcheck disable=SC2034
 DESELECTED_TESTS=(
   test/profiler/test_profiler.py::TestExperimentalUtils::test_fuzz_symbolize
+  # The device-generic TestProfilerDevice suite (pytorch/pytorch#182434) does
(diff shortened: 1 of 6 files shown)


.github/scripts/config_rocm.sh
```diff
@@ -40,4 +46,8 @@ export PYTORCH_TEST_WITH_ROCM=1
 # shellcheck disable=SC2034
 DESELECTED_TESTS=(
   test/profiler/test_profiler.py::TestExperimentalUtils::test_fuzz_symbolize
+
+  # https://github.com/pytorch/kineto/issues/1429
+  test/profiler/test_profiler.py::TestProfilerDeviceCUDA
```
ryanzhang22, line 51: weird... seems like this should've been automatically deselected for rocm tests

.github/scripts/pytorch_build_test.sh
```diff
@@ -14,11 +14,15 @@ SCRIPTS_DIR="$(cd "$(dirname "$0")" && pwd)"
 KINETO_DIR=$(pwd)
 echo "====: Kineto directory: ${KINETO_DIR}"
 
-# Clone PyTorch and replace its Kineto with PR version
-git clone --recursive --branch viable/strict https://github.com/pytorch/pytorch.git
+# Clone PyTorch as a sibling of the Kineto checkout, never inside it. Below we
+# replace PyTorch's third_party/kineto with a symlink back to KINETO_DIR. If the
+# PyTorch tree lived inside KINETO_DIR, that symlink would point at a directory
```
ryanzhang22, line 19: small nit: missing "If"

.github/scripts/config_rocm.sh
scotts: I believe these get HIPified, so the CUDA tests are also ROCm tests.

ryanzhang22 (approved):

.github/scripts/pytorch_build_test.sh
scotts: Hmm. I'm not sure what you mean?

ryanzhang22: "If" PyTorch tree lived inside KINETO_DIR, that symlink would point at a directory...

scotts: It's there, on the line above. :)

PR #1441, opened 2026-06-22
[xpupti] Enable OVERHEAD activity

aostrowski-hbn:
The xpupti plugin already wires PTI_VIEW_COLLECTION_OVERHEAD through handleOverheadActivity(), but two issues remained:

- The overhead kind name came straight from PTI's ptiViewOverheadKindToString(), whose spelling does not match the vocabulary the CUDA/CUPTI backend uses. Add a local overheadKindString() that maps PTI kinds to CUPTI spelling (Resource, Buffer Flush, Driver, Unknown) so XPU and CUDA traces are consistent. PTI_VIEW_OVERHEAD_KIND_TIME (the only kind PTI currently emits) aggregates the cost of the Level Zero calls PTI injects to instrument the workload, which matches CUPTI's "Instrumentation" overhead.

- "overhead occupancy" divided by duration() without guarding against a zero-length span; PTI can emit point-like records (start == end). Compute the percentage only when there is a non-zero overhead duration.

Add a unit test (XpuptiProfilerTest.OverheadActivity) that profiles a tiny GEMM with the OVERHEAD activity enabled and asserts the shape of the emitted records (host-side deviceId == -1, CUPTI-style name, overhead cost/count/occupancy metadata). Overhead records are folded into RunProfilerTest's activity-count check via acceptOverheadActivities(), mirroring the existing driver-activity handling, and the run is expected to produce at least one.

### DIFF
--- a/libkineto/src/plugin/xpupti/XpuptiActivityHandlers.cpp
+++ b/libkineto/src/plugin/xpupti/XpuptiActivityHandlers.cpp
@@ -306,9 +306,9 @@ namespace {
 // overheadKindString() (see cupti_strings.cpp), so XPU and CUDA traces use the
 // same vocabulary. PTI's own ptiViewOverheadKindToString() returns raw enum
 // spellings (e.g. "BUFFER_FLUSH", "BUFFER_TIME") which do not match CUDA.
-// PTI_VIEW_OVERHEAD_KIND_TIME aggregates the time PTI spends inside the Level
-// Zero calls it injects to instrument the workload, which is the same notion as
-// CUPTI's "Instrumentation" overhead.
+// PTI_VIEW_OVERHEAD_KIND_TIME aggregates the time PTI spends inside
+// the Level Zero calls it injects to instrument the workload,
+// which is the same notion as CUPTI's "Instrumentation" overhead.
 const char* overheadKindString(pti_view_overhead_kind kind) {
   switch (kind) {
     case PTI_VIEW_OVERHEAD_KIND_UNKNOWN:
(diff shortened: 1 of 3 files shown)


libkineto/src/plugin/xpupti/XpuptiActivityHandlers.cpp
```diff
@@ -301,13 +301,39 @@ void XpuptiActivityProfilerSession::handleRuntimeKernelMemcpyMemsetActivities(
...
 }
 
+namespace {
+// Map a PTI overhead kind to a human-readable name aligned with CUPTI's
+// overheadKindString() (see cupti_strings.cpp), so XPU and CUDA traces use the
+// same vocabulary. PTI's own ptiViewOverheadKindToString() returns raw enum
+// spellings (e.g. "BUFFER_FLUSH", "BUFFER_TIME") which do not match CUDA.
+// PTI_VIEW_OVERHEAD_KIND_TIME aggregates the time PTI spends inside the Level
+// Zero calls it injects to instrument the workload, which is the same notion as
+// CUPTI's "Instrumentation" overhead.
```
tsocha: Don't break the Level Zero
```suggestion
// PTI_VIEW_OVERHEAD_KIND_TIME aggregates the time PTI spends inside
// the Level Zero calls it injects to instrument the workload,
// which is the same notion as CUPTI's "Instrumentation" overhead.
```

libkineto/src/plugin/xpupti/XpuptiActivityHandlers.cpp
```diff
@@ -316,11 +342,14 @@ void XpuptiActivityProfilerSession::handleOverheadActivity(
   overhead_activity->threadId = activity->_overhead_thread_id;
   overhead_activity->addMetadata(
       "overhead cost", activity->_overhead_duration_ns);
+  // Guard against division by zero as PTI may emit point-like
+  // overhead records (start == end).
+  const auto occupancy = activity->_overhead_duration_ns > 0
+      ? activity->_overhead_duration_ns * 100 / overhead_activity->duration()
+      : 0;
```
tsocha: This comment is weird. Why do you compare a value which is not used as a divider?
I would expect that you will check if `overhead_activity->duration() > 0`

tsocha (changes requested):

tsocha (approved):

ryanzhang22 (approved):

PR #1459, opened 2026-06-30
Fix: JSON-escape double quotes in chrome-trace event names

danielsnider-aws:
# Fix: JSON-escape double quotes in chrome-trace event names

Fixes pytorch/pytorch#146900

## Summary

`ChromeTraceLogger` writes each event `name` as a bare JSON string value
(`"name": "<value>"`) but does not escape embedded double quotes. When an event
name contains a `"`, the JSON string terminates early and the **entire exported
`trace.json` becomes invalid** — `json.load()` raises
`JSONDecodeError: Expecting ',' delimiter`, and any downstream consumer
(TensorBoard, Perfetto, Chrome `chrome://tracing`, custom analyzers) fails to
load the whole trace, not just the offending event.

## How it happens

With `with_stack=True`, the profiler builds each Python-frame event name as
`"<co_filename>(<lineno>): <funcname>"`. The `co_filename` is attacker-free but
not quote-free:

- **`torch.compile` / dynamo** compile generated code whose `co_filename` is a
  dict-like string, e.g. `nn.Linear.__init__`'s
  `factory_kwargs = {"device": device, "dtype": dtype}` surfaces a "filename"
  like `{"device": "cpu", "dtype": "float32"}`.
- More generally, any frame whose source text (e.g. a `<module>` frame's line)
  contains a `"` — including ordinary module docstrings — produces the same
  corruption.

The result in the trace file:

```
"name": "{"device": "cpu", "dtype": "float32"}(12): run"
```

The unescaped `"` after `{` ends the JSON string, and the document no longer
parses.

This is the same class of bug as the previously-fixed #98432, which only
escaped backslashes (Windows paths) — quotes still slipped through.

## The fix

Add a small `escapeQuotesForJSON()` helper and apply it to the event name in
`ChromeTraceLogger::handleActivity()`, **after** the existing
`sanitizeStrForJSON()` / `sanitizeForNonReadableChars()` steps so the inserted
escape characters are not rewritten by the `\` → `/` substitution.

The helper is deliberately scoped to **bare string values** (the event name).
It is **not** applied to pre-formatted JSON fragments such as
`ITraceActivity::metadataJson()`, whose quotes are structural — escaping those
would corrupt valid metadata.

### Why quote-only escaping is sufficient here

By the time `escapeQuotesForJSON()` runs, the name has already been through:
- `sanitizeStrForJSON()` — replaces every `\` with `/` and strips `\n`, so no
  backslashes remain to double-escape, and
- `sanitizeForNonReadableChars()` — replaces the whole string with `"unknown"`
  if it contains any non-printable byte, so control characters cannot survive.

After that chain, `"` is the only JSON-special character that can remain, so
escaping it is enough to guarantee a valid JSON string value.

## Test

Adds `OutputJsonTest` (CPU-only):
- `EventNameWithQuotesProducesValidJson` — an event name containing quotes now
  produces valid JSON and the name round-trips exactly.
- `PlainEventNameIsUnchanged` — a normal name (e.g. `aten::addmm`) is untouched.

## Reproducer (from the issue)

```python
import json, torch
from torch.profiler import profile, ProfilerActivity
torch.set_num_threads(1)
g = {}
exec(compile("def run(a, b): return a @ b", '{"device": "cpu"}', "exec"), g)
x = torch.randn(8, 8)
with profile(activities=[ProfilerActivity.CPU], with_stack=True) as p:
    g["run"](x, x)
p.export_chrome_trace("/tmp/trace.json")
json.load(open("/tmp/trace.json"))   # before: JSONDecodeError; after: OK
```

Before this change the script raises `JSONDecodeError`; after it, the trace
loads and the quoted event name round-trips correctly.

### DIFF
--- a/libkineto/test/CMakeLists.txt
+++ b/libkineto/test/CMakeLists.txt
@@ -153,3 +153,16 @@ target_include_directories(GenericTraceActivityMetadataTest PRIVATE
     "${LIBKINETO_DIR}/include"
     "${LIBKINETO_DIR}/src")
 gtest_discover_tests(GenericTraceActivityMetadataTest)
+
+# OutputJsonTest
+add_executable(OutputJsonTest OutputJsonTest.cpp)
+target_link_libraries(OutputJsonTest PRIVATE
+    gtest_main
+    kineto_base kineto_api
+    nlohmann_json::nlohmann_json
+    ${XPU_XPUPTI_LIBRARY})
+target_include_directories(OutputJsonTest PRIVATE
+    "${LIBKINETO_DIR}"
+    "${LIBKINETO_DIR}/include"
+    "${LIBKINETO_DIR}/src")
+gtest_discover_tests(OutputJsonTest)
--- a/libkineto/test/OutputJsonTest.cpp
+++ b/libkineto/test/OutputJsonTest.cpp
@@ -1,0 +1,85 @@
+/*
+ * Copyright (c) Meta Platforms, Inc. and affiliates.
+ * All rights reserved.
+ *
+ * This source code is licensed under the BSD-style license found in the
+ * LICENSE file in the root directory of this source tree.
+ */
+
+#include <gtest/gtest.h>
+#include <nlohmann/json.hpp>
+
+#include <cstdio>
+#include <fstream>
+#include <iterator>
+#include <string>
+
+#include "include/GenericTraceActivity.h"
+#include "include/TraceSpan.h"
+#include "src/output_json.h"
+
+using namespace KINETO_NAMESPACE;
+using namespace libkineto;
+
+namespace {
+
+// Exposes the protected finalizeTrace(endTime) overload so the test can flush a
+// trace without constructing a Config / ActivityBuffers.
+class TestableChromeTraceLogger : public ChromeTraceLogger {
+ public:
+  explicit TestableChromeTraceLogger(const std::string& file)
+      : ChromeTraceLogger(file) {}
+  using ChromeTraceLogger::finalizeTrace;
+};
+
+std::string readFile(const std::string& path) {
+  std::ifstream f(path);
+  return std::string(
+      (std::istreambuf_iterator<char>(f)), std::istreambuf_iterator<char>());
+}
+
+// Write a trace containing a single CPU-op event with the given name, parse it
+// back, and assert the trace is valid JSON and the name round-trips exactly.
+void expectEventNameRoundTrips(const std::string& eventName) {
+  const std::string filename = ::testing::TempDir() + "kineto_output_json.json";
+
+  TraceSpan span(0, 0, "test_span");
+  GenericTraceActivity act(span, ActivityType::CPU_OP, eventName);
+  act.startTime = 100;
+  act.endTime = 200;
+  act.device = 0;
+  act.resource = 0;
+
+  TestableChromeTraceLogger logger(filename);
+  logger.handleTraceStart({}, "");
+  logger.handleGenericActivity(act);
+  logger.finalizeTrace(/*endTime=*/300);
+
+  nlohmann::json data;
+  ASSERT_NO_THROW(data = nlohmann::json::parse(readFile(filename)));
+
+  bool found = false;
+  for (const auto& event : data["traceEvents"]) {
+    if (event.contains("ph") && event["ph"] == "X" && event.contains("name") &&
+        event["name"].get<std::string>() == eventName) {
+      found = true;
+      break;
+    }
+  }
+  EXPECT_TRUE(found) << "event name did not round-trip: " << eventName;
+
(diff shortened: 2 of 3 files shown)


danielsnider-aws:
Rebased onto latest `main` and resolved the conflicts — now mergeable. The conflict was with #1453 (`Escape LOG messages in output_json.cpp`): that added `sanitizeLogStrForJSON` (for free-form log strings, replaces `"`→`'`), which is complementary to this PR's `escapeQuotesForJSON` (for the bare event `name`, backslash-escapes `"` losslessly so the name round-trips). Both helpers are kept; the merged `OutputJsonTest.cpp` now carries both #1453's log-string tests and this PR's event-name tests.

@ryanzhang22 — since you own the recent `output_json.cpp` escaping work (#1453) and the surrounding area, would you be able to review (or route this internally)? cc @scotts (you did the JSON chrome-trace writer refactors #1342/#1397).

Context: this fixes pytorch/pytorch#146900 — with `with_stack=True`, an event `name` like a dynamo co_filename `{"device": "cpu"}` (or any source line containing `"`) makes the whole exported `trace.json` invalid JSON, so downstream consumers fail to load the entire trace. The fix JSON-escapes the name, scoped to the bare-string value only (not pre-formatted metadata fragments).

libkineto/test/OutputJsonTest.cpp
```diff
@@ -46,3 +87,58 @@ TEST(OutputJsonTest, DropsControlCharsInLogStrings) {
   const std::string out = traceWithWarning("abc\tdef\rghi\n");
   EXPECT_NE(out.find("abcdefghi"), std::string::npos);
 }
+
+// Reproduces pytorch/pytorch#146900: with with_stack=True an event name can
+// contain `"` (e.g. torch.compile/dynamo-generated co_filenames such as
+// `{"device": "cpu"}`). The chrome-trace writer must JSON-escape those quotes
+// so the emitted trace remains valid JSON and the name round-trips intact.
+TEST(OutputJsonTest, EventNameWithQuotesProducesValidJson) {
```
ryanzhang22: nit: these two tests duplicate logic + round trip validation, let's pull it out into a helper with a readable name

ryanzhang22:
looks good overall, one comment + can we clean up/trim down some of the code comments, there's some duplicate blocks

scotts:
@danielsnider-aws, you'll have to resolve a merge conflict because #1460 just deleted that test file - it was previously only testing that *logging* didn't mess up the JSON, and we just stopped writing the logs to JSON. Let's make sure to add your new tests, but not reinstate the deleted tests.

danielsnider-aws:
Thanks @ryanzhang22 and @scotts — done:

- Rebased onto latest `main`. Since #1460 removed the log-to-JSON path (and `OutputJsonTest.cpp`), I did **not** reinstate `sanitizeLogStrForJSON` or the log-string tests — the file now has only the two event-name tests. (Also updated the test's `finalizeTrace(endTime)` call to match #1460's new signature.)
- Pulled the shared round-trip/validation logic into an `expectEventNameRoundTrips()` helper.
- Trimmed the duplicated code comments down.

Built + ran locally against latest main (gcc 11 / C++20): `OutputJsonTest` — 2 passed. Now mergeable.

libkineto/test/CMakeLists.txt
```diff
@@ -134,6 +134,7 @@ add_executable(TypedMetadataTest TypedMetadataTest.cpp)
 target_link_libraries(TypedMetadataTest PRIVATE
     gtest_main
     kineto_base kineto_api
+    nlohmann_json::nlohmann_json
```
ryanzhang22: did this need to be added? I don't see `TypedMetadataTest` being modified in this PR.

danielsnider-aws: Fixed. Removed. This was accidentally added during conflict resolution.

@ryanzhang22, how does this PR look now? 

Thank you.

danielsnider-aws:
Hi @ryanzhang22, 

It looks like the internal build passed but the linter failed. Let me know if there's anything I can do to help.

Thanks,

ryanzhang22:
@danielsnider-aws no worries, I will land this and take care of the lint internally

danielsnider-aws:
@ryanzhang22, thank you!

danielsnider-aws:
Now that this seems to be merged, as per:
> @[pytorchmergebot](https://github.com/pytorchmergebot) pytorchmergebot added the [Merged](https://github.com/pytorch/pytorch/issues?q=state%3Aopen%20label%3AMerged) label [3 hours ago](https://github.com/pytorch/pytorch/pull/189542#event-27824566350)

@ryanzhang22, how can I find out or track when this change is released publicly i.e. released in as a wheel in a repo like PyPI? 

Thanks.

PR #1467, opened 2026-07-06
[xpupti] Fix CPU->GPU flow links via activity type, not an API-name allowlist

aostrowski-hbn:
The XPU profiler marked a host record as the start of a CPU→GPU ("ac2g") flow arrow by exact-matching its API name against a hardcoded allowlist (`correlateRuntimeOps_`). That list did not contain `urEnqueueKernelLaunchWithArgsExp` — the launch API used by the 2025.3+ SYCL runtime — so kernel launches stopped emitting a flow start (`ph:"s"`) and the host→device arrow disappeared from the trace.

## Approach
Rather than add the new name to the list (which needs a new entry on every UR API rename/addition), decide the flow start from the record’s **activity type**: only host runtime (`XPU_RUNTIME`) records start a flow.

- The runtime view is already filtered to work-submitting APIs via `ptiViewEnableRuntimeApiClass(PTI_API_CLASS_GPU_OPERATION_CORE)`.
- Driver (`XPU_DRIVER`) records share the runtime record’s correlation id and must **not** also start a flow, or the trace gets a duplicate flow start.

This removes the name allowlist entirely (robust to future UR renames) and extracts the check into a static `startsFlow()` for unit testing. Net −35 lines.

> Note: this supersedes the original allowlist patch on this branch, following @mschilli’s suggestion to key off `GPU_OPERATION_CORE` / activity type instead of maintaining a name list. The `XPU_DRIVER`-exclusion guard is what prevents the duplicate-flow-start collision that a bare `flow.start = 1` would cause.

## No PTI version gate needed
`ptiViewEnableRuntimeApiClass` / `PTI_API_CLASS_GPU_OPERATION_CORE` exist since PTI 0.12.0 and are already called unconditionally on `main`; this change references no PTI symbol (only the Kineto `ActivityType` enum), so it builds on all currently-supported PTI versions (0.15–0.18 verified reasoning; 0.18 built).

## Test Plan
`XpuptiFlowCorrelationTest` now asserts runtime records start a flow while driver and device records do not (hardware-free, in the CTest set).

Verified on Intel Data Center GPU Max (PVC), oneAPI 2026.0 / PTI 0.18:
- `XpuptiFlowCorrelationTest` passes; full xpupti `ctest` **10/10** passes.
- Rebuilt PyTorch against the patched kineto and ran a mixed workload (H2D copy, add/relu/matmul/reduction, D2H copy): every kernel and memcpy gets **exactly one** host→device arrow, with **zero** Perfetto flow errors (no `flow_duplicate_id`) and no PyTorch duplicate-flow-start warnings.

### DIFF
--- a/libkineto/src/plugin/xpupti/XpuptiActivityHandlers.cpp
+++ b/libkineto/src/plugin/xpupti/XpuptiActivityHandlers.cpp
@@ -214,8 +214,7 @@ void XpuptiActivityProfilerSession::handleRuntimeKernelMemcpyMemsetActivities(
   if constexpr (handleRuntimeActivities) {
     trace_activity->device = activity->_process_id;
     trace_activity->resource = activity->_thread_id;
-    trace_activity->flow.start =
-        (correlateRuntimeOps_.count(trace_activity->name()) > 0);
+    trace_activity->flow.start = startsFlow(activityType);
   } else {
     trace_activity->device = getDeviceIdxFromUUID(activity->_device_uuid);
     trace_activity->resource = activity->_sycl_queue_id;
--- a/libkineto/src/plugin/xpupti/XpuptiActivityProfilerSession.cpp
+++ b/libkineto/src/plugin/xpupti/XpuptiActivityProfilerSession.cpp
@@ -19,28 +19,17 @@
 
 namespace KINETO_NAMESPACE {
 
-using namespace std::literals::string_view_literals;
-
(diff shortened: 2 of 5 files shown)


aostrowski-hbn:
@ryanzhang22 please import and review

gujinghui (approved):

gujinghui:
@ryanzhang22 @divyanshk @scotts This PR is ready for your review. Thanks.

ryanzhang22 (approved):

PR #1479, opened 2026-07-10
[xpupti] Support more than 1 device in ptiMetricsScopeConfigure

aostrowski-hbn:
Enables the XPU Scope Profiler to collect hardware metrics on more than one XPU. Previously `enableScopeProfiler` hardcoded `device_count=1` in the `ptiMetricsScopeConfigure` call, so metrics were only ever collected on a single card even on multi-GPU hosts.

## What changed
- **Default (no device list) → PTI auto-detect** (`devices=nullptr, count=0`): profiles all available devices.
- **New Kineto config key `XPUPTI_PROFILER_DEVICES=0,2,3`** selects an explicit device subset via a new `selectDeviceHandles` helper (C++20 `std::span` + `std::ranges` gather with bounds validation).
- **Version-guarded** by `PTI_VERSION_AT_LEAST(0, 18)`: multi-device support in `ptiMetricsScopeConfigure` landed in PTI 0.18. On older PTI, single-device behavior is preserved and a request for >1 device is rejected with a clear error.

Device attribution needs no change: each scope record is already correlated to its kernel activity (and thus device/resource) via `_kernel_id`.

## Testing
Verified on a 16-tile Intel Data Center GPU Max (PVC) host built against PTI 0.18:
- Host-only gtests (`XpuptiScopeProfilerConfigTest`): 8/8 pass — config-key parsing + `selectDeviceHandles` bounds validation.
- Real-HW gtests (`XpuptiScopeProfilerTest`): 3/3 pass, including a new `PerKernelScopeExplicitDevice0` exercising the explicit-subset path.
- Device-count evidence from PTI's own logs: **auto-detect initializes 16 per-device metrics handlers** vs **1** for an explicit single device — confirming multi-device configure actually engages all cards.

### DIFF
--- a/libkineto/src/plugin/xpupti/XpuptiScopeProfilerApi.cpp
+++ b/libkineto/src/plugin/xpupti/XpuptiScopeProfilerApi.cpp
@@ -6,14 +6,39 @@
  * LICENSE file in the root directory of this source tree.
  */
 
+#include <fmt/core.h>
+
 #include <algorithm>
 #include <iterator>
+#include <span>
 #include <stdexcept>
+#include <utility>
 
 #include "XpuptiScopeProfilerApi.h"
 #include "XpuptiScopeProfilerConfig.h"
 
 namespace KINETO_NAMESPACE {
+
+std::vector<pti_device_handle_t> selectDeviceHandles(
+    std::span<const pti_device_handle_t> handles,
+    std::span<const int> indices) {
+  const auto outOfRange = [handles](int idx) {
+    return idx < 0 || std::cmp_greater_equal(idx, handles.size());
+  };
+  if (const auto bad = std::ranges::find_if(indices, outOfRange);
+      bad != indices.end()) {
+    throw std::runtime_error(fmt::format(
+        "XPUPTI_PROFILER_DEVICES index {} is out of range; {} XPU device(s) available",
+        *bad,
+        handles.size()));
+  }
+  // Gather: map each requested index to its device handle, preserving order.
+  std::vector<pti_device_handle_t> selected(indices.size());
+  std::ranges::transform(indices, selected.begin(), [handles](int idx) {
+    return handles[static_cast<std::size_t>(idx)];
+  });
+  return selected;
+}
 
 XpuptiScopeProfilerApi::safe_pti_scope_collection_handle_t::
     safe_pti_scope_collection_handle_t(std::exception_ptr& exceptFromDestructor)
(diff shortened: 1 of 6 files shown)


libkineto/src/plugin/xpupti/XpuptiScopeProfilerApi.cpp
```diff
@@ -67,13 +92,53 @@ void XpuptiScopeProfilerApi::enableScopeProfiler(const Config& cfg) {
   }
 
   scopeHandleOpt_.emplace(exceptFromScopeHandleDestructor_);
+
+  const auto& requestedDevices = spcfg.xpuptiProfilerDevices();
+
+#if PTI_VERSION_AT_LEAST(0, 18)
+  if (requestedDevices.empty()) {
+    // Default: profile every available device (PTI auto-detect mode).
```
Rogersyp: This comment was inaccurate. When requested devices is 0, it will profile whichever devices the workload actually uses.  

Note: you may also want to update the PR body.

libkineto/src/plugin/xpupti/XpuptiScopeProfilerConfig.cpp
```diff
@@ -53,10 +65,14 @@ void XpuptiScopeProfilerConfig::printActivityProfilerConfig(
...
         "Xpupti Profiler measure per kernel : {}\n"
-        "Xpupti Profiler max scopes : {}\n",
+        "Xpupti Profiler max scopes : {}\n"
+        "Xpupti Profiler devices : {}\n",
         fmt::join(activitiesXpuptiMetrics_, ", "),
         xpuptiProfilerPerKernel_,
-        xpuptiProfilerMaxScopes_);
+        xpuptiProfilerMaxScopes_,
+        xpuptiProfilerDevices_.empty()
+            ? std::string("all")
```
Rogersyp: "auto" may be more accurate than "all" here - For example, on a host where the workload only touches 4 of 16 cards, only 4 get profiled, not "all."

Rogersyp:
One comment to the Testing description: "Auto-detect initializes 16 per-device metrics handlers" is true because the 16-tile PVC test workload dispatched to all 16 tiles — not because auto-detect unconditionally profiles all devices.

Rogersyp:
Nit: The PTI implementation validates uniformity for any multi-device configure:

```
if (candidate_devices.size() > 1) {
  result = ValidateDevicesUniform(candidate_devices, all_devices);
  ...
}
```
So on a mixed-GPU host, both explicit multi-device selection and auto-detect (if the workload touches >1 non-uniform device) will fail inside `ptiMetricsScopeConfigure` — surfaced through XPUPTI_CALL as a runtime error. The kineto side doesn't (and probably shouldn't) pre-validate this, but a one-line note in the comment or config-key docs ("all selected devices must be the same model") would save users a confusing failure.

libkineto/test/xpupti/XpuptiScopeProfilerConfigTest.cpp
```diff
@@ -68,3 +71,66 @@ TEST_F(XpuptiScopeProfilerConfigTest, ScopesDefaults) {
...
+  // No XPUPTI_PROFILER_DEVICES set -> empty means "all devices".
+  const KN::XpuptiScopeProfilerConfig& xpupti_cfg =
+      KN::XpuptiScopeProfilerConfig::get(cfg);
+  EXPECT_TRUE(xpupti_cfg.xpuptiProfilerDevices().empty());
+}
+
+TEST_F(XpuptiScopeProfilerConfigTest, DevicesParsedList) {
+  KN::Config cfg;
+  EXPECT_TRUE(cfg.parse("XPUPTI_PROFILER_METRICS = metric1"));
+  EXPECT_TRUE(cfg.parse("XPUPTI_PROFILER_DEVICES = 0, 2, 3"));
```
Rogersyp, line 87: nit: would be nice to add a test for testing empty "tokens", for example: 
`EXPECT_TRUE(cfg.parse("XPUPTI_PROFILER_DEVICES = 0, ,2,"));`

Trailing comma + doubled comma: empty tokens must be skipped, not parsed as index 0 or rejected as invalid integers.

Additionally, what if there are duplicated indices: XPUPTI_PROFILER_DEVICES=0,0,2?

aostrowski-hbn:
@Rogersyp solved your comments, can you take another look?

Rogersyp (approved):
LGTM! Thank you for addressing my comments.

aostrowski-hbn:
@EikanWang, @jgong5, @gujinghui please review

gujinghui (approved):

aostrowski-hbn:

ryanzhang22 (approved):

PR #1491, opened 2026-07-23
Add XPU PTI version to Metadata

PawelSwider2000:
Add the Intel PTI version to XPU Kineto trace metadata as xpupti_version.

The XPU profiler session now obtains the runtime version through `ptiVersionString()` and emits it as a JSON string value. This lets exported traces identify the PTI version that produced them, aiding diagnosis of profiler behavior across Intel PTI versions.

Add a focused XPU PTI unit test confirming that the metadata field is present and non-empty.

### DIFF
--- a/libkineto/src/plugin/xpupti/XpuptiActivityProfilerSession.cpp
+++ b/libkineto/src/plugin/xpupti/XpuptiActivityProfilerSession.cpp
@@ -11,6 +11,7 @@
 
 #include "time_since_epoch.h"
 
+#include <pti/pti_version.h>
 #include <sycl/sycl.hpp>
 
 #include <algorithm>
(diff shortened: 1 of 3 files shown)


PawelSwider2000:
@ryanzhang22 Could you please take a look, I added `xpupti_version` in a bit different way than CUDA adds `cuptiVersion`:

`addVersionMetadata("cupti_version", std::to_string(cuptiVersion));`

However the XPU is a plugin and does not inherit from `GenericActivityProfiler`. Please suggest if to unify the way off adding version metadata for traces as unifying could make bigger changes in API. Thanks

ryanzhang22 (approved):

ryanzhang22:
I think this seems fine. In the `GenericActivityProfiler` path `addVersionMetadata` eventually just flows to `addMetadata` anyways.

PR #1507, opened 2026-07-29
Add CPU-only end-to-end test for the on-demand async trace chain

scotts:
Adds a deterministic, end-to-end test which exercises the async path.

### DIFF
--- a/libkineto/src/AsyncActivityProfilerHandler.h
+++ b/libkineto/src/AsyncActivityProfilerHandler.h
@@ -11,6 +11,8 @@
 #include <array>
 #include <atomic>
 #include <chrono>
+#include <condition_variable>
+#include <cstdint>
 #include <memory>
 #include <mutex>
 #include <thread>
@@ -74,6 +76,23 @@ class AsyncActivityProfilerHandler {
 
   void ensureCollectTraceDone();
 
+  // Test-only. Returns how many async traces this handler has finalized. Each
+  // increment means a trace was fully written to its logger and the run loop
+  // returned to idle, so the trace file is on disk. A test that drives the real
+  // background loop can capture a baseline, request a trace, and wait for this
+  // count to advance to observe a completed trace deterministically. Loaded
+  // with acquire ordering so observing an advanced count also makes the
+  // finalized trace's writes visible.
+  [[nodiscard]] uint64_t completedTraceCountForTesting() const {
+    return completedTraceCount_.load(std::memory_order_acquire);
+  }
+
+  // Test-only. Blocks until the finalized-trace count reaches target, or the
+  // timeout elapses; returns true if the count was reached.
+  bool waitForCompletedTraceCountForTesting(
+      uint64_t target,
+      std::chrono::milliseconds timeout);
+
  private:
   bool shouldActivateIterationConfig(int64_t currentIter);
   bool shouldActivateTimestampConfig(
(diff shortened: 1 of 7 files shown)


libkineto/test/AsyncE2ECpuTraceTest.cpp
```diff
@@ -0,0 +1,178 @@
...
+
+  // Start "now": setSkipStartTimeForTesting makes the profiler start on
+  // the first background-loop tick regardless of wall-clock alignment, so there
+  // is no start/warmup window to miss.
+  const int64_t startMs =
+      duration_cast<milliseconds>(system_clock::now().time_since_epoch())
+          .count();
+  const std::string onDemandConfig = fmt::format(
+      "REQUEST_TRACE_ID={}\n"
+      "PROFILE_START_TIME={}\n"
```
ryanzhang22: I see that we added a bypass for `canStart`, but the async handler still checks the config profiling window to see when to stop tracing, and if the end time has passed it'll stop immediately. Do you know if we're able to start profiling within that period?

libkineto/src/AsyncActivityProfilerHandler.h
```diff
@@ -74,6 +76,23 @@ class AsyncActivityProfilerHandler {
 
   void ensureCollectTraceDone();
 
+  // Test-only. Returns how many async traces this handler has finalized. Each
```
ryanzhang22: In the past I have been iffy on putting test-only methods in prod logic unless they're also used in non-test files, purely for readability reasons. But I've had some trouble following that for testing the async methods in the past.

Maybe we can find a clever way to mock the timestamps or something to make that deterministic eventually.

ryanzhang22 (approved):
one comment on potential correctness

libkineto/test/AsyncE2ECpuTraceTest.cpp
scotts: Good call out; there's probably a race condition in there. I'll dig in.

libkineto/src/AsyncActivityProfilerHandler.h
scotts: I had convinced myself before that we couldn't, but I'm going to triple check.

scotts: Digging more, the fundamental problem is that `AsyncActivityProfilerHandler::profilerLoop()` uses the actual wallclock time to drive decisions. Because this is an end-to-end test, we definitely want to call that function; it's the function the profiling thread calls. We could put in some logic to avoid actually getting the wallclock time, but that's more intrusive than this change, which is just needed at startup.

I've put a bunch of test-only mechanisms in these sources recently because otherwise it's just not testable. Up until now, they've been mostly observational: marking that we've hit certain conditions. This is the first one that actual impacts production logic, but I think it's worth it so we can have an end-to-end, deterministic test.

libkineto/test/AsyncE2ECpuTraceTest.cpp
scotts: @ryanzhang22, yes, it's a problem, and we were actually generating an empty trace. :( I'm fundamentally changing the test, kinda counter to my comment below. The only way to do this deterministically is to drive the FSM manually, which means we don't call the profiling loop thread.

scotts:
@ryanzhang22, I changed the test completely. When I tugged on the problem with the end time that you brought up, it revealed that the test as written was against an empty profile. That's not a useful end-to-end test.

In order to make a useful, deterministic end-to-end test, I had to give up depending on the actual thread that calls `AsyncActivityProfilerHandler::profileLoop()`. Rather, we drive all FSM transitions ourselves, which allows us to control the clock and make sure that there is an actual operation to profile.

This does mean that our "end-to-end" test is not testing the production profiling thread. We have some others tests that do exercise the production profiling thread, so I think that's okay.

ryanzhang22 (approved):

PR #1508, opened 2026-07-30
Libkineto: escape quotes and control characters in JSON exporter trace attributes

muditgokhale2:
## Problem Statement

When profiling PyTorch models on TPUs (or when thread/device/op names contain quotes or special characters), Kineto generates trace JSON events where double quotes within names are unescaped (e.g. `custom_call_target="tpu_custom_call"` or `thread "with" quotes`).

This results in unescaped double quotes inside JSON string values, causing `JSONDecodeError: Expecting ',' delimiter` when parsing the trace output or loading it in Perfetto / PyTorch Profiler UI.

## Root Cause

1. `ChromeTraceLogger` in `src/output_json.cpp` formats string properties (`handleResourceInfo`, `handleDeviceInfo`, `handleOverheadInfo`) directly into JSON quotes without escaping internal double quotes (`"`).
2. `sanitizeStrForJSON` unconditionally replaced all backslashes (`\`) with forward slashes (`/`), destroying valid JSON escape sequences.

## Solution

1. Added `escapeQuotesForJSON` helper in `GenericTraceActivity.h` to escape double quotes (`"`), backslashes (`\`), and control characters.
2. Updated `handleResourceInfo`, `handleDeviceInfo`, and `handleOverheadInfo` in `src/output_json.cpp` to pass string names through `escapeQuotesForJSON`.
3. Updated `sanitizeStrForJSON` to preserve valid JSON backslash escape sequences.
4. Added unit tests in `AsyncActivityProfilerHandlerTest.cpp`.

### DIFF
--- a/libkineto/src/output_json.cpp
+++ b/libkineto/src/output_json.cpp
@@ -508,6 +508,14 @@ void ChromeTraceLogger::handleDeviceInfo(const DeviceInfo& info, int64_t time) {
     return;
   }
 
+  std::string name = info.name;
+  sanitizeStrForJSON(name);
+  escapeQuotesForJSON(name);
+
+  std::string label = info.label;
+  sanitizeStrForJSON(label);
+  escapeQuotesForJSON(label);
+
   time = transToRelativeTime(time);
   writeMetadataEvent(
       /*name=*/"process_name",
@@ -515,14 +523,14 @@ void ChromeTraceLogger::handleDeviceInfo(const DeviceInfo& info, int64_t time) {
       /*pid=*/info.id,
       /*tid=*/0,
       /*arg_key=*/"name",
-      /*arg_value=*/fmt::format("\"{}\"", info.name));
+      /*arg_value=*/fmt::format("\"{}\"", name));
   writeMetadataEvent(
       /*name=*/"process_labels",
       /*ts=*/time,
       /*pid=*/info.id,
       /*tid=*/0,
       /*arg_key=*/"labels",
-      /*arg_value=*/fmt::format("\"{}\"", info.label));
+      /*arg_value=*/fmt::format("\"{}\"", label));
   writeMetadataEvent(
       /*name=*/"process_sort_index",
       /*ts=*/time,
@@ -539,6 +547,10 @@ void ChromeTraceLogger::handleResourceInfo(
     return;
   }
 
+  std::string name = info.name;
+  sanitizeStrForJSON(name);
+  escapeQuotesForJSON(name);
+
   time = transToRelativeTime(time);
   int64_t tid = sanitizeTid(info.id);
   writeMetadataEvent(
@@ -547,7 +559,7 @@ void ChromeTraceLogger::handleResourceInfo(
       /*pid=*/info.deviceId,
       /*tid=*/tid,
       /*arg_key=*/"name",
-      /*arg_value=*/fmt::format("\"{}\"", info.name));
+      /*arg_value=*/fmt::format("\"{}\"", name));
   writeMetadataEvent(
       /*name=*/"thread_sort_index",
       /*ts=*/time,
@@ -564,6 +576,10 @@ void ChromeTraceLogger::handleOverheadInfo(
     return;
   }
 
+  std::string name = info.name;
+  sanitizeStrForJSON(name);
+  escapeQuotesForJSON(name);
+
   // TOOD: reserve pid = -1 for overhead but we need to rethink how to scale
   // this for other metadata
   time = transToRelativeTime(time);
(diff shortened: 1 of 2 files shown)


libkineto/include/GenericTraceActivity.h
```diff
@@ -26,6 +26,23 @@
...
+      case '"':  result += "\\\""; break;
+      case '\\': result += "\\\\"; break;
+      case '\n': result += "\\n";  break;
+      case '\r': result += "\\r";  break;
+      case '\t': result += "\\t";  break;
+      default:   result += c;      break;
+    }
+  }
+  return result;
+}
```
scotts: We should not have any JSON helpers in this header file. JSON related logic should stay in `output_json.h` and `output_json.cpp`, with a preference for keeping internal helpers just in the source file.

We also already have `escapeQuotesForJSON()` in `output_json.cpp`. Why have you not used that?

libkineto/test/AsyncActivityProfilerHandlerTest.cpp
```diff
@@ -848,3 +848,14 @@ TEST_F(AsyncClientTest, MemoryProfileRequestDrivesClientHooks) {
...
+TEST(AsyncActivityProfilerHandler, JSONEscapeTest) {
+  // Test escapeQuotesForJSON
+  std::string input1 = R"(op_name="fusion_1", custom_call_target="tpu_custom_call")";
+  std::string escaped1 = libkineto::escapeQuotesForJSON(input1);
+  EXPECT_EQ(escaped1, R"(op_name=\"fusion_1\", custom_call_target=\"tpu_custom_call\")");
+
+  std::string input2 = "C:\\path\\to\\file\nwith \"quotes\"";
+  std::string escaped2 = libkineto::escapeQuotesForJSON(input2);
+  EXPECT_EQ(escaped2, R"(C:\\path\\to\\file\nwith \"quotes\")");
+}
```
scotts: There's already an `OutputJsonTest.cpp` file that is the natural home for JSON related tests.

libkineto/src/output_json.cpp
```diff
@@ -69,11 +69,28 @@ constexpr std::string_view kDefaultLogFileFmt = "libkineto_activities_{}.json";
...
+          result += '\\';
+          continue;
+        }
+      }
+      result += "\\\\";
+    } else {
+      result += value[i];
+    }
+  }
+  value = std::move(result);
```
scotts: This is a regression on values such as `C:\users\person\trace.json`.

I think there's also confusion about the contracts of the existing functions. They are:
```
// For raw values that will be placed inside JSON quotes: escape everything.
std::string escapeForJSON(std::string_view s);
// For pre-escaped JSON fragments (metadataJson, metadata values): leave escapes intact.
void sanitizeJSONFragment(std::string& value);
```
The current PR does the escaping in JSON fragments that should have already been escaped.

scotts:
@muditgokhale2, thanks for your contribution! Please follow the steps to sign the CLA and address the PR comments.

libkineto/include/GenericTraceActivity.h
muditgokhale2: Thanks @scotts! You're completely right. I've removed the helper from `GenericTraceActivity.h` (reverted completely) and updated `handleDeviceInfo`, `handleResourceInfo`, and `handleOverheadInfo` in `output_json.cpp` to reuse the existing internal `escapeQuotesForJSON()` helper.

libkineto/test/AsyncActivityProfilerHandlerTest.cpp
muditgokhale2: Thanks! I've removed the test from AsyncActivityProfilerHandlerTest.cpp and moved it into libkineto/test/OutputJsonTest.cpp (ResourceInfoWithQuotesProducesValidJson).

libkineto/src/output_json.cpp
muditgokhale2: Thanks for clarifying the function contracts! I've reverted `sanitizeStrForJSON` back to its original implementation so pre-escaped JSON fragments and Windows paths (e.g. `C:\users\person\trace.json`) remain untouched. We now apply `escapeQuotesForJSON` strictly to raw string attributes (`info.name`, `info.label`) in `handleDeviceInfo`, `handleResourceInfo`, and `handleOverheadInfo`.

scotts:
@muditgokhale2, I can't move further until the CLA check passes.

muditgokhale2:
@meta-cla check

scotts:
@muditgokhale2, thank you for the changes! Everything looks great.

PR #1512, opened 2026-08-04
[Kineto] Kineto-native CUPTI PM sampling API

ryanzhang22:
Adds CUPTI PM sampling to Kineto, in the form of 2 initial core files. I'm purposely not making any decisions about how users interact with this API, how it flows into the output, etc. and deferring thinking about that for now.

The Kineto <> CUPTI interaction is managed by the CuptiPMSamplingApi, which contains simple wrapper methods over the CUPTI API to Configure, Start, Decode, and Disable PM sampling. This api is driven by the CuptiPMSamplingController, which provides and manages Kineto-side resources (final buffer, start/stop API, config validation).


libkineto/src/CuptiPMSamplingApi.cpp
```diff
@@ -0,0 +1,336 @@
...
+ */
+
+void checkCupti(CUptiResult status, const char* call) {
+  if (status == CUPTI_SUCCESS) {
+    return;
+  }
+
+  const char* error = nullptr;
+  cuptiGetResultString(status, &error);
+  throw std::runtime_error(
```
scotts: Let's use the `KINETO_THROW` utility landed in #1513. This applies to all of the `throw` statements.

libkineto/src/CuptiPMSamplingApi.h
```diff
@@ -0,0 +1,63 @@
...
+  virtual void stop();
+  virtual void disable();
+
+ private:
+  void configureCupti();
+  void resetImage();
+  CuptiPMSample decodeSample(size_t sampleIndex);
+
+  CuptiPMSamplingConfig config_;
+  std::vector<const char*> metricNamePtrs_;
```
scotts, line 61: It makes me a little nervous that we have a vector of pointers (`metricNamePtrs_`) whose lifetime is attached to a different object (`config_`). Because `metricNamePtrs_` and `config_` have the same lifetime, I think we're good, but I think we should document this relationship.

libkineto/src/CuptiPMSamplingApi.cpp
```diff
@@ -0,0 +1,336 @@
...
+
+  CUpti_PmSampling_SetConfig_Params setConfig{
+      CUpti_PmSampling_SetConfig_Params_STRUCT_SIZE};
+  setConfig.pPmSamplingObject = samplingObject_;
+  setConfig.configSize = configImage.size();
+  setConfig.pConfig = configImage.data();
+  setConfig.hardwareBufferSize = kHardwareBufferSizeBytes;
+  setConfig.samplingInterval =
+      static_cast<uint64_t>(config_.samplingInterval.count());
+  setConfig.triggerMode = CUPTI_PM_SAMPLING_TRIGGER_MODE_GPU_TIME_INTERVAL;
```
scotts, line 244: So this is not supported on A100s, which are still widely used. From [the docs](https://docs.nvidia.com/cupti/api/group__CUPTI__PM__SAMPLING__API.html#_CPPv4N28CUpti_PmSampling_TriggerMode48CUPTI_PM_SAMPLING_TRIGGER_MODE_GPU_TIME_INTERVALE):
> The trigger is based off of a fixed frequency source. The sample interval (set in the struct [CUpti_PmSampling_SetConfig_Params](https://docs.nvidia.com/cupti/api/structCUpti__PmSampling__SetConfig__Params.html#structcupti__pmsampling__setconfig__params)) is in terms of nanoseconds. Note: This trigger mode is not supported on Turing GPU architecture and GA100 GPU. It is supported on Ampere GA10x and later GPU architectures.

We can fall back to `CUPTI_PM_SAMPLING_TRIGGER_MODE_GPU_SYSCLK_INTERVAL`, but that's in clock cycles, not wallclock time, so we'd need to tune `samplingInterval`.

I think it's defensible to only provide this feature for more modern devices, but we'll need to do a runtime compatibility check.

libkineto/src/CuptiPMSamplingController.cpp
```diff
@@ -0,0 +1,231 @@
...
+    while (!stopRequested_.load(std::memory_order_relaxed)) {
+      if (decodeBatch(decodedSamples)) {
+        std::unique_lock<std::mutex> lock(waitMutex_);
+        waitCondition_.wait_for(lock, kDecodePollInterval, [this]() {
+          return stopRequested_.load(std::memory_order_relaxed);
+        });
+      }
+    }
+  } catch (...) {
+    logCurrentException("CUPTI PM sampling worker failed");
```
scotts, line 163: Do we have any way of indicating to the main thread that we're now in an error state?

libkineto/src/CuptiPMSamplingApi.cpp
```diff
@@ -0,0 +1,336 @@
...
+  CUpti_PmSampling_DecodeData_Params decode{
+      CUpti_PmSampling_DecodeData_Params_STRUCT_SIZE};
+  decode.pPmSamplingObject = samplingObject_;
+  decode.pCounterDataImage = counterDataImage_.data();
+  decode.counterDataImageSize = counterDataImage_.size();
+  CUPTI_PM_CALL(cuptiPmSamplingDecodeData(&decode));
+  if (decode.overflow != 0) {
+    LOG_FIRST_N(WARNING, 1)
+        << "CUPTI PM sampling hardware buffer overflowed; samples were lost";
+  }
```
scotts, line 323: I don't think we'll ever actually execute this condition because `CUPTI_ERROR_OUT_OF_MEMORY` is actually an error-return value, and we'll have thrown an exception in `CUPT_PM_CALL()` if we see it.

scotts: The docs also make it sound like this is something that may happen regularly, and that we may need to adjust our sampling frequency dynamically to deal with it. From https://docs.nvidia.com/cupti/main/main.html#pm-sampling-usage:
> The maximum sampling frequency without buffer overflow events depends on GPU (SM count), GPU load intensity, and overall system load. The bigger the chip and the higher the load, the lower the maximum frequency. If you need higher frequency, you can increase it until you get the overflow event which can be queried while decoding the pm sampling data using cuptiPmSamplingDecodeData() API.

libkineto/src/CuptiPMSamplingController.cpp
```diff
@@ -0,0 +1,231 @@
...
+    return false;
+  }
+  if (std::any_of(sample.values.begin(), sample.values.end(), [](double value) {
+        return !std::isfinite(value);
+      })) {
+    LOG(WARNING) << "CUPTI PM sample contains a non-finite metric value";
+    return false;
+  }
+  return true;
+}
```
scotts, line 239: [The docs](https://docs.nvidia.com/cupti/main/main.html#pm-sampling-usage) also have this warning:
> It is a known limitation that the first sample may occasionally contain outlier values. Discarding the first sample is recommended.

We could potentially do that in `validateSample()`, by just adding a `isFirstSample_` member.

libkineto/src/CuptiPMSamplingApi.cpp
```diff
@@ -0,0 +1,336 @@
...
+  CUPTI_PM_CALL(cuptiPmSamplingEnable(&enable));
+  samplingObject_ = enable.pPmSamplingObject;
+
+  CUpti_PmSampling_SetConfig_Params setConfig{
+      CUpti_PmSampling_SetConfig_Params_STRUCT_SIZE};
+  setConfig.pPmSamplingObject = samplingObject_;
+  setConfig.configSize = configImage.size();
+  setConfig.pConfig = configImage.data();
+  setConfig.hardwareBufferSize = kHardwareBufferSizeBytes;
+  setConfig.samplingInterval =
```
scotts, line 242: We should also probably set [`hwBufferAppendMode`](https://docs.nvidia.com/cupti/api/structCUpti__PmSampling__SetConfig__Params.html#_CPPv4N33CUpti_PmSampling_SetConfig_Params18hwBufferAppendModeE) here. I think it default-inits to `KEEP_OLDEST`, but `KEEP_LATEST` is probably more appropriate for us. This is also related to the `CUPTI_ERRO_OUT_OF_MEMORY` issue.

scotts:
I think it's fair to commit code without unit tests when it's separate from all production code and it's part of iterating on a new design. But we should make it part of the build - right now, it's not being compiled at all.

libkineto/src/CuptiPMSamplingController.h
```diff
@@ -0,0 +1,63 @@
...
+  void decodeLoop();
+  bool decodeBatch(std::vector<CuptiPMSample>& decodedSamples);
+  void drain(std::vector<CuptiPMSample>& decodedSamples);
+  void teardown();
+  bool validateConfig() const;
+  bool validateSample(const CuptiPMSample& sample) const;
+  void logCurrentException(const char* fallback);
+
+  CuptiPMSamplingConfig config_;
+  CuptiPMSamplingApi& api_;
```
scotts: I think we'll want `CuptiPMSamplingController` to actually own the `CuptiPMSamplingApi` object rather than having a reference. I think it's straightforward to just take the config in the constructor, and then create the object there.

libkineto/src/CuptiPMSamplingApi.h
```diff
@@ -0,0 +1,63 @@
...
+  CUpti_Profiler_Host_Object* hostObject_{nullptr};
+  CUpti_PmSampling_Object* samplingObject_{nullptr};
```
scotts, line 65: We should document that `hostObject_` and `samplingObject_` are non-owning references.

libkineto/src/CuptiPMSamplingApi.cpp
ryanzhang22: Good catch. I'll probably just swap this over to KEEP_LATEST + a log if possible. Dynamic sampling imo is going to need some thinking -- if an overflow happens we've already lost samples, and variable sample rate can be harder to debug. I'll definitely expose the buffer size as a tunable var though.

libkineto/src/CuptiPMSamplingApi.cpp
ryanzhang22: Hm, it would be nice to be able to support A100. Let me look into it.

libkineto/src/CuptiPMSamplingApi.cpp
scotts: I'd love a solution that doesn't expose the buffer size to end-users. But based on reading the docs, we'll need to get some experience with this in practice to understand how to use it effectively, and then from there, how to expose it to users.

libkineto/src/CuptiPMSamplingController.cpp
ryanzhang22: yeah, we can add a flag, similar to how async activity profiling handles it

libkineto/src/CuptiPMSamplingController.cpp
```diff
@@ -0,0 +1,253 @@
...
+    teardown();
+  }
+  return active_;
+}
+
+void CuptiPMSamplingController::stop() {
+  if (!prepared_) {
+    return;
+  }
+  if (active_) {
```
scotts, line 121: I think that we're using `active_` to track two things, and that may get us into some trouble. Right now, it's (I think?) tracking both the status of `api_.start()` and if `decodeThread_` needs to be joined.

But we're setting `active_ = false` in `teardown()`, and we call `teardown()` in the exception handler in `start()`. I think that means we could have a situation where `start()` fails, but we never join the thread in `stop()`. I'm not sure if we need to track the status of `api_` and `decoderThread_` separately, or if we can use one value and we just need to be careful about how use it.

ryanzhang22: Good point. I think that scenario isn't reachable in practice because nothing (except thread construction) can throw after `active_` is set to `true`, but it's worth being stricter about the lifecycles here to prevent anything from creeping in down the line.

We can probably just say that `active_` tracks whether the API is collecting samples and that we check `decoderThread_.joinable()` to join the decode thread on teardown. And we can inline `teardown` into `stop` to make it the single cleanup path to make sure that on exception both `active_` and `decoderThread_` are handled.

libkineto/CMakeLists.txt
```diff
@@ -27,6 +27,7 @@ set(KINETO_LIBRARY_TYPE "default" CACHE STRING
 set_property(CACHE KINETO_LIBRARY_TYPE PROPERTY STRINGS default shared static)
 option(KINETO_BUILD_TESTS "Build kineto unit tests" ON)
 option(KINETO_BUILD_BENCHMARKS "Build kineto benchmarks" OFF)
+option(KINETO_ENABLE_CUPTI_PM_SAMPLING "Build CUPTI PM sampling support" OFF)
```
scotts, line 30: Because this defaults off, nothing is actually compiling it in our CI. And our current CUDA CI job actually can't, because it's on 12.6: https://github.com/pytorch/kineto/blob/15f4784a3471d4db8b31720d528f2f57cd2b1f40/.github/workflows/linux_cuda_kineto.yml#L29

I think this is a good time to add the latest CUDA version to our matrix. We're going to want to do something like this: [https://github.com/meta-pytorch/torchcodec/blob/main/.github/workflows/linux_cuda_wheel.yaml#L73](https://github.com/meta-pytorch/torchcodec/blob/a87fcd5570845b10b917ca1cd97b938fd68734a0/.github/workflows/linux_cuda_wheel.yaml#L73). I think we should do [cuda_min_version, cuda_max_version], which becomes [12.6, 13.2].

libkineto/src/CuptiPMSamplingApi.cpp
```diff
@@ -0,0 +1,369 @@
...
+ *
+ * This source code is licensed under the BSD-style license found in the
+ * LICENSE file in the root directory of this source tree.
+ */
+
+#include <cuda.h>
+
+// hwBufferAppendMode and KEEP_LATEST were added in CUDA 12.8
+#if defined(HAS_CUPTI_PM_SAMPLING) && defined(CUDA_VERSION) && \
+    CUDA_VERSION >= 12080
```
scotts: Since the entire feature is version gated, I'd like to see if we can entirely control this with the build. That is, if someone turned on the build for this, but it's not supported, I think a compile-time error is better than compiling empty source files.

This is similar to what we're trying to do with device-specific code, where we're trying to get the compile time guards out of the source code and control it with the build.

libkineto/src/CuptiPMSamplingController.h
```diff
@@ -0,0 +1,63 @@
...
+ private:
+  void decodeLoop();
+  bool decodeBatch(std::vector<CuptiPMSample>& decodedSamples);
+  void drain(std::vector<CuptiPMSample>& decodedSamples);
+  bool validateConfig() const;
+  bool validateSample(const CuptiPMSample& sample) const;
+  void logCurrentException(const char* fallback);
+
+  CuptiPMSamplingConfig config_;
+  CuptiPMSamplingApi api_;
```
scotts, line 49: If we every want to mock the sampling API for tests, it might be easier for this to become a `std::unqiue_ptr<CuptiSamplingApi>`. It would be passed into the constructor, but it would be an ownership transfer. We can cross that bridge later.

libkineto/src/CuptiPMSamplingApi.cpp
```diff
@@ -0,0 +1,369 @@
...
+bool CuptiPMSamplingApi::decode(std::vector<CuptiPMSample>& samples) {
+  // Fetch the next result from CUPTI. Decodes the collected counters
+  // inline into the counterDataImage_.
+  CUpti_PmSampling_DecodeData_Params decode{
+      CUpti_PmSampling_DecodeData_Params_STRUCT_SIZE};
+  decode.pPmSamplingObject = samplingObject_;
+  decode.pCounterDataImage = counterDataImage_.data();
+  decode.counterDataImageSize = counterDataImage_.size();
+  CUPTI_PM_CALL(cuptiPmSamplingDecodeData(&decode));
+
```
scotts, line 324: We no longer check `decoder.overflow` - we probably should, just so that it shows up in the logs.

libkineto/src/CuptiPMSamplingApi.cpp
```diff
@@ -0,0 +1,369 @@
...
+  KINETO_THROW(
+      std::runtime_error,
+      std::string{call} +
+          " failed: " + (error != nullptr ? error : "unknown CUPTI error"));
+}
+
+#define CUPTI_PM_CALL(call)    \
+  do {                         \
+    checkCupti((call), #call); \
+  } while (false)
```
scotts: We actually already have a macro for this that we should probably reused, `CUPTI_CALL`: https://github.com/pytorch/kineto/blob/15f4784a3471d4db8b31720d528f2f57cd2b1f40/libkineto/src/DeviceUtil.h#L34-L47

libkineto/src/CuptiPMSamplingController.cpp
```diff
@@ -0,0 +1,253 @@
...
+    return false;
+  }
+  if (sample.rawEndTimestamp < sample.rawStartTimestamp) {
+    LOG(WARNING) << "CUPTI PM sample has an invalid timestamp interval";
+    return false;
+  }
+  if (std::any_of(sample.values.begin(), sample.values.end(), [](double value) {
+        return !std::isfinite(value);
+      })) {
+    LOG(WARNING) << "CUPTI PM sample contains a non-finite metric value";
```
scotts: All of these logs may best be `LOG_FIRST_N`.

libkineto/src/CuptiPMSamplingApi.cpp
```diff
@@ -0,0 +1,369 @@
...
+  counterDataImage_.resize(counterDataSize.counterDataSize);
+  resetImage();
+}
+
+void CuptiPMSamplingApi::resetImage() {
+  CUpti_PmSampling_CounterDataImage_Initialize_Params params{
+      CUpti_PmSampling_CounterDataImage_Initialize_Params_STRUCT_SIZE};
+  params.pPmSamplingObject = samplingObject_;
+  params.counterDataSize = counterDataImage_.size();
+  params.pCounterData = counterDataImage_.data();
```
scotts, line 269: I think we need to do error checking on all of the object members (`counterDataImage_`, `samplingObject_`, `hostObject_`) on all of the public member functions that use them. This guards against users calling the API in the wrong order.

libkineto/CMakeLists.txt
ryanzhang22: sg, updated the CI with 12.6/13.2 versions of `linux-cuda-[VERSION]-kineto-build-and-test`, plus conditional building PM sampling ON/OFF triggers.

libkineto/src/CuptiPMSamplingApi.cpp
ryanzhang22: I wanted to make sure that the wrapper we used for API calls explicitly throws an error, with the philosophy that the API should fail immediately and let the controller handle cleanup instead of continuing in a partially initialized state. `CUPTI_CALL` only logs, but I can add `CUPTI_CALL_THROW` or something to `DeviceUtil.h` if it's cleaner to generalize -- I don't have a strong preference either way.

libkineto/src/CuptiPMSamplingController.cpp
```diff
@@ -0,0 +1,248 @@
...
+    logCurrentException("Failed to start CUPTI PM sampling");
+    stop();
+  }
+  return active_;
+}
+
+void CuptiPMSamplingController::stop() {
+  // Notify the decode thread and join it if possible
+  if (decodeThread_.joinable()) {
+    stopRequested_.store(true, std::memory_order_relaxed);
```
scotts: We actually have a subtle race condition here. Because we don't hold `waitMutex_` before setting `stopRequested_`, and notifying `waitCondition_`, it's possible that can happen between when `waitCondition_.wait_for()` in `decodeLoop()` sees `stopRequested_ == false` and it registers as a waiter on the notification.

When that happens, `waitCondition_.notify_one()` notifies no one. The `waitCondition_.wait_for()` will timeout and we'll be okay, but it's not what we intend. We can be resilient to this by just making sure we grab the mutex:
```
  if (decodeThread_.joinable()) {
      {
        // Set the flag under waitMutex_ so the decode thread cannot miss the
        // notification between evaluating wait_for's predicate and blocking.
        std::lock_guard<std::mutex> lock(waitMutex_);
        stopRequested_ = true;
      }
      waitCondition_.notify_one();
      decodeThread_.join();
```

libkineto/src/CuptiPMSamplingApi.cpp
scotts: @ryanzhang22, gotcha on motivation. I do think it's worth making a generic macro named after these semantics in `DeviceUtils.h`.

libkineto/src/CuptiPMSamplingApi.cpp
```diff
@@ -0,0 +1,383 @@
...
+    KINETO_THROW(
+        std::runtime_error, "CUPTI PM sampling: counterDataImage_ is empty");
+  }
+}
+
+void CuptiPMSamplingApi::configureCupti() {
+  // CUpti_Device_GetChipName_Params stores device info used in other CUPTI
+  // calls.
+  CUpti_Device_GetChipName_Params chip{
+      CUpti_Device_GetChipName_Params_STRUCT_SIZE};
```
scotts, line 132: When CI actually starts picking these up for compilation, you're going to hit some errors because we're compiling with `-Wall -Wextra -pedantic -Werror`. That means`-Wmissing-field-initializers` applies, so we'll get tons of errors in this file in all of the `CUpti_Device_*Params` structs. We should do what the old range profiling code did, and also probably copy the comment over.

Here's the old comment and pragma: https://github.com/pytorch/kineto/blob/b5d29fc07379f57016f64b19ed6908f79b7f7b4b/libkineto/src/CuptiRangeProfilerApi.cpp#L27-L47

libkineto/src/CuptiPMSamplingController.cpp
ryanzhang22: oh interesting, i didn't realize `wait_for` wasn't atomic. Will add the mutex.

libkineto/src/CuptiPMSamplingApi.cpp
```diff
@@ -0,0 +1,383 @@
...
+        CUpti_Profiler_Host_Deinitialize_Params_STRUCT_SIZE};
+    params.pHostObject = hostObject_;
+    if (CUPTI_CALL(cuptiProfilerHostDeinitialize(&params)) == CUPTI_SUCCESS) {
+      hostObject_ = nullptr;
+    }
+  }
+
+  // Retain handles on failure so cleanup can be retried
+  if (samplingObject_ != nullptr || hostObject_ != nullptr) {
+    return;
```
scotts, line 381: Should we log? The destructor will retry, but after that the objects will leak.

libkineto/src/CuptiPMSamplingController.cpp
```diff
@@ -0,0 +1,248 @@
...
+}
+
+void CuptiPMSamplingController::drain(
+    std::vector<CuptiPMSample>& decodedSamples) {
+  // Called at the end to fully drain the hardware buffer
+  bool isBufferDrained;
+  do {
+    isBufferDrained = decodeBatch(decodedSamples);
+  } while (!isBufferDrained);
+}
```
scotts, line 197: I'm torn on what to do here - the drain is unbounded, depending on eventually get the right signal from the CUPIT calls. If CUPTI is ever in a bad state, then we can get into trouble here, particularly since this is called in the destructor. But I don't know what a reasonable bound is.

I'm okay if we don't do anything about it now - I'm not sure what we can do - but let's put a comment here that this is a potential problem.

libkineto/src/CuptiPMSamplingApi.cpp
ryanzhang22: The CUPTI_CALLs in `disable` would log the CUPTI-side error. I can add a final log to the destructor as a warning that things can leak.

libkineto/src/CuptiPMSamplingController.cpp
```diff
@@ -0,0 +1,248 @@
...
+  prepared_ = false;
+}
+
+const std::vector<std::string>& CuptiPMSamplingController::metricNames() const {
+  return config_.metricNames;
+}
+
+std::vector<CuptiPMSample> CuptiPMSamplingController::samples() const {
+  std::lock_guard<std::mutex> lock(samplesMutex_);
+  return samples_;
```
scotts: This is actually a potentially expensive operation, as we're going to do a deep-copy of the samples. If we're okay dropping the const and actually clearing out the samples, we can do:
```
  std::vector<CuptiPMSample> takeSamples() {
    std::lock_guard<std::mutex> lock(samplesMutex_);
    return std::exchange(samples_, {});
  }
```
We use `std::exchange` rather than `std::move` because moving would leaves `samples_` ["valid but unspecified"](https://en.cppreference.com/cpp/utility/move) and we're going to use it again.

Something Claude suggested that we can explore if we are spending a lot of time on allocations is flattening the layout:
```
  struct CuptiPMSampleSeries {
    std::vector<uint64_t> startTimestamps;  // N
    std::vector<uint64_t> endTimestamps;    // N
    std::vector<double> values;             // N * numMetrics, stride numMetrics
    size_t numMetrics{0};
  };
```
Note that `values` then is a 2D vector we maintain. But because it's more complicated, I think this is something we can explore *if* we observe a lot of overhead when copying.

scotts (approved):
Amazing start, thank you!

libkineto/src/CuptiPMSamplingController.cpp
scotts: I think `wait_for` is atomic with regards to the mutex, but when the other side is not coordinating on that mutex, the change can slip in between the read and the waiter-registration.

PR #1522, opened 2026-08-06
[Kineto] Used typed metadata in Kineto's XPU plugin

ryanzhang22:
Previously, all non-json output formats needed to parse JSON strings and convert them to typed representations before writing, which was super inefficient. In general, using JSON as the canonical Kineto output format is inflexible and leads to people needing to write hacky code to get around its limitations. What's more, when we want to add a new metadata field, it took some annoying work to have it reflect in non-JSON outputs.

Our solution to these problems is a visitor class over typed metadata fields. We have already migrated CUDA and ROCm to this and have gotten nice performance/code quality/quality of life improvements. This PR migrates XPU's `addMetadata` callsites to typed metadata and adds the fields to the catalog so that they can also get the benefits of this migration (and so that we can start to delete our JSON parsing code).

### DIFF
--- a/libkineto/src/plugin/xpupti/XpuptiActivityProfiler.cpp
+++ b/libkineto/src/plugin/xpupti/XpuptiActivityProfiler.cpp
@@ -7,14 +7,62 @@
  */
 
 #include "XpuptiActivityProfiler.h"
+#include "MetadataFieldCatalog.h"
 #include "ThrowUtil.h"
+#include "TypedMetadataJson.h"
 #include "XpuptiScopeProfilerApi.h"
 #include "XpuptiScopeProfilerSession.h"
 
 #include <fmt/ranges.h>
 #include <sycl/sycl.hpp>
+#include <utility>
 
 namespace KINETO_NAMESPACE {
+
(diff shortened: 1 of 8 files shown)


libkineto/src/plugin/xpupti/XpuptiActivityProfiler.cpp
```diff
@@ -7,15 +7,63 @@
...
 #include <sycl/sycl.hpp>
+#include <utility>
 
 namespace KINETO_NAMESPACE {
 
+namespace DevicePropertyFields = libkineto::DevicePropertyMetadataFields;
+
+namespace {
+
+void visitXpuDeviceMetadata(
```
ryanzhang22, line 26: not used yet but it would be once protobuf is added to OSS

ryanzhang22:
@gujinghui Can you take a look at this PR to migrate XPU code to use the new Kineto metadata format? Pre-warning that most of this code was Codex-generated. Let me know if you need some more context for this, thanks.

scotts (approved):

PR #1527, opened 2026-08-07
[Kineto] Enable registering CUPTI PM sampling as a child profiler

ryanzhang22:
Adds the PM sampling activity profiler extending the activity profiler interfaces. It's kind of irritating that we have this dependency because PM sampling is not activity profiling, but this is by far the easiest way to implement this without doing a major refactor of how Kineto handles child profilers.

Mostly direct routing logic, except for `buildTraceBuffer`, which performs the same timestamp conversion that activity profiling does and then pushes collected metrics into `GenericTraceActivity`'s counter handling methods. This means that activity profiling must succeed in instantiating the timestamp convertor, or PM sampling will log an error and return early.

Next is handling the output logic.

### DIFF
--- a/libkineto/src/CuptiPMSamplingProfiler.cpp
+++ b/libkineto/src/CuptiPMSamplingProfiler.cpp
@@ -1,0 +1,174 @@
+/*
+ * Copyright (c) Meta Platforms, Inc. and affiliates.
+ * All rights reserved.
+ *
+ * This source code is licensed under the BSD-style license found in the
+ * LICENSE file in the root directory of this source tree.
+ */
+
+#include "CuptiPMSamplingProfiler.h"
+
+#include <algorithm>
+#include <set>
+#include <string>
+#include <utility>
+
+#include "ActivityType.h"
+#include "CuptiTimestamp.h"
+#include "Logger.h"
+#include "libkineto.h"
+#include "output_base.h"
+
+namespace KINETO_NAMESPACE {
+namespace {
+
+const std::string kProfilerName{"CUPTI PM Sampling"};
+const std::set<libkineto::ActivityType> kSupportedActivities{
+    libkineto::ActivityType::HARDWARE_COUNTERS};
+
+} // namespace
+
+CuptiPMSamplingSession::CuptiPMSamplingSession(
+    const CuptiPMSamplingConfig& config)
+    : controller_(config) {}
+
+CuptiPMSamplingSession::~CuptiPMSamplingSession() = default;
+
+bool CuptiPMSamplingSession::prepare() {
+  // PM sampling and activity profiling share the same timestamp callbacks.
+  // This can only run after activity profiling enables the callbacks in
+  // enableGpuTracing
+  if (!isCuptiTimestampSourceReady()) {
+    LOG(WARNING) << "CUPTI PM sampling requires timestamp setup from the "
+                    "CUPTI Activity profiler";
+    return false;
+  }
+
+  // prepare() enables CUPTI PM sampling and acquires exclusive access during
+  // the parent profiler's warmup phase. Collection itself begins in start().
+  if (!controller_.prepare()) {
+    LOG(WARNING) << "CUPTI PM sampling failed to prepare CUDA device "
+                 << controller_.deviceId();
+    return false;
+  }
+  return true;
+}
+
+void CuptiPMSamplingSession::start() {
+  controller_.start();
+}
+
+void CuptiPMSamplingSession::stop() {
+  if (controller_.stop()) {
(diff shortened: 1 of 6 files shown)


libkineto/src/CuptiPMSamplingProfiler.h
```diff
@@ -0,0 +1,74 @@
...
+  [[nodiscard]] std::unique_ptr<libkineto::CpuTraceBuffer> buildTraceBuffer();
+
+  int32_t deviceId_;
+  CuptiPMSamplingController controller_;
+  std::unique_ptr<libkineto::CpuTraceBuffer> traceBuffer_;
+};
+
+// The naming is unfortunate here, because CUPTI PM sampling is not activity
+// profiling. However, Kineto's registration logic is based on
+// IActivityProfiler, so this is the simplest way to introduce a new profiler.
```
scotts, line 53: I'm open to refactoring that to `IProfiler`, but that would require some internal changes as well.

libkineto/src/CuptiPMSamplingProfiler.h
```diff
@@ -0,0 +1,74 @@
...
+class CuptiPMSamplingProfiler final : public libkineto::IActivityProfiler {
+ public:
+  explicit CuptiPMSamplingProfiler(CuptiPMSamplingConfig config);
+
+  [[nodiscard]] const std::string& name() const override;
+  [[nodiscard]] const std::set<libkineto::ActivityType>& availableActivities()
```
scotts, line 59: `availableActivities()` is another misnomer. We can just live with it for now.

libkineto/src/CuptiPMSamplingProfiler.cpp
```diff
@@ -0,0 +1,164 @@
...
+  if (!isCuptiTimestampSourceReady()) {
+    LOG(WARNING) << "CUPTI PM sampling requires timestamp setup from the "
+                    "CUPTI Activity profiler";
+    return false;
+  }
+  return controller_.prepare();
+}
+
+void CuptiPMSamplingSession::start() {
+  static_cast<void>(controller_.start());
```
scotts: Rather than ignoring the return value, is there some useful action we can take? The controller itself would have already logged.

ryanzhang22: hmm... there's not much i can think of that doesn't duplicate that the controller already does. We could make `controller_.start()` return void for now since it's not used.

libkineto/src/CuptiPMSamplingProfiler.cpp
```diff
@@ -0,0 +1,173 @@
...
+  // PM sampling and activity profiling share the same timestamp callbacks.
+  // This can only run after activity profiling enables the callbacks in
+  // enableGpuTracing
+  if (!isCuptiTimestampSourceReady()) {
+    LOG(WARNING) << "CUPTI PM sampling requires timestamp setup from the "
+                    "CUPTI Activity profiler";
+    return false;
+  }
+
+  if (!controller_.prepare()) {
```
scotts, line 49: I think it's worth calling out that we get exclusive access to do PM sampling at this point, which is during the warmup phase of a schedule. I don't think that's wrong, but it's noteworthy.

libkineto/src/CuptiPMSamplingProfiler.cpp
```diff
@@ -0,0 +1,173 @@
...
+  buffer->span = libkineto::TraceSpan{0, 0, kProfilerName};
+
+  bool hasActivities = false;
+  const auto samples = controller_.takeSamples();
+  const auto& metricNames = controller_.metricNames();
+  for (const auto& sample : samples) {
+    const auto start = convertCuptiTimestamp(sample.rawStartTimestamp);
+    const auto end = convertCuptiTimestamp(sample.rawEndTimestamp);
+
+    if (!hasActivities) {
```
scotts: We could just do `if (buffer->activities.empty())`.

libkineto/src/CuptiPMSamplingProfiler.cpp
```diff
@@ -0,0 +1,173 @@
...
+  return true;
+}
+
+void CuptiPMSamplingSession::start() {
+  controller_.start();
+}
+
+void CuptiPMSamplingSession::stop() {
+  controller_.stop();
+  traceBuffer_ = buildTraceBuffer();
```
scotts: We might want to guard modifying `traceBuffer_` on whether `controller._stop()` actually did a stop operation. That would probably require having `controller_.stop()` return a bool.

scotts (approved):

Issue #1478, opened 2026-07-10
RFC: Upstreaming Kineto to PyTorch


scotts:
We plan to upstream Kineto into PyTorch. As a part of that process, we wanted to get feedback from the community about our plans.

### What is Kineto and how is it used?

**Kineto is a C++ library for profiling GPUs.** It directly supports devices from Nvidia, AMD and Intel, and it has extension mechanisms for out-of-tree support. While implemented generically, **Kineto is primarily used within PyTorch**. It is included as a submodule in PyTorch and is statically linked into PyTorch at compile time.

### Why is Kineto separate from PyTorch?
Kineto grew out an [internal Meta effort](https://research.facebook.com/publications/datacenter-scale-analysis-and-optimization-of-gpu-machine-learning-workloads/) to understand and optimize GPU usage fleet-wide. It started as a generic GPU profiling library that gained traction because it was light-weight and enabled on-demand (asynchronous) tracing. Early on, it was used to understand the behavior of Caffe2 jobs.

Because Kineto was so useful internally, there was interest in open sourcing it. A natural home was PyTorch, as most GPU usage had moved to PyTorch models. **Kineto was added as a PyTorch submodule, and open sourced as a PyTorch GitHub project.**

One of the initial goals in open sourcing Kineto was for it to be a general purpose GPU profiling library that could be used on its own, including in frameworks other than PyTorch. 

### Why upstream Kineto into PyTorch?

The initial goal of Kineto being used outside of PyTorch was never realized. But that’s not a bad thing: **it’s better for PyTorch to have a dedicated GPU profiling layer**. Kineto is deeply integrated into PyTorch, and it’s common for new profiling features to require changes to both Kineto and PyTorch itself.

There are significant costs to maintaining Kineto in its own repo:

- **When new profiling features do require changes to both Kineto and PyTorch, we are forced to merge multiple PRs across both repos.** This process typically requires making backwards compatible changes to Kineto, bumping the Kineto submodule hash in PyTorch, changing PyTorch to take advantage of the new code, then finally removing any backwards compatibility layer from Kineto. If PyTorch’s GPU profiling layer was in the same repo, this would be one PR.
- **Kineto’s CI replicates parts of PyTorch’s CI.** Specifically, the most important tests for Kineto are the PyTorch profiling tests for each supported device. Running these tests requires compiling PyTorch from source with the Kineto changes. The Kineto CI to enable this testing is a subset of PyTorch’s CI.
[...]

PR #1529, opened 2026-08-10
Make TempTraceFile work on Windows

scotts:
More work in support of upstreaming Kineto to PyTorch: https://github.com/pytorch/kineto/issues/1478

**Problem** — Five tests are excluded from Windows builds for one reason: they compile `TestUtils.cpp`, which creates temporary trace files with mkstemps from unix "unistd.h". MSVC has neither. PyTorch's CI does build C++ tests on Windows, so that exclusion is coverage we would rather not carry over.

**Approach** — Replace mkstemps with the same guarantee assembled from portable pieces: pick a name in `std::filesystem::temp_directory_path`, create it with `O_EXCL` (`_O_EXCL` on Windows), and retry on collision. Exclusive creation is the property that mattered; the random component only has to keep retries rare. Removal moves to `std::filesystem::remove`, so the only conditional code left is the one call that creates a file without racing.

Paths are built with `std::filesystem::path::generic_string()`, so they use forward slashes on Windows too. Callers pass them to Kineto as log files and read them back out of trace JSON, and a separator that needs no escaping survives both.

Four tests come off the Windows exclusion list:
`ActivityProfilerControllerTest`, `AsyncActivityProfilerHandlerTest`, `OutputJsonTest`, and `CuptiActivityProfilerTest`.

Two tests stay excluded for reasons that are not `TestUtils`, and the CMakeLists now says which is which per test rather than attributing all of them to `unistd.h`.
1. `AsyncE2ECpuTraceTest` drives an on-demand config, and `Config` only accepts on-demand trace files under a directory defaulting to `/tmp`.
2. `ConfigLoaderPollThreadExceptionTest` calls `setenv`.
 
Two more, `SyncActivityProfilerHandlerTest` and `GenericActivityProfilerTeardownTest`, use nothing Unix-specific at all but have never been built on Windows; enabling those deserves its own change and its own Windows run.

Because we don't have Windows tests in Kineto, bumping the submodule hash in PyTorch will be the real test.

### DIFF
--- a/libkineto/test/CMakeLists.txt
+++ b/libkineto/test/CMakeLists.txt
@@ -51,6 +51,10 @@ target_link_libraries(ActivityProfilerControllerTest PRIVATE
 gtest_discover_tests(ActivityProfilerControllerTest)
 
 # AsyncActivityProfilerHandlerTest
+# Builds on Windows, but one of its cases does not: MetadataJsonFormatingTest
+# sets the environment with unix setenv, so the source compiles that case out
+# under _WIN32. The exclusion is per test rather than per target here because
+# the other fifteen cases run there fine.
 add_executable(AsyncActivityProfilerHandlerTest
     AsyncActivityProfilerHandlerTest.cpp
     TestUtils.cpp)
--- a/libkineto/test/CMakeLists.txt
+++ b/libkineto/test/CMakeLists.txt
@@ -41,9 +41,29 @@ target_link_libraries(ConfigLoaderTest PRIVATE
 gtest_discover_tests(ConfigLoaderTest)
 
 # ActivityProfilerControllerTest
-# Not supported on Windows, uses unix "unistd.h"
+add_executable(ActivityProfilerControllerTest
+    ActivityProfilerControllerTest.cpp
+    TestUtils.cpp)
+target_link_libraries(ActivityProfilerControllerTest PRIVATE
+    gtest_main
+    kineto_base kineto_api
+    ${XPU_XPUPTI_LIBRARY})
+gtest_discover_tests(ActivityProfilerControllerTest)
+
+# AsyncActivityProfilerHandlerTest
+add_executable(AsyncActivityProfilerHandlerTest
+    AsyncActivityProfilerHandlerTest.cpp
+    TestUtils.cpp)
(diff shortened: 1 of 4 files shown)


libkineto/test/TestUtils.cpp
```diff
@@ -11,39 +11,111 @@
...
+// provides for the name it settles on. Returns false when the name is already
+// taken so the caller can pick another, and swallows every other failure the
+// same way: the caller reports after exhausting its attempts.
+bool createExclusive(const std::string& path) {
+#ifdef _WIN32
+  int fd = -1;
+  const errno_t err = _sopen_s(
+      &fd,
+      path.c_str(),
+      _O_CREAT | _O_EXCL | _O_RDWR,
```
ryanzhang22, line 49: I was googling these objects and it looks like they are included through `#include <fcntl.h>`? Are we missing that include in this file?

libkineto/test/CMakeLists.txt
```diff
@@ -41,38 +41,42 @@ target_link_libraries(ConfigLoaderTest PRIVATE
...
+    ActivityProfilerControllerTest.cpp
+    TestUtils.cpp)
+target_link_libraries(ActivityProfilerControllerTest PRIVATE
+    gtest_main
+    kineto_base kineto_api
+    ${XPU_XPUPTI_LIBRARY})
+gtest_discover_tests(ActivityProfilerControllerTest)
+
+# AsyncActivityProfilerHandlerTest
+add_executable(AsyncActivityProfilerHandlerTest
```
ryanzhang22, line 58: codex is calling out the `setenv/unsetenv` calls in `AsyncActivityProfilerHandlerTest` as additional windows-incompatible calls that might need to be migrated before this starts running in CI

libkineto/test/TestUtils.cpp
scotts: Potentially - I'll investigate as well.

libkineto/test/CMakeLists.txt
scotts: I'll dig in there as well.

The problem we have here is that we won't know for certain this is okay until we bump the submodule hash in PyTorch. And I don't think it's worth it to spin up Windows tests in Kineto right now.

ryanzhang22 (approved):

PR #1530, opened 2026-08-10
[Kineto] Wire CUPTI PM Sampling up to output json and register it as a child profiler

ryanzhang22:
Does the next layer of wiring for PM sampling. On the output side, we integrate with the existing Counter handlers in output_json.cpp. One the input side, we take in device id and sampling metric names as args, then pass this down into the PM sampling session constructor.


libkineto/src/CuptiPMSamplingProfiler.h
```diff
@@ -53,8 +53,6 @@ class CuptiPMSamplingSession final
 // IActivityProfiler, so this is the simplest way to introduce a new profiler.
 class CuptiPMSamplingProfiler final : public libkineto::IActivityProfiler {
  public:
-  explicit CuptiPMSamplingProfiler(CuptiPMSamplingConfig config);
```
ryanzhang22, line 56: I should've left a comment on this before, but this is removed as it was a placeholder for the CuptiPMSamplingConfig object which is now constructed from the kineto config in `configure()`.

libkineto/include/Config.h
```diff
@@ -228,6 +228,15 @@ class Config : public AbstractConfig {
...
 
+  [[nodiscard]] const std::vector<std::string>& cuptiPMSamplingMetricNames()
+      const {
+    return cuptiPMSamplingMetricNames_;
+  }
+
+  [[nodiscard]] int32_t cuptiPMSamplingDeviceId() const {
+    return cuptiPMSamplingDeviceId_;
+  }
+
```
scotts, line 239: More of an observation: it looks like our configs are not "generic." Instead of trying to abstract config options per device, we just expose the device-specific option directly. That means the config will be the union of all device options.

libkineto/src/CuptiPMSamplingProfiler.cpp
```diff
@@ -145,13 +146,28 @@ const std::set<libkineto::ActivityType>& CuptiPMSamplingProfiler::
...
         const std::set<libkineto::ActivityType>& activityTypes,
-        const libkineto::Config& /*config*/) {
+        const libkineto::Config& config) {
   if (activityTypes.find(libkineto::ActivityType::HARDWARE_COUNTERS) ==
       activityTypes.end()) {
     return nullptr;
   }
 
-  auto session = std::make_unique<CuptiPMSamplingSession>(config_);
+  // Translate from the generic Kineto config into the sampling-specific config
```
scotts: Nit: as noted above, they're not really generic.

scotts (approved):

PR #1531, opened 2026-08-10
Fix a ConfigLoaderTest flake and three poll-thread bugs it uncovered

scotts:
Chasing an intermittent `ConfigLoaderTest` failure on Mac CI turned up four real
defects:

### 1. `ConfigLoaderTest.ThreadRepeatedlyPollsOnDemandConfig` flake

**Problem** — The test asserts the poll thread read the on-demand config twice,
but waits on the thread's loop counter to decide when to look. The counter
advances once per iteration whether or not that iteration polled. An iteration
that wakes a hair short of its deadline skips the read, bumps the counter anyway,
and polls on the next pass. The test wakes on the bumped counter, stops the
thread, and finds one read.

**Fix** — Wait on the read count the test actually asserts, using the handshake
pattern `ConfigLoaderPollThreadExceptionTest` already uses.

### 2. Lost wakeup on shutdown

**Problem** — `stopUpdateThread()` can block for the whole poll interval, up to
300 seconds. The stopper raises the flag then notifies, but the loop waits on a
bare timeout and only tests the flag after waking, so a notify that lands while
the thread is still on its way into the wait is dropped. This is the stall the
loop's own comment says the notify exists to prevent. It is intermittent — the
same binary tore down instantly on one run and took 300.01 s on the next.

**Fix** — Give the wait a predicate on the stop flag. Full `ctest` drops from
300 s to 5 s.

### 3. Poll cadence on the wall clock

**Problem** — The loop times its reload deadlines against `system_clock`, which
is not monotonic. A backward NTP step holds off both reloads for the size of the
step. At small scale, the loop sleeps on a condition variable measured against
the steady clock and then compares wall-clock timestamps — the drift between the
two is what produces the flake in 1.

**Fix** — Schedule on `steady_clock`, the clock the condition variable already
uses. An unused `system_clock` parameter on the daemon path goes with it.

### 4. Cadence lost after a failed base-config reload

**Problem** — When a base-config reload throws, on-demand polling drops to once
every 300 seconds instead of every 5, because the cadence refresh sits inside the
same `try` block as the reload. On a host whose base-config reads keep failing,
profiling requests sit unnoticed for minutes.

**Fix** — Refresh the cadence after the guard. A failed reload leaves the
previous config in force, and that config's cadence is the one to keep polling
at.

### Testing

82/82 tests pass, suite wall time 300 s → 5 s. The flaky test survives 30
isolated-process runs and 15 single-process suite runs. Each fix was confirmed to
fail without it: forcing the poll thread to wake at half its interval reproduces
the CI failure exactly against the old test, and the new cadence assertion
reports 300 against the old ordering.

### DIFF
--- a/libkineto/test/ConfigLoaderTest.cpp
+++ b/libkineto/test/ConfigLoaderTest.cpp
@@ -284,13 +284,14 @@ TEST_F(ConfigLoaderTest, ThreadPollsDaemonAndDispatchesOnDemandConfig) {
 // The thread re-polls the on-demand config on its update cadence, so over a
 // couple of intervals it reads the daemon more than once.
 //
-// This waits on the probe's read count rather than the thread's iteration
-// counter. An iteration can complete without reading: the thread wakes from a
-// wait_for measured against the steady clock, then gates the read on a
-// system_clock deadline, so a spurious wakeup or a small divergence between the
-// two clocks leaves it a hair short of the deadline. It skips the read, bumps
-// the iteration counter, and reads on the following iteration instead. Waiting
-// for two iterations therefore does not mean two reads have happened.
+// This waits on the probe's read count, not the thread's iteration counter:
+// the counter tracks loop passes, and a pass need not read. The loop serves a
+// base-config deadline and an on-demand deadline, sleeping until whichever
+// falls first, so a pass woken for the base config finds the on-demand
+// deadline still ahead and reads nothing. A pass woken for the on-demand
+// deadline can skip it too, since the wait only promises to reach that
+// deadline while the read is gated on having passed it. The counter advances
+// either way, so waiting for two passes would not mean two reads happened.
 //
 // This runs at the real cadence, so it takes a few seconds. It cannot be sped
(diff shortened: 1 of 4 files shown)


libkineto/test/ConfigLoaderTest.cpp
```diff
@@ -244,6 +284,14 @@ TEST_F(ConfigLoaderTest, ThreadPollsDaemonAndDispatchesOnDemandConfig) {
 // The thread re-polls the on-demand config on its update cadence, so over a
 // couple of intervals it reads the daemon more than once.
 //
+// This waits on the probe's read count rather than the thread's iteration
+// counter. An iteration can complete without reading: the thread wakes from a
+// wait_for measured against the steady clock, then gates the read on a
+// system_clock deadline, so a spurious wakeup or a small divergence between the
```
ryanzhang22: "then gates the read on a system_clock deadline" -- is this still an accurate description? I couldn't find any remaining usage of `system_clock` in these files.

ryanzhang22 (approved):

libkineto/test/ConfigLoaderTest.cpp
scotts: Fixed.

Issue #1275, opened 2026-02-25
Reject CUPTI Buffer when full for 12.9+


sraikund16:
According to NV, passing in a null pointer during Cupti request should result in a request rejection in 12.9+. For versions lower than this, there was a segfault previously. 

Code pointer: https://github.com/pytorch/kineto/blob/main/libkineto/src/CuptiActivityApi.cpp#L150

PR #1534, opened 2026-08-11
(bugfix): CUPTI buffer rejection on v27+

gtnv:
without this gate the buffer lmit in kineto can crash the process on pre-12.9 CUPTI before tracing actually shuts down and finishes.

the pr gates Kineto's null-buffer rejection on CUPTI v27+ with fallback below 27.

References:
- [CUPTI activity-buffer callback contract](https://docs.nvidia.com/cupti/12.9/api/group__CUPTI__ACTIVITY__API.html#_CPPv432CUpti_BuffersCallbackRequestFunc)
- [CUDA 12.9 release notes; missing-buffer crash fix](https://docs.nvidia.com/cupti/release-notes/release-notes.html#updates-in-cuda-12-9)
- [CUPTI version API — v27 maps to CUDA 12.9](https://docs.nvidia.com/cupti/12.9/api/group__CUPTI__VERSION__API.html#c.CUPTI_API_VERSION)

relevant/simialr issue: https://github.com/pytorch/kineto/issues/1275

cc @ryanzhang22

### DIFF
--- a/libkineto/src/CuptiActivityApi.cpp
+++ b/libkineto/src/CuptiActivityApi.cpp
@@ -56,9 +56,6 @@ inline void reenableCuptiCallbacks_(CuptiCallbackApi& cbapi) {
   }
 }
 
-CuptiActivityApi::CuptiActivityApi(uint32_t cuptiVersion)
-    : canRejectBuffer_(cuptiVersion >= kCuptiBufferRejectionMinVersion) {}
-
 CuptiActivityApi& CuptiActivityApi::singleton() {
   static auto* instance = new CuptiActivityApi();
   return *instance;
@@ -152,22 +149,21 @@ void CuptiActivityApi::bufferRequested(
   std::lock_guard<std::mutex> guard(mutex_);
   LOG(VERBOSE) << "CUPTI buffer requested";
 
-  *buffer = nullptr;
-  *size = 0;
-  *maxNumRecords = 0;
-
-  const auto retainedBufferCount = allocatedGpuTraceBuffers_.size() +
-      (readyGpuTraceBuffers_ ? readyGpuTraceBuffers_->size() : 0);
   if (!stopCollection &&
-      static_cast<int64_t>(retainedBufferCount) >= maxGpuBufferCount_) {
+      static_cast<int64_t>(allocatedGpuTraceBuffers_.size()) >=
+          maxGpuBufferCount_) {
     stopCollection = true;
-    LOG(WARNING) << "Exceeded max GPU buffer count (" << retainedBufferCount
+    LOG(WARNING) << "Exceeded max GPU buffer count ("
+                 << allocatedGpuTraceBuffers_.size()
                  << " >= " << maxGpuBufferCount_ << ") - terminating tracing";
   }
   // CUPTI fixed a crash when clients reject a buffer request in API v27
   // (CUDA 12.9). Older or unknown versions must receive a valid buffer while
   // the controller observes stopCollection and terminates tracing.
   if (stopCollection && canRejectBuffer_) {
+    *buffer = nullptr;
+    *size = 0;
+    *maxNumRecords = 0;
     return;
   }
 
@@ -176,6 +172,8 @@ void CuptiActivityApi::bufferRequested(
   *size = kBufSize;
 
   allocatedGpuTraceBuffers_[*buffer] = std::move(buf);
+
+  *maxNumRecords = 0;
 }
 
 std::unique_ptr<CuptiActivityBufferMap> CuptiActivityApi::activityBuffers() {
@@ -251,7 +249,6 @@ void CuptiActivityApi::clearActivities() {
   {
     std::lock_guard<std::mutex> guard(mutex_);
     if (allocatedGpuTraceBuffers_.empty()) {
-      readyGpuTraceBuffers_ = nullptr;
       return;
     }
   }
--- a/libkineto/src/CuptiActivityApi.cpp
+++ b/libkineto/src/CuptiActivityApi.cpp
@@ -56,12 +56,8 @@ inline void reenableCuptiCallbacks_(CuptiCallbackApi& cbapi) {
   }
 }
 
-CuptiActivityApi::CuptiActivityApi() {
-  CUPTI_CALL(cuptiGetVersion(&cuptiVersion_));
-}
-
 CuptiActivityApi::CuptiActivityApi(uint32_t cuptiVersion)
-    : cuptiVersion_(cuptiVersion) {}
+    : canRejectBuffer_(cuptiVersion >= kCuptiBufferRejectionMinVersion) {}
 
 CuptiActivityApi& CuptiActivityApi::singleton() {
   static auto* instance = new CuptiActivityApi();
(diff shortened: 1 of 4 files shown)


libkineto/src/CuptiActivityApi.cpp
```diff
@@ -55,6 +56,13 @@ inline void reenableCuptiCallbacks_(CuptiCallbackApi& cbapi) {
   }
 }
 
+CuptiActivityApi::CuptiActivityApi() {
+  CUPTI_CALL(cuptiGetVersion(&cuptiVersion_));
```
scotts: It's not okay to call CUPTI APIs directly in the constructor because we actually create the `CuptiActivityApi` singleton even when doing CPU only profiling. We intentionally defer initializing CUPTI until as late as possible. Let's move this call into `enableCuptiActivities()`.

libkineto/src/CuptiActivityApi.cpp
```diff
@@ -153,14 +161,15 @@ void CuptiActivityApi::bufferRequested(
...
-    return;
+    // CUPTI fixed a crash when clients reject a buffer request in API v27
+    // (CUDA 12.9). Older or unknown versions must receive a valid buffer while
+    // the controller observes stopCollection and terminates tracing.
+    if (cuptiVersion_ >= kCuptiBufferRejectionMinVersion) {
+      *buffer = nullptr;
+      *size = 0;
+      *maxNumRecords = 0;
+      return;
+    }
```
scotts: As written, this is trading a segfault for an OOM; see https://github.com/pytorch/kineto/pull/1362. Neither is good.

I think we can keep this logic, but then try to mitigate the problems by also checking `stopCollection` in `bufferCompleted()`. If it's true, just remove the buffer and do not move it to `readyGpuTraceBuffers_`. That way we avoid keeping live buffers when `stopCollection` is true.

The current logic: https://github.com/pytorch/kineto/blob/8b1c92370b157bcd514cb6883649c865371f2fce/libkineto/src/CuptiActivityApi.cpp#L278-L294

At line 286, I think we can do:
```
  if (stopCollection) {
    // Tracing is terminating and these records are already forfeit — on CUPTI
    // >= v27 we reject the request outright and CUPTI drops them for us. Free
    // the buffer rather than accumulating it in readyGpuTraceBuffers_, which
    // has no cap.
    ++droppedBufferCount_;
    allocatedGpuTraceBuffers_.erase(it);
    return;
  }
```
I think that will make OOMs less likely, and still avoid the segfault.

libkineto/src/CuptiActivityApi.cpp
gtnv: oh ok gotcha. 👍 fixed

libkineto/src/CuptiActivityApi.cpp
gtnv: right on. i added that and made the limit count ready buffers too

libkineto/src/CuptiActivityApi.cpp
```diff
@@ -55,6 +56,9 @@ inline void reenableCuptiCallbacks_(CuptiCallbackApi& cbapi) {
   }
 }
 
+CuptiActivityApi::CuptiActivityApi(uint32_t cuptiVersion)
+    : canRejectBuffer_(cuptiVersion >= kCuptiBufferRejectionMinVersion) {}
+
```
scotts: We should get rid of this constructor so that only `enableCuptiActivites()` explicitly sets `canRejectBuffer_`. I get that it helps make the tests simple, but it's at the cost of making the production paths confusing - a reader may assume that we're somehow figuring this out dynamically.

libkineto/src/CuptiActivityApi.cpp
```diff
@@ -147,19 +151,23 @@ void CuptiActivityApi::bufferRequested(
     size_t* maxNumRecords) {
   std::lock_guard<std::mutex> guard(mutex_);
   LOG(VERBOSE) << "CUPTI buffer requested";
-  if (static_cast<int64_t>(allocatedGpuTraceBuffers_.size()) >=
-      maxGpuBufferCount_) {
+
+  *buffer = nullptr;
+  *size = 0;
+  *maxNumRecords = 0;
```
scotts: This logic should stay inside the if condition - it should clearly be what we do on the failure path.

libkineto/src/CuptiActivityApi.cpp
```diff
@@ -147,19 +151,23 @@ void CuptiActivityApi::bufferRequested(
...
+
+  const auto retainedBufferCount = allocatedGpuTraceBuffers_.size() +
+      (readyGpuTraceBuffers_ ? readyGpuTraceBuffers_->size() : 0);
```
scotts: Why are we including `readyGpuTraceBuffers_` in the calculation? If the motivation is to make it "more accurate," I think that's unnecessary and more confusing.

scotts: On talking with Claude on this issue, I get more the motivation of including `readyGpuTraceBuffers_`, as it's the thing growing unbounded. But that is a big behavior change. If we're going to make that change, it has to be in a different PR.

libkineto/test/CuptiActivityProfilerTest.cpp
```diff
@@ -305,16 +324,76 @@ class MockCuptiActivities : public CuptiActivityApi {
...
-      size_t* size,
-      size_t* maxNumRecords) {
-    this->bufferRequested(buffer, size, maxNumRecords);
-  }
-
   std::unique_ptr<MockCuptiActivityBuffer> activityBuffer;
 };
 
+TEST(CuptiActivityApiTest, KeepsReturningValidBuffersBeforeVersion27) {
+  CallbackCuptiActivityApi cupti(26);
```
scotts: Since the constructor will no longer take a CUPTI version, we can make `canRejectBuffer_` protected and then we can set it in tests.

gtnv:
@scotts gotcha, fixed now. tested this on h100  @ 1 buffer limit: v26 returned 17 valid buffers, v27 accepted 1 and rejected 17, both exited w no crash or retained buffers as intended. all 103 kinteo tests passed.

lmk what other concerns u have.

libkineto/src/CuptiActivityApi.cpp
```diff
@@ -284,6 +288,12 @@ void CuptiActivityApi::bufferCompleted(
       return;
     }
 
+    if (stopCollection) {
+      // Discard returned buffers once tracing is terminating.
+      allocatedGpuTraceBuffers_.erase(it);
+      return;
```
scotts, line 293: Let's change this to `if (stopCollection && !_canRejectBuffer)`. That way, we'll only drop buffers on the versions where we're still returning buffers past the limit. (For the versions where we *can* drop the buffers, all buffers we see in `bufferCompleted()` should be pre-limit buffers with records we want to process.)

scotts:
@gtnv, thanks for your patience working on this! One small change, then we're good to go.

libkineto/src/CuptiActivityApi.cpp
gtnv: ahh ok. 👍

gtnv:
@scotts nw. thank you!

scotts (approved):
@gtnv, thanks for tackling this!

Issue #1533, opened 2026-08-11
XpuptiScopeProfilerTest.PerKernelScope is consistently failing


scotts:
See: https://github.com/pytorch/kineto/actions/runs/31429105419/job/93786166858?pr=1529
```
    Which is: 0
  numMetrics * repeatCount
    Which is: 15

/pytorch/kineto/libkineto/test/xpupti/XpuptiTestUtilities.cpp:85: Failure
Expected equality of these values:
  map.size()
    Which is: 6
  expSize
    Which is: 8

/pytorch/kineto/libkineto/test/xpupti/XpuptiTestUtilities.cpp:92: Failure
Expected equality of these values:
  countsMap.size()
    Which is: 2
  expMap.size()
    Which is: 3

/pytorch/kineto/libkineto/test/xpupti/XpuptiTestUtilities.cpp:101: Failure
Expected equality of these values:
  itCountsMap->second
    Which is: 4
  itExpArray->second
    Which is: 5

/pytorch/kineto/libkineto/test/xpupti/XpuptiTestUtilities.cpp:100: Failure
Expected equality of these values:
  itCountsMap->first
    Which is: 15
  itExpArray->first * repeatCount
    Which is: 10

/pytorch/kineto/libkineto/test/xpupti/XpuptiTestUtilities.cpp:101: Failure
Expected equality of these values:
  itCountsMap->second
    Which is: 2
  itExpArray->second
    Which is: 1

/pytorch/kineto/libkineto/test/xpupti/XpuptiTestUtilities.cpp:85: Failure
Expected equality of these values:
  map.size()
    Which is: 3
  expSize
    Which is: 4

/pytorch/kineto/libkineto/test/xpupti/XpuptiTestUtilities.cpp:100: Failure
Expected equality of these values:
  itCountsMap->first
    Which is: 5
  itExpArray->first * repeatCount
    Which is: 10

/pytorch/kineto/libkineto/test/xpupti/XpuptiTestUtilities.cpp:101: Failure
Expected equality of these values:
  itCountsMap->second
    Which is: 1
  itExpArray->second
    Which is: 2

[  FAILED  ] XpuptiScopeProfilerTest.PerKernelScope (203 ms)
[----------] 1 test from XpuptiScopeProfilerTest (203 ms total)

[----------] Global test environment tear-down
[==========] 1 test from 1 test suite ran. (203 ms total)
[  PASSED  ] 0 tests.
[  FAILED  ] 1 test, listed below:
[...]

PR #1536, opened 2026-08-12
Skip PerKernelScope UT

kdrozd-dev:
Temporarily skip `XpuptiScopeProfilerTest.PerKernelScope`, which has failed every XPU CI run since it first executed. Tracked by #1533.

The root cause is silent PTI failure on one of the devices:
`PTI:[13:07][-warning-]2170:2170 pti_metrics_scope_helper.h:953 CreateProfilerForDevice: StartProfiling failed for device 0x2209d238 (105); skipping metrics for this device (likely held by another process)`
This silent failure communicated with only a warning causes the test to fail with a value mismatch not an error hiding real cause. Despite that every call to `ptiMetricsScope*` returns `PTI_SUCCESS` making Kineto activity count come up short of what is expected. As far as I know there is no public PTI API that exposes that silent failure making it impossible to detect / report this in Kineto.


tsocha (approved):

kdrozd-dev:
@ryanzhang22 @gujinghui @scotts

ryanzhang22 (approved):

PR #1545, opened 2026-08-14
Discard completed CUPTI buffers when nothing is in flight

scotts:
**Problem**

Kineto drops activities recorded during the warmup period so it does not show up in the collected trace, and it drops leftover activitoes when resetting between sessions. Both go through `clearActivities()`, which gives up early whenever no buffers are in flight with CUPTI.

But having nothing in flight means there is nothing to ask CUPTI to flush; it says nothing about whether CUPTI has **already handed back completed buffers**. In the common case CUPTI returns buffers promptly, and the warmup flush runs on a runloop tick after everything has come back - the early return fires and the warmup records survive. They either land in the trace the user collects or carry over into the next session.

This problem is increased with #1534, where we still give back valid buffers after crossing the size limit for certain CUPTI versions.

**Fix**

Release the completed buffers before returning. This mirrors what #1362 did one function down in `activityBuffers()`. This also mitigates the harm of giving back valid buffers when we've hit the size limit, as we're actually clearing them out on the other side.

The new `ClearsCompletedBuffers` case fails without the fix. It also documents the two preconditions that keep the buffer tests away from live CUPTI - a null `CUcontext` that reaches `cuptiActivityGetNumDroppedRecords` when run at verbosity 1, and the drained allocated-buffer map that lets `activityBuffers()` and `clearActivities()` short-circuit instead of asking CUPTI to flush.

### DIFF
--- a/libkineto/src/CuptiActivityApi.cpp
+++ b/libkineto/src/CuptiActivityApi.cpp
@@ -253,6 +253,10 @@ void CuptiActivityApi::clearActivities() {
   {
     std::lock_guard<std::mutex> guard(mutex_);
     if (allocatedGpuTraceBuffers_.empty()) {
+      // Nothing is in flight, so there is nothing to flush - but CUPTI may have
+      // already handed back completed buffers, and callers ask us to clear so
+      // that those records do not reach the trace.
+      readyGpuTraceBuffers_.reset();
       return;
     }
   }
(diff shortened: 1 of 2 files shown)


libkineto/src/CuptiActivityApi.cpp
```diff
@@ -265,7 +269,7 @@ void CuptiActivityApi::clearActivities() {
   // for active tracing during warmup.
   std::lock_guard<std::mutex> guard(mutex_);
   // Throw away ready buffers as a result of above flush
-  readyGpuTraceBuffers_ = nullptr;
+  readyGpuTraceBuffers_.reset();
```
scotts, line 272: Stylistic drive-by. I think that `.reset()` is more clear than ` = nullptr`, and it's better to be consistent.

ryanzhang22 (approved):

PR #1546, opened 2026-08-20
[Kineto] CUPTI PM sampling improvements

ryanzhang22:
Two changes:
1. Support A100 machines, which do not support sampling frequency based on duration, only clock cycles. I had hoped to convert a requested duration to clock cycles by querying the device properties, but the frequency listed on there might not be the one that is used when profiling is run. I don't want to get into a cycle where I'm tuning some equation to make this work, so I am hardcoding something that seems reasonable for now and am open to changing this later. The initial value is based on the PM sampling doc example listing 100000 cycles -> 100 us.
2. Truncate and PM samples that fall outside the start, end range of profiling. Note that we need to fully drop instead of trimming any samples that fall out of range because the metrics are aggregated over the time period (mean, min, max, etc.) so changing the time period would change the value of the metric.

### DIFF
--- a/libkineto/src/CuptiPMSamplingApi.cpp
+++ b/libkineto/src/CuptiPMSamplingApi.cpp
@@ -49,6 +49,12 @@ namespace {
 constexpr size_t kHardwareBufferSizeBytes = 64 * 1024 * 1024;
 constexpr uint32_t kMaxSamplesPerDecode = 1024;
 
+// GA100 only supports variable-frequency SYSCLK sampling. One option is to
+// measure hardware clock frequency and estimate the number of cycles needed to
+// fill a certain sampling duration, but clock frequency is not necessarily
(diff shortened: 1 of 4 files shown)


sanrise (approved):
The filtering logic, the locking around the sampler lifecycle, decoder shutdown, and the CUPTI trigger/timestamp handling all look correct to me.

One nit before this lands: the new capture-window filtering and span recomputation don't have test coverage yet. You could, if you are in the mood for it, add a focused case that feeds three samples, one overlapping the window on the left, one fully contained, one overlapping on the right, and asserts that only the contained sample survives and the recomputed span matches it? That pins the boundary behavior so it can't silently regress later.

Otherwise this looks good.

ryanzhang22:
@sanrise Thanks for taking a look -- and yeah, I have a PR coming out with a bunch of profiler unit tests, including one that covers these new changes.

PR #1549, opened 2026-08-26
[xpupti] Don't emit ac2g flow endpoints on XPU_DRIVER records

aostrowski-hbn:
## What
Stop the XPU (xpupti) profiler plugin from attaching a CPU→GPU ("ac2g") flow
endpoint to `XPU_DRIVER` (Level Zero `ze*`) records.

## Why
The SYCL runtime record, the Level Zero driver (`ze*`) record and the device
kernel all share one PTI correlation id, and
`handleRuntimeKernelMemcpyMemsetActivities` set `flow.id`/`flow.type` on all of
them — so the driver record became an extra ac2g flow endpoint. In the exported
Chrome/Perfetto trace this draws a **redundant host→host arrow** from the
runtime "submit" slice to its own nested `ze*` child (already its child on the
same track), and the non-launch `ze*` calls — which get fresh correlation ids —
emit dangling flow finishes with no matching start. `ac2g` means *async
CPU→GPU*: a flow should link a host launch to the device op, never both start
and end on the host.

## How
A single `ac2gFlowRole()` helper maps the record's activity type to
`{None, Source, Destination}`, and `flow.id`, `flow.type` and `flow.start` are
set together from it. The host runtime record is the `Source`, kernel / memcpy /
memset are `Destination`s, and `XPU_DRIVER` records get `None` — so their flow
id stays 0 and `output_json`'s `flowId() > 0` guard emits no link. The `ze*`
slices themselves stay in the trace; only their redundant arrows go.

Naming the flow type in the predicate is deliberate rather than provisional:
`ac2g` is the only flow this plugin can emit — `fwdbwd` links are generated by
the PyTorch CPU-side profiler (`generateForwardBackwardLink`) from autograd
sequence numbers and never reach a device backend. `Source`/`Destination` also
says *which* end a record is, which the reviewed `carriesFlow` name did not.
Folding the two predicates (`startsFlow` + `carriesFlow`) into one removes a
second list of activity types that could drift out of sync with the first.

## Test
`XpuptiActivityHandlersTest` asserts the observable flow fields on hand-built
PTI records rather than re-asserting the predicate's own table: the runtime
record is a flow start, kernel / memcpy / memset are flow ends, and the
`XPU_DRIVER` record carries neither a flow id **nor** a flow type — an id with
type 0 would reach `handleGenericLink`, which then logs "Unknown flow type" for
every such record.

Built and the full xpupti ctest suite run on Intel GPU hardware (Intel® Arc™
Pro B-Series Graphics and Intel Data Center GPU Max Series); this revision was
re-verified on Intel Data Center GPU Max Series against PTI 1.2.0 — 22/22 pass
(`XpuptiScopeProfilerTest.PerKernelScope` is skipped upstream, #1533).
Before/after Perfetto traces confirm the redundant driver arrows drop to zero
while the runtime→kernel arrows are unchanged.

A companion PyTorch profiler test (kept in draft until this lands) guards the
same property end-to-end.

Companion (draft): https://github.com/pytorch/pytorch/pull/194904


aostrowski-hbn:
@guangyey, @gujinghui, @EikanWang please review

PawelSwider2000 (approved):

libkineto/src/plugin/xpupti/XpuptiActivityProfilerSession.h
```diff
@@ -73,6 +73,10 @@ class XpuptiActivityProfilerSession
   // Static so it can be unit-tested without real hardware.
   static bool startsFlow(ActivityType activityType);
 
+  // Whether the record is a CPU->GPU flow endpoint (runtime source or device
+  // destination); XPU_DRIVER subspans are not. Static for unit testing.
+  static bool carriesFlow(ActivityType activityType);
```
tsocha: You are checking `XPU_RUNTIME` inside, it's not CPU->GPU flow endpoint.

aostrowski-hbn: Correct — `carriesFlow` returned true for `XPU_RUNTIME`, which is the host side, so
"CPU→GPU flow endpoint" was the wrong description, and the name never said which of
the two flow types (fwdbwd, ac2g) it meant.

Both predicates are now folded into one `ac2gFlowRole()` returning
`{None, Source, Destination}`: the flow type is named in the predicate,
`Source`/`Destination` says which end a record is, and `flow.id`, `flow.type` and
`flow.start` are set together in one place instead of coming from two activity-type
lists that could drift apart.

Scoping the name to ac2g is accurate rather than provisional: `fwdbwd` links are
generated by the PyTorch CPU-side profiler (`generateForwardBackwardLink`, keyed on
autograd `forward_tid_`/`sequence_number_`) and never reach a device backend — and
`GenericTraceActivity::flow` is a single field, so a record can carry at most one
flow anyway.

I also dropped `XpuptiFlowCorrelationTest`. It re-asserted the switch's own table,
and the predicate was a public static member only so the test could reach it; the
helper is now file-local in `XpuptiActivityHandlers.cpp` and
`XpuptiActivityHandlersTest` asserts the resulting `Flow` fields instead — including
that the driver record's flow *type* stays 0 along with its id, since an id with
type 0 reaches `handleGenericLink` and logs "Unknown flow type" per record.

`main` merged in; re-verified on Intel Data Center GPU Max Series against PTI 1.2.0
(built from master), xpupti ctest 22/22.

libkineto/src/plugin/xpupti/XpuptiActivityHandlers.cpp
```diff
@@ -200,12 +224,13 @@ void XpuptiActivityProfilerSession::handleRuntimeKernelMemcpyMemsetActivities(
...
   trace_activity->endTime = activity->_end_timestamp;
   trace_activity->threadId = activity->_thread_id;
-  // Avoid a redundant flow arrow from the XPU_RUNTIME submit to its nested
-  // XPU_DRIVER (ze*) subspan: leave the flow id 0 so output_json's
-  // `flowId() > 0` guard skips the endpoint.
-  if (carriesFlow(activityType)) {
+  // Records with no role keep flow id 0, which output_json's `flowId() > 0`
+  // guard skips -- that is what suppresses the arrow.
+  if (const auto role = ac2gFlowRole(activityType);
+      role != Ac2gFlowRole::None) {
```
tsocha: I'm not a fan of this construct. It can be easily misread IMHO.
Could you move role declaration above the if statement?

aostrowski-hbn: Done.

tsocha (approved):

libkineto/src/plugin/xpupti/XpuptiActivityHandlers.cpp
```diff
@@ -159,6 +159,30 @@ static void addTimestampMetadata(
...
+// to the device operation it enqueued.
+enum class Ac2gFlowRole : uint8_t { None, Source, Destination };
+
+// XPU_DRIVER (ze*) subspans share the runtime record's correlation id but play
+// no part: they are nested under the submit on the same host track, so an arrow
+// to them would run host->host.
+Ac2gFlowRole ac2gFlowRole(ActivityType activityType) {
+  switch (activityType) {
+    case ActivityType::XPU_RUNTIME:
+      return Ac2gFlowRole::Source;
```
gujinghui, line 174: Here we assume the XPU_RUNTIME is the only source for all flows. May I know whether the XPU_DRIVER will be the source of some flows, when the XPU_RUNTIME is disabled by user, or libraries will launch kernels by driver API directly?

aostrowski-hbn: Good point. The driver will be the source when runtime is not traced now. As this needs to be checked per each activity I switched us to a faster way of checking which activities we trace (ActivityTypeMask) instead of the default std::set.

gujinghui (approved):

aostrowski-hbn:

aostrowski-hbn:
The CUDA failure is unrelated and seen also in other PRs (#1562, #1561, #1558).

aostrowski-hbn:
@ryanzhang22 @scotts can you take a look? The unrelated CI failure should not block this PR from being merged.

ryanzhang22:
Sorry was out on vacation for a few days. Will take a look.

libkineto/src/plugin/xpupti/XpuptiActivityProfilerSession.cpp
```diff
@@ -31,15 +31,6 @@ namespace KINETO_NAMESPACE {
 uint32_t XpuptiActivityProfilerSession::iterationCount_ = 0;
 std::vector<DeviceUUIDsT> XpuptiActivityProfilerSession::deviceUUIDs_ = {};
 
-bool XpuptiActivityProfilerSession::startsFlow(ActivityType activityType) {
-  // Only host runtime records start the CPU->GPU flow. The runtime view is
-  // already filtered to work-submitting APIs via
-  // ptiViewEnableRuntimeApiClass(PTI_API_CLASS_GPU_OPERATION_CORE). Driver
```
ryanzhang22, line 37: In XPU, is there an expectation that all driver calls produce a kernel event? I see that PTI_API_CLASS_GPU_OPERATION_CORE is used for filtering of runtime events, do you need the same thing for driver events?

JFYI for CUDA jobs we check if we need a flow arrow on the API call level: https://github.com/pytorch/kineto/blob/main/libkineto/src/CuptiCbidRegistry.cpp#L87

This lets us filter out individual calls (but we don't do the runtime -> driver fallback).

libkineto/src/plugin/xpupti/XpuptiActivityApi.cpp
```diff
@@ -204,12 +204,12 @@ void warnIfIttNotifyLibInvalid() noexcept {
...
 
 void XpuptiActivityApi::enableXpuptiActivities(
-    const std::set<ActivityType>& selected_activities) {
+    const ActivityTypeMask& selected_activities) {
   XPUPTI_CALL(ptiViewSetCallbacks(
       bufferRequestedTrampoline, bufferCompletedTrampoline));
 
   externalCorrelationEnabled_ = false;
-  for (const auto& activity : selected_activities) {
+  selected_activities.forEach([this](ActivityType activity) {
```
tsocha, line 16: Why did you switch from range based for to `forEach`?

libkineto/src/plugin/xpupti/XpuptiActivityApi.cpp
```diff
@@ -259,12 +259,12 @@ void XpuptiActivityApi::enableXpuptiActivities(
...
     }
-  }
+  });
 }
 
 void XpuptiActivityApi::disablePtiActivities(
-    const std::set<ActivityType>& selected_activities) {
-  for (const auto& activity : selected_activities) {
+    const ActivityTypeMask& selected_activities) {
+  selected_activities.forEach([](ActivityType activity) {
```
tsocha, line 59: This is the same lambda as line 212, isn't it?
Maybe you can reuse it for both `ForEach`s.

libkineto/src/plugin/xpupti/XpuptiActivityHandlers.cpp
```diff
@@ -178,17 +178,18 @@ static void addTimestampMetadata(
...
-// no part: they are nested under the submit on the same host track, so an arrow
-// to them would run host->host.
-Ac2gFlowRole ac2gFlowRole(ActivityType activityType) {
+enum class Ac2gFlowRole { None, Source, Destination };
+
+// Denotes the role an activity plays for Async CPU to GPU flow arrows. Both
+// XPU_RUNTIME and XPU_DRIVER events have the same correlation IDs and thus can
+// be sources, but exactly one must be chosen to avoid superfluous host->host
+// arrows. Prefer XPU_RUNTIME unless it is not being traced.
+Ac2gFlowRole ac2gFlowRole(ActivityType activityType, bool runtimeTraced) {
```
tsocha, line 22: `runtimeTraced` feels disconnected from its usage.
Maybe it should be called like `AllowDriverAsSource` or something like this?

libkineto/src/plugin/xpupti/XpuptiActivityTypeMask.h
```diff
@@ -0,0 +1,51 @@
...
+namespace KINETO_NAMESPACE {
+
+// A set of activity types held as a bitset: built once, queried per record.
+// Sized from the enum, so activity types can be added without outgrowing it,
+// and a value outside the enum throws instead of shifting out of range.
+class ActivityTypeMask {
+ public:
+  explicit ActivityTypeMask(const std::set<ActivityType>& types) {
+    for (const auto type : types) {
+      bits_.set(index(type));
```
tsocha: Maybe we should rename it to `as_index` or `get_index`. Something like this.
```suggestion
      bits_.set(as_index(type));
```

libkineto/src/plugin/xpupti/XpuptiActivityTypeMask.h
```diff
@@ -0,0 +1,51 @@
...
+    }
+  }
+
+  bool contains(ActivityType type) const {
+    return bits_.test(index(type));
+  }
+
+  // Calls visit() for each activity type in the set, in ascending order.
+  void forEach(std::invocable<ActivityType> auto&& visit) const {
```
tsocha, line 35: OK, now I see why you have switched to `forEach`.

libkineto/src/plugin/xpupti/XpuptiActivityApi.cpp
tsocha: Partially resolved in this comment: https://github.com/pytorch/kineto/pull/1549#discussion_r4060600644

libkineto/src/plugin/xpupti/XpuptiActivityProfilerSession.cpp
aostrowski-hbn: Not all driver calls produce kernels, but once we have filtering in place https://github.com/pytorch/kineto/pull/1570 virtually all of them will (except barriers, I think).

The user might still want to filter out driver events.

tsocha (approved):

ryanzhang22 (approved):

ryanzhang22:
@aostrowski-hbn Could you guys take a look at the XPU CI failure? Seems like it's happening on other PRs too.

PR #1551, opened 2026-08-27
[Kineto] CUPTI PM Sampling -- API layer unit tests

ryanzhang22:
Add unit tests for the CUPTI API layer (which is the wrapper over CUPTI). Also adds one test for output json.

For the CUPTI API tests, we're using a mock PM Sampler API to test behavior without actually needing to enable the profiler.

### DIFF
--- a/libkineto/test/CMakeLists.txt
+++ b/libkineto/test/CMakeLists.txt
@@ -135,6 +135,18 @@ target_include_directories(CuptiActivityProfilerTest PRIVATE
 gtest_discover_tests(CuptiActivityProfilerTest)
 
 if(KINETO_ENABLE_CUPTI_PM_SAMPLING)
+# CuptiPMSamplingApiTest
+add_executable(CuptiPMSamplingApiTest
+    CuptiPMSamplingApiTest.cpp)
+target_link_libraries(CuptiPMSamplingApiTest PRIVATE
+    gtest_main
+    kineto_base kineto_api)
+target_include_directories(CuptiPMSamplingApiTest PRIVATE
+    "${LIBKINETO_DIR}"
+    "${LIBKINETO_DIR}/include"
+    "${LIBKINETO_DIR}/src")
+gtest_discover_tests(CuptiPMSamplingApiTest)
+
 # CuptiPMSamplingProfilerTest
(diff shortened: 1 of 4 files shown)


libkineto/test/CuptiPMSamplingApiTest.cpp
```diff
@@ -0,0 +1,488 @@
...
+  EXPECT_EQ(samples[1].rawStartTimestamp, 100);
+  EXPECT_EQ(samples[1].rawEndTimestamp, 120);
+  EXPECT_EQ(samples[1].values, (std::vector<double>{1.25, 2048.0}));
+  EXPECT_EQ(samples[2].rawStartTimestamp, 130);
+  EXPECT_EQ(samples[2].rawEndTimestamp, 170);
+  EXPECT_EQ(samples[2].values, (std::vector<double>{2.5, 4096.0}));
+  api.disable();
+}
+
+TEST_F(CuptiPMSamplingApiTest, ReportsFullCounterDataAsUndrained) {
```
sanrise, line 454: Nit:

This hits COUNTER_DATA_FULL but with no completed samples, so a regression that drops a full batch (up to 1024) would still pass green. coukd consider add one fake sample and asserting it gets appended while decode() returns false? That'd lock down the full-batch path.

libkineto/test/CMakeLists.txt
```diff
@@ -135,6 +135,18 @@ target_include_directories(CuptiActivityProfilerTest PRIVATE
 gtest_discover_tests(CuptiActivityProfilerTest)
 
 if(KINETO_ENABLE_CUPTI_PM_SAMPLING)
+# CuptiPMSamplingApiTest
+add_executable(CuptiPMSamplingApiTest
```
sanrise, line 139: Quick q on wiring: do these actually run in the internal Buck build, or is this CMake/OSS-CI only? Afaik internal Buck doesn't build the PM-sampling impl, so if we're expecting fbcode coverage we'd need CUDA-compatibl Buck targets.

ryanzhang22: PM Sampling isn't wired up internally yet -- so no BUCK atm

PR #1553, opened 2026-08-31
[XPUPTI] Fix buffer lifetime and report processed-record stats with some minor improvements.

tsocha:
Port the CUPTI completed-buffer fix to the XPU plugin: activityBuffers() now returns already-completed buffers when nothing is in flight, clearActivities() drops them on the early-return path, and bufferCompleted() rejects unknown buffer pointers instead of dereferencing an invalid map iterator.

Replace processActivities()' std::pair<int, int> return with a named ActivitiesStats struct and log the record count/byte total, wiring up LOGGER_OBSERVER_ADD_EVENT_COUNT which XPU never reported.

Match the signature convention used by XpuptiActivityApi::processActivities and avoid copying the std::function on every call. The handler is only invoked in-scope, never stored.

### DIFF
--- a/libkineto/src/plugin/xpupti/XpuptiActivityApi.cpp
+++ b/libkineto/src/plugin/xpupti/XpuptiActivityApi.cpp
@@ -8,7 +8,7 @@
 
 #include "XpuptiActivityApi.h"
 
-#include <chrono>
+#include <cstdlib>
 #include <filesystem>
 #include <stdexcept>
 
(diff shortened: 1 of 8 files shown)


tsocha:
@aostrowski-hbn @gujinghui Please review.

aostrowski-hbn (approved):

gujinghui (approved):

tsocha:
@ryanzhang22 @scotts Could you review this PR, please?

ryanzhang22 (approved):

PR #1554, opened 2026-08-31
[Kineto] Fix CUPTI teardown race when two profiling sessions start in quick succession

ryanzhang22:
We are trying to enable `TEARDOWN_CUPTI=1` by default. One problem with this is that if two profiling sessions run in quick succession, because teardown happens async as a callback on the next CUDA call, the teardown for the first session can happen in the middle of the second session. This also happens if collection runs on a schedule and the user doesn't use a wait period. The result is usually that the second session unexpectedly doesn't have any GPU events.

To fix this, this PR adds a mechanism for the second session to cancel a pending CUPTI teardown. I prefer this over other solutions:
- Making the second session wait for the first to finish teardown is an unnecessary cost, because we're starting CUPTI back up as soon as it's torn down.
- Flipping TEARDOWN_CUPTI to 0 under schedules with no wait period doesn't help for sequential sessions. Worse, we can't be sure what the minimum number of wait iterations before teardown becomes worth it is, since this is model dependent.
- Having the user flip TEARDOWN_CUPTI manually seems like a cop-out.

How it works:
1. When a session completes, it queues teardown and moves the teardown state to `Preparing`
2. The teardown thread enables the callbacks needed for finalization, flushes CUPTI, and moves state to `Pending`
3. If a new session starts while finalization is `Pending`, it moves it to `Restoring`, cancelling the finalization.
4. If the teardown comes first, it moves the state to `Finalizing`. Any new session must then wait for the finalization to complete before starting.
5. New session preparation happens on the main thread, so we are sensitive to deadlocks. If new session preparation is waiting on finalization from a previous session, we cap the wait period to 10 seconds. This will break GPU profiling for the rest of the Kineto lifetime, but I'd rather break profiling than hang the training job.

Transitions between states are atomic so that all the threads/callbacks involved don't race with each other.

Testing:
Two simple tests
```
# Test 1
with profile(activities=[ProfilerActivity.CPU, ProfilerActivity.CUDA]) as prof:
    workload(...)
with profile(activities=[ProfilerActivity.CPU, ProfilerActivity.CUDA]) as prof:
    workload(...)

# Test 2
with profile(
    activities=[ProfilerActivity.CPU, ProfilerActivity.CUDA],
    schedule=schedule(warmup=1, active=2, repeat=2)
) as prof:
    workload(...)
```

Previously, the second session in both examples would result in no GPU activities.

I also had codex generate a test bench with a bunch of different scenarios.

> - 100 real-CUDA sync/async interleaving repetitions in each reinitialization mode, including both sync-to-async and async-to-sync transitions;
  > - pending-cancellation, callback-wins, Finalizing, Restoring, and timeout state-machine tests;
  > - repeated wait=0 schedules;
  > - early stop, exceptions, dynamic collection toggling, CUDA graphs, multistream and multithread workloads;
  > - process-exit and multi-process contention scenarios;
  > - behavior comparisons with teardown unset and set to 0;
  > - Kineto callback, activity-profiler, synchronous/asynchronous handler, and controller tests.
  >
  > The largest stress campaigns covered 24,256 case rows, 2,286 validated traces, and 19,322 teardowns, with no teardown timeout, deadlock, malformed trace, or missing required GPU activity.

### DIFF
--- a/libkineto/src/CuptiActivityApi.cpp
+++ b/libkineto/src/CuptiActivityApi.cpp
@@ -26,6 +26,9 @@ namespace KINETO_NAMESPACE {
 // case, there will be 32 buffers contending for the mutex.
 constexpr size_t kBufSize(4 * 1024 * 1024);
 constexpr uint32_t kCuptiBufferRejectionMinVersion = 27;
+// Stop waiting for the previous teardown after this interval and continue the
+// trace without GPU activities.
+constexpr auto kTeardownTimeout = seconds(10);
 
 inline bool cuptiTearDown_() {
   auto teardown_env = getenv("TEARDOWN_CUPTI");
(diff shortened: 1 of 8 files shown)


nzmsv:
I was also looking at this area - enabling `TEARDOWN_CUPTI`, and found this PR.  Great work!  I had prepared a fix that was mostly the same, but this is even nicer. 

I do have a follow-up improvement.  For my use case it's important that profiling deterministically shuts down.  I want to stop profiling, checkpoint the app, and resume profiling.

The root cause is that teardown doesn't actually happen until CUPTI invokes the next callback (https://docs.nvidia.com/cupti/main/main.html#dynamic-attach-and-detach).  But what triggers the callback?  It runs when the next CUDA API call returns.  In other words, teardown is entirely driven by the *workload*!  This is easy to prove: simply don't call any CUDA APIs, and the teardown will never happen.

How can we make this better?  Just call a side-effect-free CUDA API ourselves and trigger the callback.  I chose `cudaRuntimeGetVersion` because it's lightweight and doesn't modify state.

Here's a commit that's on top of yours: https://github.com/nzmsv/kineto/commit/d06cca69e0d4acb93464b227fe905b21b52b486f.  There's also a fixup on that branch that takes care of a TODO.

ryanzhang22:
Thanks @nzmsv -- if we add the deterministic shutdown by manually injecting`cudaRuntimeGetVersion`, isn't this just solution 1 (making the second session wait for the first to finish teardown) in this PR's summary? IIUC, then there would be no need for this cancellation PR.

If that's the case, imo we should do one of two things:
1. Have anyone who needs deterministic shutdown run a the CUDA_CALL outside Kineto in user code instead
2. If the logic absolutely needs to live in Kineto, block this behind a env var

As a general observation, Kineto is harder to reason about because we already have so many env vars controlling different aspects of CUPTI startup/shutdown, so I prefer not to bake it in unless it's a common use case. WDYT?

nzmsv:
Yes it is - as you say, a lock is still needed to make the start of the next profiling session wait for  teardown of the previous.  But it could just be a simple point fix.  That's the version I prepared before I saw this PR 😀 

I prefer the "injection" variant because for checkpointing the application is already quiescing, so  relying on it to issue CUDA calls is flaky.

ryanzhang22:
@nzmsv Got it. I'm okay with that logic going in as long as it needs to be explicitly enabled.

nzmsv:
What I liked about this change is the timeout: in case the teardown doesn't happen for whatever reason, there is cancellation instead of deadlock.

Why prefer a setting?  Wouldn't it be better to make teardown work out of the box?  Is the concern that a function is added to the profile that the application did not call?

ryanzhang22:
@nzmsv Are you asking why we don't deterministically teardown by default? My main concern is that `cuptiFinalize` is a heavy call and this change would make it run on every profiler exit, so repeated profiling sessions or profiling on a schedule could incur a performance penalty. That's the reason I put out this PR to be able to cancel a pending finalization in the first place.

nzmsv:
But the teardown itself is already gated by the TEARDOWN_CUPTI environment variable.  A call to cudaRuntimeGetVersion is a local memory read (I benchmarked it - adds microseconds).  The callback itself would normally still run without it - it would just be triggered by one of the application's calls.  I'd say that the trigger barely adds any overhead.

It can be combined with the cancellation idea - the commit I linked to is on top of this PR.  It's just a little nudge to make sure the teardown always makes forward progress.

ryanzhang22:
@nzmsv This PR is part of an effort to enable `TEARDOWN_CUPTI` by default, so that env var should go away, and every user will get this teardown logic. My concern isn't about the `cudaRuntimeGetVersion` call itself though, it's about unconditionally triggering `cuptiFinalize` on teardown and the fact that repeated profiling sessions will pay the startup/shutdown cost even when they don't need to.

ryanzhang22:
@nzmsv Thinking about it some more, I don't think we should land both cancellation and the explicit teardown trigger, it would make the complexity from this PR not really worth it.

I think the two options are to either:
1. Land this PR + make explicit trigger opt in, with explicit trigger behind a env gate. This preserves cancellation for back-to-back sessions and repeated schedules while supporting workloads that require teardown to make progress independently.
2. Land the explicit trigger, but modify this PR to keep only the timeout + block to make sure sequential profiling sessions wait for pending teardowns to complete.

Option 1 avoids unexpected latency between sessions, particularly at boundaries and repeat=N schedules with no wait period, but the use case is narrow/hypothetical and you are coming to us with a more concrete request.

Option 2 makes things simpler, I'd probably still want to add a killswitch for the explicit trigger until we validate no problems though (it can be on by default, I just need something toggle-able for internal).

I'll update the PR with the changes for option 2.

nzmsv:
@ryanzhang22 Agreed. If TEARDOWN_CUPTI becomes the default then most applications would already go through teardown - they would be doing enough CUDA calls to act as a trigger regardless.

So then cancellation preserves lazy CUPTI teardown - but only for applications that are quiesced.  Seems like the opposite of making TEARDOWN_CUPTI the default.

Here's a variant of the fix I had that I think is basically Version 2: https://github.com/nzmsv/kineto/commit/cf022a4a6bf44291aa9725817cef835b507eb4de

ryanzhang22:
@nzmsv To be clear, TEARDOWN_CUPTI was always a trigger for lazy teardown, and I still want to preserve that (in that we don't want profiler exit to wait for CUPTI to tear down, it'll always be on a separate thread + next CUDA call). I suspect that for most people waiting for the next CUDA_CALL to happen naturally is fine, but since your use case needs a stronger guarantee, then adding the manual trigger seems reasonable? I have no idea if there was a historical reason why they didn't do this previously.

I haven't reviewed that commit you linked fully yet, but it does seem like it needs the timeout + I'd prefer it the manual trigger was done separately. Do you want to modify it or do you want me to put out a PR?

nzmsv:
The interlock definitely either needs the trigger - or a timeout.  If there is no timeout then the trigger is not optional - an idle app will hang.

And apps could definitely just call some CUDA API when stopping profiling.  It's just that this is surprising behavior for users: they set TEARDOWN_CUPTI, drop the profiling object from their `with` clause, and expect it to tear down, not stick around forever.  Then they come to the CUDA team thinking that checkpointing is broken :) 

I'd prefer an explicit teardown.  It's only truly lazy for idle apps - and they are precisely the ones that don't care if there's a bit of extra latency due to actually doing teardown - they are quiesced anyway.  Or is this intuition wrong?

ryanzhang22:
@nzmsv By lazy I just mean that Kineto doesn't wait for CUPTI to tear down before exiting, it's on a separate thread. But ack on the need to do interlock + timeout and manual trigger in the same PR to prevent hang.

nzmsv:
You are right - stopping the profiler does not actually guarantee that CUPTI has torn down, even with an explicit trigger.  The teardown is still asynchronous.  Hmm - there's not really a way to confirm, is there?  That would need a new API to be exposed.

ryanzhang22:
@nzmsv Yeah, for this PR plus your requirements interlock + manual trigger seems like a good idea, but I guess it's another story if you need stronger guarantees.

PR #1555, opened 2026-09-01
[xpupti] Report the hardware engine per GPU op via PTI v2 records

aostrowski-hbn:
## Why

Intel GPU traces put one swim lane per SYCL queue. A SYCL queue is a *software*
queue: the driver maps its operations onto hardware engines, several queues can
share one engine, and a single queue's copies can be spread over several copy
engines. Nothing in the trace says which engine an operation actually ran on, so
two questions users regularly ask cannot be answered from a trace today:

- Did these two overlapping slices really run concurrently, or are they on the
  same engine with misleading timestamps?
- Did this copy ride a dedicated copy engine, or the compute engine?

Both matter when judging compute/copy overlap. On parts that expose a single
compute engine, overlapping SYCL-queue lanes do not imply concurrent execution
at all — and there is currently no way to tell that from the trace.

## How

PTI 1.0 introduced v2 device activity records (`pti_view_record_kernel_v2` and
friends) which extend the v1 layout with the hardware engine an operation ran
on. They are a binary extension of v1, reported under the same view kinds, so
the plugin only changes the record type it casts to — no new view kinds to
enable, no change in collection.

The engine is then emitted as per-op metadata:

- `engine_ordinal` — the engine group (compute, copy, …)
- `engine_index` — the engine within that group

matching how Level Zero reports command queue groups.

**Swim lanes are deliberately left unchanged:** still one per SYCL queue, named
`Stream <id>`, mirroring how CUDA lays out streams.

### Alternative considered and rejected: keying swim lanes on the engine

Splitting lanes per hardware engine is worse in both directions:

- When more than one copy engine is enabled, the Level Zero adapter
  **round-robins a single queue's copies across them**
  (`ur_queue_group_t::getQueueIndex`). Measured on Intel Data Center GPU Max
  Series with `UR_L0_USE_COPY_ENGINE=0:7`: the 32 copies of **one in-order
  queue** landed on **8 different engines** — one main copy engine plus seven
  link copy engines. Engine-keyed lanes would shred that single stream across
  eight tracks. (With default settings the same queue stays on a single copy
  engine, so this is configuration-dependent — but the configuration is a plain
  environment variable.)
- Conversely, on parts that expose a single compute engine, all compute lanes
  collapse into one, merging genuinely distinct queues.

Ordering is preserved either way: even with the copies spread over eight
engines, the trace shows **zero overlapping copy slices**, because the in-order
queue serialises them. So an engine axis does not buy overlap insight that the
queue axis lacks — which is the other reason to keep it out of the lane key.

Per-op metadata gives the same information without changing the shape of the
trace, and keeps the lane axis (the SYCL queue) the one that actually separates
independent streams of work.

Naming the engine *type* in the lane label was also dropped: it needs an
additional PTI query and buys nothing once lanes are not engine-keyed.

## Tests

`XpuptiActivityHandlersTest` gains three cases:

- `KernelActivityExposesEngineIds`
- `MemoryCopyActivityExposesEngineIds`
- `SwimLanesStayPerSyclQueueNotPerEngine` — pins the lane behaviour: a kernel and
  a copy on *different* engines but the *same* SYCL queue must share one lane,
  named `Stream 64`.

`ZeroDurationMemoryCopyOmitsBandwidth` moves to the v2 record type: a hand-built
v1 record would now be read past its end.

Built and ran on Intel Data Center GPU Max Series against PTI 1.2.0 —
`XpuptiActivityHandlersTest` 13/13 pass.

## Note on the minimum PTI version

Building the xpupti plugin now requires **PTI 1.0 or newer**, which is where the
v2 records were introduced.

### DIFF
--- a/libkineto/include/MetadataFieldCatalog.h
+++ b/libkineto/include/MetadataFieldCatalog.h
@@ -185,6 +185,8 @@ inline constexpr MetadataField<uint64_t> kBytes{"bytes"};
 inline constexpr MetadataField<uint64_t> kCommunicatorId{"Communicator_id"};
 inline constexpr MetadataField<uint64_t> kCorrelation{"correlation"};
 inline constexpr MetadataField<int64_t> kDevice{"device"};
+inline constexpr MetadataField<uint64_t> kEngineIndex{"engine_index"};
+inline constexpr MetadataField<uint64_t> kEngineOrdinal{"engine_ordinal"};
 inline constexpr MetadataField<uint64_t> kKernelId{"kernel_id"};
 inline constexpr MetadataField<double> kMemoryBandwidthGbps{
     "memory bandwidth (GB/s)"};
--- a/libkineto/src/plugin/xpupti/XpuptiActivityHandlers.cpp
+++ b/libkineto/src/plugin/xpupti/XpuptiActivityHandlers.cpp
@@ -167,11 +167,11 @@ void XpuptiActivityProfilerSession::handleRuntimeKernelMemcpyMemsetActivities(
   constexpr bool handleRuntimeActivities =
       std::is_same_v<pti_view_memory_record_type, pti_view_record_api_t>;
   constexpr bool handleKernelActivities =
-      std::is_same_v<pti_view_memory_record_type, pti_view_record_kernel>;
+      std::is_same_v<pti_view_memory_record_type, pti_view_record_kernel_t>;
   constexpr bool handleMemcpyActivities =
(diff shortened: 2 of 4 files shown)


aostrowski-hbn:
@jgong5 @EikanWang @gujinghui — CI is green; requesting an Intel PR approver review for the Draft→Open gate.

gujinghui (approved):

aostrowski-hbn:

ryanzhang22 (approved):

PR #1557, opened 2026-09-07
[xpupti] Guard v2 device records by PTI version

aostrowski-hbn:
## Why

The v2 device activity records (`pti_view_record_kernel_v2`, `pti_view_record_memory_copy_v2`,
`pti_view_record_memory_fill_v2`) only exist from PTI 0.18. #1555 started consuming them
unconditionally, so the plugin no longer compiles against an older PTI. The XPU `n-1` build
images still ship PTI 0.17, so `linux-jammy-xpu-n-1-py3.10 / build` and
`win-vs2022-xpu-n-1-py3 / build` on pytorch/pytorch main fail with
`'pti_view_record_kernel_v2' does not name a type`.

## How

Pick the record version with the `PTI_VERSION_AT_LEAST` macro the plugin already provides, and
fall back to the v1 records below PTI 0.18, where the hardware engine is then not reported. The
aliases move to namespace scope so the plugin tests build the same records the plugin consumes;
the tests asserting the engine metadata are compiled for PTI 0.18+ only.

### DIFF
--- a/libkineto/src/plugin/xpupti/XpuptiActivityHandlers.cpp
+++ b/libkineto/src/plugin/xpupti/XpuptiActivityHandlers.cpp
@@ -281,14 +281,16 @@ void XpuptiActivityProfilerSession::handleRuntimeKernelMemcpyMemsetActivities(
         XpuFields::kSyclQueue, static_cast<uint64_t>(activity->_sycl_queue_id));
     trace_activity->addMetadataQuoted(
         "l0 queue", handleToHexString(activity->_queue_handle));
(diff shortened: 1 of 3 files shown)


aostrowski-hbn:
@EikanWang @gujinghui @guangyey @jgong5 please review. Failure is unrelated (CUDA-specific, was already visible in #1556).

chuanqi129 (approved):

aostrowski-hbn:

ryanzhang22 (approved):

PR #1561, opened 2026-09-09
[Kineto] Add lookback window and sampling interval as config options

ryanzhang22:
These were introduced as inputs in https://github.com/pytorch/pytorch/pull/196065, but are not yet being parsed by Kineto (only cupti monitor/cuspy).

Lookback window is a proxy for buffer capacity, which we previously fixed as `maxSamplesForDecode` but is now calculated as `max(1, lookbackWindow / samplingInterval)`.

### DIFF
--- a/libkineto/include/Config.h
+++ b/libkineto/include/Config.h
@@ -236,6 +236,16 @@ class Config : public AbstractConfig {
     return performanceMetricsDeviceId_;
   }
 
+  [[nodiscard]] std::chrono::nanoseconds performanceMetricsSamplingInterval()
+      const {
+    return performanceMetricsSamplingInterval_;
+  }
+
+  [[nodiscard]] std::chrono::nanoseconds performanceMetricsLookbackWindow()
+      const {
+    return performanceMetricsLookbackWindow_;
+  }
+
   [[nodiscard]] bool memoryProfilerEnabled() const {
     return memoryProfilerEnabled_;
   }
@@ -375,6 +385,10 @@ class Config : public AbstractConfig {
   // Performance metrics
   std::vector<std::string> performanceMetricNames_;
   int32_t performanceMetricsDeviceId_{-1};
+  std::chrono::nanoseconds performanceMetricsSamplingInterval_{
+      std::chrono::milliseconds{1}};
+  std::chrono::nanoseconds performanceMetricsLookbackWindow_{
+      std::chrono::seconds{10}};
 
   // CUPTI Timestamp Format
   bool useTSCTimestamp_{true};
--- a/libkineto/src/Config.cpp
+++ b/libkineto/src/Config.cpp
@@ -9,10 +9,8 @@
 #include "Config.h"
 #include "ThrowUtil.h"
 
-#include <cerrno>
 #include <cmath>
 #include <cstdlib>
-#include <limits>
 
 #include <fmt/chrono.h>
 #include <fmt/format.h>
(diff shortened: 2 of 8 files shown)


libkineto/src/Config.cpp
```diff
@@ -228,6 +235,23 @@ bool isAllowedOnDemandTraceFile(const string& path) {
   return path.starts_with(dir) && path.find("..") == string::npos;
 }
 
+std::optional<nanoseconds> parseMilliseconds(std::string_view text) noexcept {
+  double count = 0;
+  const auto [parseEnd, error] =
+      std::from_chars(text.begin(), text.end(), count);
```
sanrise: std::from_chars needs const char*, but string_view::begin() isn't convertible under MSVC, so the common config target fails to compile on supported Windows builds. Use text.data() / text.data() + text.size()

libkineto/src/CuptiPMSamplingApi.cpp
```diff
@@ -272,15 +272,15 @@ void CuptiPMSamplingApi::configureCupti() {
...
-  // depends on sampling config, metrics, etc.
+  // Size the counter data image to hold the requested lookback window. CUPTI
+  // 12.x requires SetConfig to run before GetCounterDataSize.
   CUpti_PmSampling_GetCounterDataSize_Params counterDataSize{
       CUpti_PmSampling_GetCounterDataSize_Params_STRUCT_SIZE};
   counterDataSize.pPmSamplingObject = samplingObject_;
   counterDataSize.pMetricNames = metricNamePtrs_.data();
   counterDataSize.numMetrics = metricNamePtrs_.size();
-  counterDataSize.maxSamples = kMaxSamplesPerDecode;
+  counterDataSize.maxSamples = static_cast<uint32_t>(
```
sanrise: unbound int64_t sample count narrowed to CUPTI's uint32_t maxSamples. 

Configs abov UINT32_MAX wrap (possibly to zero), and large in-range values request enormous allocations.

libkineto/src/Config.cpp
sanrise, line 241: from_chars needs raw pointers. string_view::begin()/end() are not raw pointers on MSVC, so this breaks the Windows build. Please use text.data() and text.data() + text.size().

libkineto/src/CuptiPMSamplingApi.cpp
sanrise, line 282: maxSamples is a uint32_t, but this calculation can be much larger. A large lookback or tiny interval can wrap, possibly to zero, or request a huge buffer. Can we validate it before casting?

libkineto/src/CuptiPMSamplingApi.cpp
```diff
@@ -272,15 +272,15 @@ void CuptiPMSamplingApi::configureCupti() {
...
+      std::max<int64_t>(1, config_.lookbackWindow / config_.samplingInterval));
```
sanrise, line 283: GA100 uses the fixed 1M-cycle interval, but this sizes the buffer using the requested nanosecond interval, which GA100 ignores. This makes the lookback inaccurate, and zero causes division by zero. Can we use the effective GA100 cadence or reject this setting?

libkineto/include/Config.h
```diff
@@ -375,6 +385,10 @@ class Config : public AbstractConfig {
   // Performance metrics
   std::vector<std::string> performanceMetricNames_;
   int32_t performanceMetricsDeviceId_{-1};
+  std::chrono::nanoseconds performanceMetricsSamplingInterval_{
+      std::chrono::milliseconds{1}};
+  std::chrono::nanoseconds performanceMetricsLookbackWindow_{
```
sanrise, line 390: nit: With the 1ms interval above, this 10-second default increases capacity from 1,024 to 10,000 samples. Do we have memory and decode-cost measurements for this increase?

sanrise (changes requested):

libkineto/include/Config.h
ryanzhang22: thanks, updated. This was being used as the default for cuspy but we shouldn't need such a big buffer here.

sanrise (approved):

Issue #1359, opened 2026-04-10
[Rocprofiler-sdk] segfault on exit


scotts:
PR #1329 removed a hack to avoid segfaults on exit from rocprofiler. We had to remove it because it was causing more harm than good: it was swallowing _all_ failures. That is, no matter what happened, the process would return 0. That meant that if we hit errors in production, the process returned 0. If we hit errors in test, the process returned 0. As a consequence, across many systems, both in production and test, real failures were being misclassified as success.

An example of the segfault on exit:
```
INFO:2026-03-25 12:10:48 3903892:3904171 RocmActivityProfiler.cpp:68] HIP versions. Rocprofiler-sdk: 1.0; Runtime: 70051831; Driver: 70051831
*** Aborted at 1774465848 (Unix time, try 'date -d @1774465848') ***
*** Signal 11 (SIGSEGV) (0x7f5920dfccd0) received by PID 3903892 (pthread TID 0x7f59391db940) (linux TID 3903892) (code: address not mapped to object), stack trace: ***
    @ 0000000014132a1e folly::symbolizer::(anonymous namespace)::signalHandler(int, siginfo_t*, void*)
                       ./fbcode/folly/debugging/symbolizer/SignalHandler.cpp:551
    @ 000000000004455f (unknown)
                       /home/engshare/third-party2/glibc/2.34/src/glibc-2.34/signal/../sysdeps/unix/sysv/linux/libc_sigaction.c:8
                       -> /home/engshare/third-party2/glibc/2.34/src/glibc-2.34/signal/../sysdeps/unix/sysv/linux/x86_64/libc_sigaction.c
    @ 00000000000652da (unknown)
    @ 0000000000056352 (unknown)
    @ 00000000000567e8 (unknown)
    @ 0000000000046f74 (unknown)
    @ 00000000000471d8 (unknown)
    @ 000000000008765d (unknown)
    @ 0000000000090389 (unknown)
    @ 0000000000090603 (unknown)
    @ 0000000000062222 (unknown)
    @ 0000000000380359 (unknown)
    @ 00000000003805b4 (unknown)
    @ 00000000000478b7 __run_exit_handlers
                       /home/engshare/third-party2/glibc/2.34/src/glibc-2.34/stdlib/exit.c:113
    @ 00000000000479c9 exit
                       /home/engshare/third-party2/glibc/2.34/src/glibc-2.34/stdlib/exit.c:143
    @ 000000000002c65d __libc_start_call_main
[...]

PR #1564, opened 2026-09-12
[ROCm] Fix teardown memory corruption and truncated traces on the rocprofiler-sdk backend

haishuok0525:
## The bug

On the rocprofiler-sdk backend, stopping a trace corrupts the heap, and the trace it publishes is usually missing most of the GPU work without reporting anything.

Both symptoms come from the same place. `rocprofiler-sdk` does not emplace a kernel dispatch record at the moment a flush runs. The record is produced from the SDK's own signal handler thread ([`rocprofiler::hsa::AsyncSignalHandler`](https://github.com/ROCm/rocm-systems/blob/b1ff9c55a073d52c8f66b7119902632f8e2a18ab/projects/rocprofiler-sdk/source/lib/rocprofiler-sdk/hsa/queue.cpp#L231)) once the dispatch completion signal fires, and `rocprofiler_stop_context()` [does not wait](https://github.com/ROCm/rocm-systems/blob/b1ff9c55a073d52c8f66b7119902632f8e2a18ab/projects/rocprofiler-sdk/source/lib/rocprofiler-sdk/context/context.cpp#L366-L419) for sessions already in flight. So records keep arriving for tens of milliseconds after collection has stopped — straight through teardown.

`RocprofLogger::stopLogging()` flushed once and returned, so the rest of teardown then ran while the SDK thread was still appending to `rows_` and `externalCorrelations_` under `rowsMutex_`. Two places touch those containers without taking the mutex:

- `processActivities()` walks both in place. A `push_back` that reallocates mid-walk invalidates the iteration. #1424 has since added a third unlocked traversal of `rows_` via `backfillAsyncStreams`.
- `clearLogs()` calls `rows_.clear()` and `externalCorrelations_[i].clear()` unlocked, concurrently with those `push_back`s. That one is a write-write race, and it is the more serious of the two: it frees a buffer another thread is growing.

So the outcome is memory corruption, not merely a bad read. glibc catches it when it next inspects the heap:

```
INFO: ... output_json.cpp:1133] Renamed the trace file to /tmp/rpm/out.json
malloc(): unsorted double linked list corrupted
```

Note where that lands — after the trace was written, in teardown, which is where `clearLogs()` runs.

How it surfaces depends on timing. Sometimes the invalid iteration faults first:

```
libkineto::RocprofActivityApi::processActivities(...)                       <-- faults here
libkineto::RocmActivityProfiler::processGpuActivities(ActivityLogger&)
libkineto::GenericActivityProfiler::processTraceInternal(ActivityLogger&)
libkineto::SyncActivityProfilerHandler::stopTrace()
libkineto::ActivityProfilerController::syncStopTrace()
libkineto::ActivityProfilerProxy::stopTrace()
main
```

and sometimes the process gets as far as writing the trace and then dies, or wedges, on the corrupted heap. It is a race, so the rate moves with machine load: batches on one unchanged build and workload have run from 6/20 to 18/20 failures. Between a third and nine tenths of runs.

## The other half: silently truncated traces

The runs that survive are still wrong. Instrumenting the row count around teardown shows how much is still in flight when the single flush returns:

```
KDIAG rows: afterFirstFlush=1937 afterSleep=6215 afterSecondFlush=6600
```

The first flush returns with 29% of the records. A single-GPU HIP graph replay that dispatches exactly 6400 kernels (32 nodes x 200 replays, plain `torch.profiler`, no ROCm-specific options) reported between 1251 and 5681 of them, a median of 1874. Not one of the 15 surviving runs was complete.

The loss is FIFO rather than random, which is what makes it dangerous: the trace is a prefix of the real one, so it is internally consistent and looks plausible. Every one of the 32 graph nodes appears 54 to 58 times instead of 200, so a per-node analysis still produces an answer, just a wrong one. The same shape shows up on a production SGLang trace: 1507 kernels over 32 node ids, each appearing 44 to 49 times.

## The fix

Stop the context so no new dispatches can be recorded, then flush until the row count stops growing. Take each container under its own mutex — `processActivities()` swaps them out and walks a local copy, which also covers the `#1424` traversal, and `clearLogs()` clears them locked.

Draining first makes a late append unlikely but not impossible, so the locking still has to hold for the case where the bounded wait expires.

The drain converges in 3 flushes on the repro, so teardown grows by roughly 30ms. The attempt bound exists so a pathological producer cannot hang teardown; if it is ever reached, the code warns and publishes what arrived, which is exactly today's behaviour.

## Alternatives that do not work

Two simpler fixes were measured first, and neither is sufficient.

| Approach | Failures | Complete traces | Kernel median |
|---|---|---|---|
| main, flush then stop | 15 / 30 | 0 of 15 | 1874 (29.3%) |
| ordering corrected only, stop then flush once | 13 / 30 | 0 of 17 | 1595 (24.9%) |
| `hipDeviceSynchronize()`, then stop and flush | 5 / 20 | 0 of 15 | 1894 (29.6%) |
| this change | 0 / 30 | 30 of 30 | 6400 (100%) |

Main's ordering is wrong on its own terms — flushing before stopping leaves the window between the two calls able to produce records — but correcting only that moves neither symptom. The failure rate difference is noise (Fisher two-sided p = 0.80) and the truncation is untouched.

It is still a precondition for the fix rather than an incidental cleanup. Once the context is stopped, no new dispatch can be recorded, so only the already in-flight ones remain and the row count is able to converge. Flushing first, or not stopping at all, leaves an application that keeps dispatching able to feed the buffer indefinitely, and the loop below would never settle.

`hipDeviceSynchronize()` is the other intuitive fix and it is not enough either. It guarantees the GPU work finished and the completion signals fired, but not that the SDK has run its handler for those signals, and that handler is what emplaces the record.

## Why this has to be a loop

The obvious objection is that polling is a workaround for a missing barrier, so it is worth saying why the barrier cannot simply be borrowed.

rocprofiler-sdk does track exactly what we need. [`hsa::Queue`](https://github.com/ROCm/rocm-systems/blob/b1ff9c55a073d52c8f66b7119902632f8e2a18ab/projects/rocprofiler-sdk/source/lib/rocprofiler-sdk/hsa/queue.hpp#L164-L179) counts in-flight dispatches in `_active_kernels`, incremented at dispatch and decremented in `async_complete()`, which the async handler calls only *after* `dispatch_complete()` has emplaced the record. `Queue::sync()` waits on that count reaching zero.

It is not a barrier a tool can use, for two reasons. First, `sync()` is a single ~5s wait that returns `bool`; teardown callers warn and proceed, and the SDK's own [kernel replay drain](https://github.com/ROCm/rocm-systems/blob/b1ff9c55a073d52c8f66b7119902632f8e2a18ab/projects/rocprofiler-sdk/source/lib/rocprofiler-sdk/hsa/queue.cpp#L136-L163) has to call it in a loop of up to 12 slices before giving up. Second, and more fundamentally, that drain only converges because replay holds a per-agent writer lock that blocks new dispatches while it waits, so in-flight work only decreases. A profiler stopping a trace has no such lock: the application frequently keeps dispatching, so "zero in flight" is not a state that is guaranteed to arrive.

So the loop is not a workaround for the SDK's loop — it is the same shape of solution under weaker preconditions, which is also why it needs a bound. What would let this collapse into a single call is not `sync()` being exposed, but a documented guarantee on the buffer side: that `rocprofiler_flush_buffer()` delivers records whose dispatch had already completed when it was called. Today it documents no semantics, and `rocprofiler_destroy_buffer()` suggests calling flush first "to make sure the buffer is empty". Raised with the SDK separately.

## Validation

Single GPU, HIP graph replayed around a profiling session, 30 runs per arm under gdb:

| | before | after |
|---|---|---|
| teardown crash | 15 / 30 | 0 / 30 |
| complete trace | 0 / 15 surviving runs | 30 / 30 |
| kernels reported | 1251 - 5681 of 6400 (median 1874) | 6400 every run |

All 15 crashes hit the same site, `processActivities`. The drain converges in 3 flushes on 27 of the 30 runs and 1 or 2 on the rest.

Re-measured after rebasing onto current main, both arms back to back on an idle MI355X / ROCm 10.0.0, 20 runs each, with a 60s timeout per run so a wedged process is counted rather than waited on:

| | main | main + this change |
|---|---|---|
| clean runs | 10 / 20 | 20 / 20 |
| segfault | 4 | 0 |
| hung after corruption | 6 | 0 |
| glibc heap messages | 6 | 0 |
| kernels on surviving runs | mean 1361, range 955 - 1845 | 6400 every run |

**Not specific to HIP graphs.** Graph replays make the race easy to hit, because graph-launched dispatch rows reach the callback thread only after `hipGraphLaunch` has returned. But issuing the same 6400 dispatches as plain stream launches fails too, and truncates worse. 20 runs per cell:

| | workload | failures | kernels (mean) |
|---|---|---|---|
| main | graph | 8 / 20 | 2781 (43%) |
| main | stream | 5 / 20 | 478 (7%) |
| this change | graph | 0 / 20 | 6400 (100%) |
| this change | stream | 0 / 20 | 6400 (100%) |

So this is a fix for the backend generally, not for a graph-specific corner.

**Isolated against #1567**, since the two ROCm PRs currently open touch the same file and it should be clear which one is load-bearing here. 20 runs per variant:

| variant | failures | kernels (mean) |
|---|---|---|
| main | 18 / 20 | 3374 (53%) |
| main + this change | 0 / 20 | 6400 (100%) |
| main + #1567 | 16 / 20 | 3994 (62%) |
| main + both | 0 / 20 | 6400 (100%) |

#1567 neither causes nor mitigates either symptom. This change fixes both with or without it.

**Traces are otherwise unchanged:** five runs per arm with four concurrent streams match on event counts per category, external id coverage, per-stream kernel distribution, and kernel-to-runtime correlation pairing. `RocmActivityProfilerTest` passes, 11 of 11.

## Possible relation to #1359

#1359 is an open report of a SIGSEGV on this backend whose signal arrives from `__run_exit_handlers`, with the middle frames unsymbolized. I cannot claim this fixes it. But it is worth noting that the failure here is heap corruption inflicted during profiler teardown, which is detected whenever the allocator next looks — and "later, at process exit, in frames that have nothing to do with the profiler" is precisely what that looks like. Same backend, and a defect whose signature matches. The MSLK test #1359 names is public, so it seems worth running against this change; happy to do it if someone can point me at a way to run MSLK on ROCm.

## Repro

Plain `torch.profiler` plus a HIP graph replay, no ROCm-specific profiler options needed. There is also a standalone C++ reproducer that drives libkineto through its public API and needs no torch, which is what produced the numbers above — posted in a comment below.


haishuok0525:
Attaching the standalone reproducer referenced in the description, in case it is useful for reviewing this. It drives libkineto directly through its public API, so it needs no torch and no ROCm-specific profiler options — just a GPU and a ROCm build of this repo.

It captures a HIP graph, replays it inside a trace window, and saves a chrome trace. Since the number of dispatches is known exactly (`E2E_NODES * E2E_REPLAYS`, 6400 by default), both symptoms are countable: whether the process survived `stopTrace()`, and how many kernel events the trace actually contains.

`E2E_STREAM=1` issues the same number of dispatches as plain stream launches instead, which is what separates "graphs are involved" from "records are still arriving during teardown".

gdb is not in the image I was testing in, so it installs a signal handler and prints its own frames. One caveat that turned out to matter: on a corrupted heap, that handler can itself wedge, because `backtrace_symbols_fd` allocates. That is why the table in the description counts hangs separately from segfaults — the underlying event is the same, and a 60s timeout per run is enough to tell them apart.

**Build**

```bash
# libkineto
cmake -DKINETO_BACKEND=rocm -DROCM_SOURCE_DIR=/opt/rocm -DCMAKE_BUILD_TYPE=Release \
      /path/to/kineto/libkineto
make -j kineto

# the repro, against the static lib just built
K=/path/to/kineto/libkineto
hipcc -std=c++20 -O2 -c e2e_graph.cpp -I$K/include -I$K/third_party/fmt/include -o repro.o
hipcc repro.o libkineto.a -L/opt/rocm/lib -lrocprofiler-sdk -ldl -lpthread -o repro
```

**Run**

```bash
export HIP_VISIBLE_DEVICES=0
for i in $(seq 20); do
  rm -f out.json
  timeout 60 env E2E_OUT=$PWD/out.json ./repro > run$i.log 2>&1
  rc=$?
  k=$(python3 -c 'import json;t=json.load(open("out.json"));
print(sum(1 for e in t.get("traceEvents",[]) if e.get("cat")=="kernel"))' 2>/dev/null || echo 0)
  echo "run $i: rc=$rc kernels=$k/6400"
done
grep -h 'malloc()\|free():\|corrupted' run*.log | sort | uniq -c
```

`rc=0` is a clean run, `rc=139` a segfault, `rc=124` a process that had to be timed out.

On current main expect a mix of clean runs, segfaults and timeouts, and `kernels` well under 6400 even on the clean ones — 10 of 20 clean and a mean of 1361 kernels on the last batch I ran. With this PR applied, 20 of 20 clean and 6400 every time.

<details>
<summary><code>e2e_graph.cpp</code></summary>

```cpp
// End-to-end check of kineto PRs #1564 and #1567 using the real libkineto,
// not a transcription of it.
//
// Profiles a HIP graph replay through kineto's own public API and writes a
// chrome trace. Whether the PRs work is then a question about the trace file:
// does every graph kernel carry "graph id" and "graph node id", are there as
// many distinct node ids as the graph has nodes, and did the trace keep all
// the dispatches instead of being truncated at teardown.
//
// E2E_NODES   nodes in the captured graph
// E2E_REPLAYS replays inside the trace window

#include <hip/hip_runtime.h>

#include <libkineto.h>

#include <csignal>
#include <cstdio>
#include <cstdlib>
#include <execinfo.h>
#include <set>
#include <string>
#include <unistd.h>

// gdb is not available in the test image, so catch the fault in-process and
// print the frames. The point is to see which component is on the stack when
// an unpatched build dies at teardown.
extern "C" void segv_handler(int sig)
{
    void*  frames[32];
    int    n = backtrace(frames, 32);
    fprintf(stderr, "\n=== caught signal %d, %d frames ===\n", sig, n);
    fflush(stderr);
    backtrace_symbols_fd(frames, n, STDERR_FILENO);
    _exit(139);
}

#define HC(x)                                                                            \
    do                                                                                   \
    {                                                                                    \
        hipError_t e = (x);                                                              \
        if(e != hipSuccess)                                                              \
        {                                                                                \
            fprintf(stderr, "HIP error %s at %d: %s\n", #x, __LINE__,                    \
                    hipGetErrorString(e));                                               \
            exit(1);                                                                     \
        }                                                                                \
    } while(0)

__global__ void e2ekernel(float* p, int n)
{
    int i = blockIdx.x * blockDim.x + threadIdx.x;
    if(i < n) p[i] += 1.0f;
}

static long envLong(const char* k, long d)
{
    const char* v = getenv(k);
    if(!v || !*v) return d;
    char* e = nullptr;
    long  x = strtol(v, &e, 0);
    return x > 0 ? x : d;
}

int main()
{
    signal(SIGSEGV, segv_handler);
    signal(SIGABRT, segv_handler);

    const long nodes   = envLong("E2E_NODES", 32);
    const long replays = envLong("E2E_REPLAYS", 200);
    const int  n       = 1024;

    float* buf = nullptr;
    HC(hipMalloc(&buf, n * sizeof(float)));
    HC(hipMemset(buf, 0, n * sizeof(float)));
    hipStream_t s;
    HC(hipStreamCreate(&s));

    // Capture the graph before tracing starts, the way a framework would.
    hipStream_t cap;
    HC(hipStreamCreate(&cap));
    HC(hipStreamBeginCapture(cap, hipStreamCaptureModeGlobal));
    for(long i = 0; i < nodes; ++i)
        hipLaunchKernelGGL(e2ekernel, dim3(1), dim3(64), 0, cap, buf, n);
    hipGraph_t g = nullptr;
    HC(hipStreamEndCapture(cap, &g));
    hipGraphExec_t exec = nullptr;
    HC(hipGraphInstantiate(&exec, g, nullptr, nullptr, 0));

    // Warm up outside the trace window.
    for(int i = 0; i < 5; ++i) HC(hipGraphLaunch(exec, s));
    HC(hipStreamSynchronize(s));

    libkineto_init(false, true);
    std::set<libkineto::ActivityType> types;
    auto&                             profiler = libkineto::api().activityProfiler();
    libkineto::api().initProfilerIfRegistered();
    profiler.prepareTrace(types);

    for(int i = 0; i < 5; ++i) HC(hipGraphLaunch(exec, s));
    HC(hipStreamSynchronize(s));

    // E2E_STREAM=1 issues the same number of dispatches as plain launches, to
    // separate "graphs are involved" from "records arrive during stopTrace".
    const bool plain = getenv("E2E_STREAM") != nullptr;
    printf("start trace: %ld dispatches expected (%s)\n",
           nodes * replays,
           plain ? "plain stream launches" : "graph replays");
    profiler.startTrace();

    if(plain)
        for(long i = 0; i < nodes * replays; ++i)
            hipLaunchKernelGGL(e2ekernel, dim3(1), dim3(64), 0, s, buf, n);
    else
        for(long i = 0; i < replays; ++i) HC(hipGraphLaunch(exec, s));
    HC(hipDeviceSynchronize());

    auto trace = profiler.stopTrace();
    printf("activities in trace: %zu\n", trace->activities()->size());

    const char* out = getenv("E2E_OUT");
    std::string path = (out && *out) ? out : "/tmp/e2e_trace.json";
    trace->save(path);
    printf("saved %s\n", path.c_str());
    return 0;
}
```

</details>

libkineto/src/RocprofLogger.cpp
```diff
@@ -811,10 +817,37 @@ void RocprofLogger::stopLogging() {
     return;
   logging_ = false;
 
-  // Flush buffers
   auto& globalContext = getGlobalContext();
-  rocprofiler_flush_buffer(globalContext.buffer);
   rocprofiler_stop_context(globalContext.context);
+
+  // Stopping the context does not wait for dispatches that are already in
```
ryanzhang22, line 823: Is continuously polling the only way to wait for rocprofiler-sdk to potentially finish draining? In the CUDA paths we have a CUPTI API which blocks and forces a flush, even for incomplete buffer records.

I see that the other changes have made concurrent modification on the rows vector safer, which is a good change, but I'm wondering if there's any way we can have a stronger guarantee on completeness here?

mwootton:
So the 816 and 817 lines are swapped in the original file.  Trivial mistake.

> rocprofiler_stop_context()
> rocprofiler_flush_buffer()

is the correct order.

I thought that should be sufficient.  I had an agent dig through the rocprofiler_flush_buffer() implementation and it appears not (see the TLDR above).  This is the same dumb situation as roctracer; and it probably needs the same workaround.  The RoctracerLogger used to track the highest outstanding correlation id at the time of the 'stop' and waited for that record to come through.  Rocprofiler-sdk can guaranty record delivery on finalize, but not mid flight stop; which is what we need here.

The heap corruption if from the late callbacks happening after tracing is stopped; those should not be happening if the flushing is done correctly.  So the stop -> flush -> readout phases SHOULD be exclusive without the mutex.  We can add the mutex for safety.  But in this case it is also "papering over" the bigger problem.

It's a pretty big change to adopt the correlation tracking.  I will inquire about a behavior change in the sdk.  There are other backends that don't have this drain problem; so this is a choice.

libkineto/src/RocprofLogger.cpp
haishuok0525: Thanks, and sorry for the slow reply.

As far as I can tell, rocprofiler-sdk has no equivalent of CUPTI's forced flush. `rocprofiler_flush_buffer()` only hands over records already emplaced in the buffer, and a dispatch record is emplaced later, by the SDK's own signal handler thread once the completion signal fires. The SDK guarantees delivery at finalize, but not at a mid-run stop, which is the case we need here.

The option with a hard guarantee would be tracking correlation ids: note the highest one issued at stop and wait until its record comes through. That's a considerably bigger change. @mwootton is raising a behaviour change with the SDK team instead.

In the meantime I've cut this PR down to just the bounded wait: flush, stop, then flush until the row count has been stable for 30ms, with a 5s total timeout that warns. That alone fixes both the truncation and the crash on the repro (numbers are in the updated description). The locking changes are out of this PR and can follow separately if wanted.

haishuok0525:
> So the 816 and 817 lines are swapped in the original file.

Thanks Michael. Updated as you suggested: this PR now only fixes the ordering and adds the wait. It flushes, stops the context, then keeps flushing until the row count is stable, with a 5s total timeout that logs a warning. The locking changes are gone.

The wait alone is enough on the repro. Over 40 runs, main segfaulted or hung on 27 and never produced a complete trace. With this change all 40 exited cleanly and 39 were complete; the one exception was short by a single kernel. Details are in the description.

Agreed that correlation tracking or an SDK-side drain is the real fix. Thanks for raising it with the SDK team.

libkineto/src/RocprofLogger.cpp
```diff
@@ -811,10 +813,36 @@ void RocprofLogger::stopLogging() {
     return;
   logging_ = false;
 
-  // Flush buffers
   auto& globalContext = getGlobalContext();
   rocprofiler_flush_buffer(globalContext.buffer);
```
ryanzhang22, line 817: isn't this still wrong order (should be stop, then flush)? tbh we can probably just `rocprofiler_stop_context` and then let the loop below handle flush

libkineto/src/RocprofLogger.cpp
ryanzhang22: If we want to simplify this, wouldn't the safest thing be to keep the changes that guard concurrent modification instead? The sleep loop can still let stuff past if we're unlucky, and I'd much rather live with the consequences of dropping a few events than segfaulting a training job.

haishuok0525: That's fair, and some history would help here.

The first version of this PR did both: stop the context, flush until the row count stopped growing, and take `rowsMutex_` / `externalCorrelationsMutex_` in `processActivities()` and `clearLogs()`.

After discussing it with @mwootton, I narrowed it for two reasons. First, the root cause is on the SDK side: rocprofiler-sdk has no reliable drain at a mid-run stop, and that is being raised with the SDK team. We didn't want the mutex to read as the fix for that, when what it really does is make a racy teardown survivable. Second, once the wait drains the in-flight records, nothing is appending to `rows_` when `processActivities()` and `clearLogs()` run, so the normal path doesn't need the lock. I checked that before dropping it: over 40 runs with the wait alone there were 0 crashes, against 27 of 40 crashed or hung on main.

You're right that the loop is a heuristic, though. If the 5s timeout expires, or a long-running kernel completes after the row count already looked stable, a record can still land while those functions are running, and without the lock that's heap corruption again rather than a few missing events. I agree with your trade-off: a slightly short trace is much better than taking down a training job. I'm happy to put the locking back in this PR. @mwootton, any objection?

libkineto/src/RocprofLogger.cpp
haishuok0525: The flush before stop came from the same discussion: it follows the flush / disable / flush pattern in the SDK samples, so whatever is already buffered gets delivered before the context stops. You're right that it's redundant here, since the loop flushes after the stop anyway. I'll drop it, so it becomes stop, then the flush loop.

PR #1570, opened 2026-09-21
[XPUPTI] Limit the number of events recorded for XPU_DRIVER activity.

tsocha:
Currently PTI reports 200+ types of events for `XPU_DRIVER` activity type.
This PR reduce it to 20 related to launching kernels.
It will reduce size of the trace and limit the noise caused by L0 calls.

Narrow XPU_DRIVER to PTI's GPU_OPERATION_CORE class. The class filter has to follow ptiViewEnable(), which enables all api ids and clears the granularity flag; the first granular call then resets the group to all-disabled before enabling its class.

zeCommandQueueExecuteCommandLists submits work to the GPU but is absent from PTI's class list, so enable it explicitly until that is fixed upstream.


tsocha:
@gujinghui please review.

libkineto/src/plugin/xpupti/XpuptiActivityApi.cpp
```diff
@@ -231,11 +231,24 @@ void XpuptiActivityApi::enableXpuptiActivities(
...
+        XPUPTI_CALL(ptiViewEnableDriverApiClass(
+            /*enable=*/1,
+            PTI_API_CLASS_GPU_OPERATION_CORE,
+            PTI_API_GROUP_LEVELZERO));
+        // Queue submission submits work to the GPU but is absent from PTI's
+        // GPU_OPERATION_CORE class.
+        XPUPTI_CALL(ptiViewEnableDriverApi(
+            /*enable=*/1,
+            PTI_API_GROUP_LEVELZERO,
+            pti_api_id_driver_levelzero::zeCommandQueueExecuteCommandLists_id));
```
gujinghui, line 251: @tsocha  Should we ask the PTI team to add this API to the PTI_API_CLASS_GPU_OPERATION_CORE list?

Ideally, we should avoid hardcoding specific API names in pytorch/kineto. Asking the PTI team to handle this internally on their end would save us the effort of maintaining this custom bypass in the future.

gujinghui: @tsocha Could you help address my below questions with this change? Thanks.
1. If the kernel is launched by SYCL API, both SYCL kernel launch and L0 kernel launch will be traced, or only traces the SYCL kernel launch?
2. If a lib launches kernel with L0 API directly, will the kernel launch be traced?
3. All other implicit L0 calls will be hidden, for example, the event creations, host synchronization?

jfedorov: @tsocha @gujinghui Yes - adding zeCommandQueueExecuteCommandLists to PTI_API_CLASS_GPU_OPERATION_CORE (in PTI ) makes sense and this is to add one line. 

@tsocha can you pls do it.

tsocha: `pti_api_id_driver_levelzero::zeCommandQueueExecuteCommandLists_id` will be added into `PTI_API_CLASS_GPU_OPERATION_CORE` soon, but it won't be available until new oneAPI will be used in PyTorch.
Are you OK to keep it here for now?

1. SYCL will produce L0 kernel launch event so it will be traced.
2. It should be captured by XPU_DRIVER activity. [list of events](https://github.com/intel/pti-gpu/blob/master/sdk/src/view_handler.h#L93)
3. Yes, it will be hidden. For synchronization events I need to create a follow-up PR to enable `XPU_SYNC` activity type by default, it have to be enabled in [PyTorch](https://github.com/pytorch/pytorch/blob/main/torch/csrc/profiler/kineto_shim.cpp#L52).

gujinghui: Thanks for quick response.

> Yes, it will be hidden. For synchronization events I need to create a follow-up PR to enable XPU_SYNC activity type by default, it have to be enabled in [PyTorch](https://github.com/pytorch/pytorch/blob/main/torch/csrc/profiler/kineto_shim.cpp#L52).

@tsocha  Can we have this first? The SYNC activity is important.

> pti_api_id_driver_levelzero::zeCommandQueueExecuteCommandLists_id will be added into PTI_API_CLASS_GPU_OPERATION_CORE soon, but it won't be available until new oneAPI will be used in PyTorch.
> Are you OK to keep it here for now?

We'd better have the fix in PTI first. @jfedorov May I know the plan for upcoming PTI release? Thanks.

ryanzhang22:
@tsocha @gujinghui Just a heads up on https://github.com/pytorch/kineto/issues/1571.

We are planning to lock this repo tomorrow -- you'll need to recreate this PR in pytorch/pytorch on Monday if this hasn't landed by then. Dir structure and everything should be the same, so it should be trivial to copy the changes over. Thanks!

tsocha:
@ryanzhang22
Thank you for the reminder.
I wanted to pass the review process before the lock.

I will migrate this PR to the PyTorch repository when all issues will be resolved.

